PR #45, opened 2022-11-03
Pip package, update readme, contributing.md

dlwh:
could be broken up, but they're kinda the same thing...

CONTRIBUTING.md is shamelessly copied from mistral.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,104 @@
+Contributing
+============
+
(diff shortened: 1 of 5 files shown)


dlwh:
i should mention i know basically nothing about setup.py scripts, so it's all copy-paste and stuff

yifanmai (approved):

yifanmai:
Python packaging is surprisingly confusing. -_- I think it's okay to do what we're doing as long as we're consistent across the organisation projects.

PR #53, opened 2022-11-04
Forgetful Causal Masking and Alibi bias

### LINKED ISSUE #47: Add FCM to GPT-2 impl
should be compatible with HF export, which is an important feature for now

dlwh:
Implements [Forgetful Causal Masking](https://arxiv.org/abs/2210.13432) and [Alibi](https://arxiv.org/abs/2108.12409) attention biases.

I only integrated the former into GPT-2 because it only impacts training and not inference. Alibi is inference-time so would break HF model export.

Closes #47

### DIFF
--- a/src/haliax/core.py
+++ b/src/haliax/core.py
@@ -723,7 +723,7 @@ def _broadcast_order(a: NamedArray, b: NamedArray, require_subset: bool = True) 
         # maybe just add a context manager to allow it?
         raise ValueError(
             f"Cannot broadcast {a} and {b}: no subset relationship. "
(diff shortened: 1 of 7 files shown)


dlwh:
@yifanmai if you want to look at this one you're welcome to. I can also just merge ML-heavy PRs like this one. (I've been ninja-merging some fixups, but this one has actual logic)

yifanmai (approved):
The Python aspects look good. I'll take a closer look at the ML aspects later. Approving to unblock.

dlwh:
i can do work off of other branches, so it's no big deal

PR #55, opened 2022-11-07
Remove furl, merge checkpoint_test and test_checkpoint

dlwh:
furl ended up causing more trouble than it's worth. switched to os.path.join on @yifanmai 's recommendation.

Still need to test this against GCS so not ready to merge


yifanmai:
Thinking about `os.path.join` a bit more... this should work fine on Linux and local paths, Linux on cloud storage paths, and Windows on local paths... but it will break on Windows and cloud storage paths, but I presume that combination should never happen?

yifanmai:
Let's just use os.path.join for now and not support windows. Also, ow.

dlwh:
merging this since it's blocking and you already saw most of it

PR #56, opened 2022-11-07
add '/files/*' to git safe directory

dlwh:
This may be a little much, but... wandb wants to access the git repo to get the current sha for logging, but git doesn't want to do that on NFS/shared directories. git's `safe.directory` fixes that. I've been using it for the main levanter, but this adds it for everything in /files/

It seems like there ought to be a way to get a git sha safely without this, but i dunno.

### DIFF
--- a/scripts/setup-tpu-vm.sh
+++ b/scripts/setup-tpu-vm.sh
@@ -16,10 +16,13 @@ retCode=$?
 #sudo apt update
 #sudo apt upgrade -y
 
-# install python 3.10 and nfs
-sudo apt install -y software-properties-common
+# install python 3.10, latest git, and nfs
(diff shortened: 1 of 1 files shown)


yifanmai:
Does this mean that wand is getting git hashes of other git repos in /files/*? I'm confused why files outside /files/levanter would affect the git hash.

dlwh:
sorry: there are currently 3 checkouts/worktrees of levanter in `/files`: levanter, levanter32, levanter-lakhmidi I could just glob levanter* if you'd prefer.

yifanmai:
Ah ok, this sounds good.

yifanmai (approved):

PR #59, opened 2022-11-09
switch to using haliax activation functions 

dlwh:
so we can remove some grossness in GPT-2


dlwh:
this is boring so going to merge

PR #60, opened 2022-11-13
recompute total_tokens and total_flops so it works with resume

dlwh:


yifanmai:
Are `tokens_per_example` and `flops_per_example` constant across steps?

yifanmai (approved):

dlwh:
for GPT and most transformer model they are. As long as you're using constant length sequences and not cutting off the compute early with some kind of early exit transformer

PR #67, opened 2022-11-22
Expose length of datasets so we can start/stop

dlwh:
Builds on #64

previously we treated datasets as having unknown length, which was kinda problematic sometimes.

This changes behavior: in training, we have to explicitly loop forever, and we now skip the last little bit of shards that aren't the shortest shard. Previously we would skip the last little bit of every "batch" of docs written to the token cache

### DIFF
--- a/src/levanter/data/text.py
+++ b/src/levanter/data/text.py
@@ -61,9 +61,29 @@ class TokenSeqDataset(ShardableDataset[Sequence[int]]):
         return TokenSeqDataset(self.doc_cache.shard(shard_id, num_shards), self.seq_len, self.stride)
 
     def __iter__(self) -> Iterator[Sequence[int]]:
+        extra_tokens = None  # BatchEncoding of the last tokens from the previous doc
         for doc in self.doc_cache:
-            for encoded_slice in concatenate_and_group_texts(doc, self.seq_len, self.stride):
-                yield encoded_slice["input_ids"]
+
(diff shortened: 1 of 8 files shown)


src/levanter/data/text.py
```diff
@@ -147,28 +174,34 @@ def shard(self, shard_index, num_shards):
         if num_shards == 1:
             return self
 
-        return TokenizedDocumentCache(self.cache_dir, self.cache_files[shard_index::num_shards], self.flatten_docs)
-
-
-def _read_cache_file(file, flatten: bool = False) -> Iterator[BatchEncoding]:
```
dlwh, line 153: the diff sucks here. _read_cache_file is now just a method

tests/test_tokenized_document_cache.py
```diff
@@ -128,3 +128,36 @@ def doc_i(i: int):
             for i in range(len(reconstructed)):
                 as_listed = BatchEncoding(data={k: [vv.tolist() for vv in v] for k, v in reconstructed[i].items()})
                 assert as_listed == docs[i]
+
+
+def test_token_seq_dataset_len_is_correct():
+    # num_docs, seq_len, doc_length
+    configurations_to_test = [
```
dlwh, line 135: probably good to do hypothesis tests at some point

dlwh:
this one at least is ready i think

PR #68, opened 2022-11-28
JAX SMI for memory utilization, initial cut at CUDA docs

dlwh:
TPUs don't have an equivalent to nvidia-smi, and on GPU nvidia-smi is not useful for memory utilization because jax/xla reserves 90% up front.

I copy pasted code from https://github.com/ayaka14732/jax-smi, then did some nasty parsing to get the values out. (@yifanmai you might know of better ways of interacting with pprof output?)

Also includes a bit of docs for CUDA since I was doing debugging this as I was setting up CUDA.


yifanmai:
[jax.profiler.device_memory_profile](https://jax.readthedocs.io/en/latest/_autosummary/jax.profiler.device_memory_profile.html#jax.profiler.device_memory_profile) returns a serialized [`Profile` protobuf](https://github.com/google/pprof/blob/main/proto/profile.proto); you can call [`ParseFromString`](https://developers.google.com/protocol-buffers/docs/pythontutorial#parsing-and-serialization) to get the actual `Profile`.

There's some work involved with doing this - you'd have to compile the `profile.proto` with `protoc`, and you'd need to reimplement the aggregations that you want - but I think it's preferable than having to manage a Go environment.

dlwh:
OK yeah, thanks. I'm not super inclined to do it at the moment, but I agree
it would be better to do that.

dlwh:
tons more to do, but this is a decent checkpoint

PR #71, opened 2022-12-13
fix tests for jax 0.4.{0, 1)

dlwh:


tests/haliax/test_partitioning.py
```diff
@@ -59,21 +59,9 @@ def test_pjit_class_init():
             mod = named_pjit(MyModuleInit)()
 
         assert mod.named.array.shape == (Dim2.size, Dim3.size)
-        assert mod.named.array.sharding_spec.mesh_mapping == (
```
dlwh, line 62: this relied on internals and there's no good way to do it anymore AFAICT

dlwh:
merging to fix build

PR #101, opened 2023-03-31
Fix path and improve setup guide

Ivan-Zhou:
The primary purpose of this PR is to update the path of the setup script in readme (`infra/spin-up-vm.sh` instead of `infra/spin-up-tpu-vm.sh`). I also made some minor modification to the "Automatic Setup" section. 

The setup script is very useful 👍

### DIFF
--- a/docs/Getting-Started-TPU-VM.md
+++ b/docs/Getting-Started-TPU-VM.md
@@ -31,9 +31,11 @@ and you can't just set up one machine, but a whole cluster. We have some scripts
 
 ### Automatic Setup
 
-(CRFM folks: also see below)
+You can use `infra/spin-up-vm.sh` to create a TPU VM instance. In addition to creating the instance, it will set up
+the venv on each worker, and it will clone the repo to `~/levanter/`.
 
-You can use `infra/spin-up-tpu-vm.sh` to create a TPU VM instance:
+**For Public Users**:
+
 ```bash
 bash infra/spin-up-tpu-vm.sh <name> -z <zone> -t <type> [--preemptible]
 ```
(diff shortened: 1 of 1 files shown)


docs/Getting-Started-TPU-VM.md
```diff
@@ -43,25 +45,24 @@ Defaults are:
...
+bash infra/spin-up-vm.sh <name> -z <zone> -t <type> [--preemptible] -s infra/setup-tpu-vm-nfs.sh
 ```
 
 In addition to creating the instance, it will also mount the `/files/` nfs share to all workers, which has a good
 venv and a copy of the repo.
 
+**Notes**:
+- This uploads setup scripts via scp. If the ssh-key that you used for Google Cloud requires passphrase or your ssh key
+path is not `~/.ssh/google_compute_engine`, you will need to modify the script.
+- The command will spam you with a lot of output, sorry.
```
dlwh, line 61: we should probably add a -q flag or something. I am sorry about the spam

docs/Getting-Started-TPU-VM.md
```diff
@@ -43,25 +45,24 @@ Defaults are:
...
-####
-If you use a preemptible instance, you probably want to use the "babysitting" script that automatically re-creates
-the VM. That's explained down below in the "Running Levanter GPT-2" section.
-
+**For Stanford CRFM Users**:
 
-### CRFM Setup
-
-Stanford CRFM folks can pass a different setup script to `infra/spin-up-tpu-vm.sh` to get our NFS automounted:
+Stanford CRFM folks can pass a different setup script to `infra/spin-up-vm.sh` to get our NFS automounted:
```
dlwh, line 50: should we maybe rename the script? WDYT?

dlwh (approved):

dlwh:
thanks!

docs/Getting-Started-TPU-VM.md
Ivan-Zhou: The log is helpful for me to know what is going on and what failed. Maybe we can add an argument for showing/hiding the log :)

docs/Getting-Started-TPU-VM.md
Ivan-Zhou: Do you mean renaming the shell script back to `spin-up-tpu-vm.sh`? I agree that's more specific. What was your intention when you renaming it to `spin-up-vm.sh`?

Ivan-Zhou:
I'm going to merge this PR for now. 
1. For long output from the setup script, I don't think it is critical for now. We can address it in a future PR. 
2. For renaming of `spin-up-vm.sh`, I need a little bit more context, but I can definitely do it in another PR.

PR #102, opened 2023-04-04
Fix issues in Overview

Ivan-Zhou:
Fix a few minor issues in the overview doc

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -283,14 +283,13 @@ as far as I'm aware, and don't help with model partitioning.
 You can skip this part if you're familiar with the basics of how Transformers are implemented. You might want to skim at
 least the attention section to see how Haliax is used.
 
-The whole implementation is [here](https://www.github.com/stanford-crfm/levanter/blob/main/levanter/models/gpt2.py),
+The whole implementation is [here](../src/levanter/models/gpt2.py).
 (If you look at the whole thing, I caution you to skip over the torch serialization compatibility parts because they're
 messy and not that interesting for our purposes here.) I'll walk through the more interesting bits. In terms of basic
 structure and grouping of classes it's pretty similar to standard PyTorch implementations.
 
-XXX this comes across as even more defensive than I intended
-If you want, you can compare it with Andrei Karpathy's [mingpt implementation](https://github.com/karpathy/minGPT/blob/master/mingpt/model.py)
-. Ours is a bit longer (even excluding the torch serialization parts) for a few reasons:
+If you want, you can compare it with Andrej Karpathy's [minGPT implementation](https://github.com/karpathy/minGPT/blob/master/mingpt/model.py).
+Ours is a bit longer (even excluding the torch serialization parts) for a few reasons:
 * Boilerplate from declaring fields for modules
 * More type annotations
 * Tricks to improve stability from the [Mistral project](https://crfm.stanford.edu/2021/08/26/mistral.html#eureka/)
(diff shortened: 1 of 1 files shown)


docs/Overview.md
```diff
@@ -414,11 +413,11 @@ def __call__(self, hidden_states: NamedArray, layer_idx, inference: bool = True,
 If you're not used to the `tril`-as-a-mask trick, it's a way to create a causal mask so that the attention scores
 for a query can only attend to the past. The `tril` function creates a lower triangular matrix. It's equivalent to:
 ```python
-causal_mask = jnp.zeros(seq_len, key_seq_len)
+causal_mask = jnp.zeros((seq_len, key_seq_len))
 for i in range(seq_len):
     for j in range(key_seq_len):
-        if j >= i:
```
Ivan-Zhou, line 420: Since it is lower triangular matrix, I believe the sign should be `<=` instead of `>=`.

With `>= i:`, we get: 

```
[[1. 1. 1.]
 [0. 1. 1.]
 [0. 0. 1.]]
```

Instead, with `<= i`, we get:
```
[[1. 0. 0.]
 [1. 1. 0.]
 [1. 1. 1.]]
```

docs/Overview.md
```diff
@@ -414,11 +413,11 @@ def __call__(self, hidden_states: NamedArray, layer_idx, inference: bool = True,
...
-            mask[i, j] = 1
+        if j <= i:
```
Ivan-Zhou, line 419: Since it is lower triangular matrix, I believe the sign should be `<=` instead of `>=`.

With `>= i:`, we get: 

```
[[1. 1. 1.]
 [0. 1. 1.]
 [0. 0. 1.]]
```

Instead, with `<= i`, we get:
```
[[1. 0. 0.]
 [1. 1. 0.]
 [1. 1. 1.]]
```

docs/Overview.md
```diff
@@ -283,14 +283,13 @@ as far as I'm aware, and don't help with model partitioning.
 You can skip this part if you're familiar with the basics of how Transformers are implemented. You might want to skim at
 least the attention section to see how Haliax is used.
 
-The whole implementation is [here](https://www.github.com/stanford-crfm/levanter/blob/main/levanter/models/gpt2.py),
+The whole implementation is [here](../src/levanter/models/gpt2.py).
```
Ivan-Zhou, line 286: The URL is no longer valid. I changed it to the relative path instead.

docs/Overview.md
```diff
@@ -283,14 +283,13 @@ as far as I'm aware, and don't help with model partitioning.
...
 (If you look at the whole thing, I caution you to skip over the torch serialization compatibility parts because they're
 messy and not that interesting for our purposes here.) I'll walk through the more interesting bits. In terms of basic
 structure and grouping of classes it's pretty similar to standard PyTorch implementations.
 
-XXX this comes across as even more defensive than I intended
-If you want, you can compare it with Andrei Karpathy's [mingpt implementation](https://github.com/karpathy/minGPT/blob/master/mingpt/model.py)
-. Ours is a bit longer (even excluding the torch serialization parts) for a few reasons:
+If you want, you can compare it with Andrej Karpathy's [minGPT implementation](https://github.com/karpathy/minGPT/blob/master/mingpt/model.py).
```
Ivan-Zhou, line 291: Correct typo of Andrej's first name.

docs/Overview.md
Ivan-Zhou: Actually, I found a critical difference in implementation with minGPT:

In minGPT ([here](https://github.com/karpathy/minGPT/blob/master/mingpt/model.py#L63)), it sets cells to be `-inf` in which the tril matrix (self.bias) is 0

```python
self.register_buffer("bias", torch.tril(torch.ones(config.block_size, config.block_size)).view(1, 1, config.block_size, config.block_size))
att = att.masked_fill(self.bias[:,:,:T,:T] == 0, float('-inf'))
```

Whereas in Levantar's implementation, it sets `-1e9` in places where causal_mask is 1. 

```
causal_mask = hax.tril(hax.ones((self.SeqLen, self.KeySeqLen), dtype=bool), self.SeqLen, self.KeySeqLen)
attn_scores = hax.where(causal_mask, attn_scores, -1e9)
```

What do you think @dlwh ? Did I miss anything in the implementation?

Ivan-Zhou: Please skip my earlier comment on the comparison of tril between minGPT and levanter. I realized the nuance.

docs/Overview.md
dlwh: oops

dlwh (approved):

PR #111, opened 2023-04-11
Scan Module to hide "scan over layers" unpleasantness in gpt2

### LINKED ISSUE #105: Make a ScanLayers Module
vmap/scan/fold scares people when it's used for anything other than the batch dim. It's unfortunately necessary for compilation times etc.

> I found a vmap in the Gpt2Transformer code and it freaked me out – shouldn’t there be one big vmap around the use of the model in the code elsewhere or something? Now I’m worried I won’t put the vmaps where they should be.

We use `vmap` in GPT-2 initialization for so-called "scan layers" (q.v. https://docs.kidger.site/equinox/tricks/#improve-compilation-speed-with-scan-over-layers)

It'd be nice to hide this in a module: homongeneous stacks of layers are pretty common.

I have a 90% done branch for this. The big thing missing is just fixing GPT-2 torch export.

https://github.com/stanford-crfm/levanter/compare/scan_module?expand=1

dlwh:
Fixes #105.

A "Stacked" wraps another module and produces a "stacked" version of it, where an input is applied to each instance of the stacked module in sequence. This is useful for e.g. transformers where you have multiple instances of the same transformer block and the input is applied in a fold/for loop in sequence.

It's similar in spirit to an equinox.nn.Sequential (or torch.nn.Sequential), but it must be homogeneous. In Jax, this is much cheaper to compile than a sequential (or moral equivalent), because Jax compiles the module's method once, instead of unrolling the sequential and compiling everything as a giant graph. In Jax, this pattern is often called "scan layers" or "scan over layers".

### DIFF
--- a/src/haliax/jax_utils.py
+++ b/src/haliax/jax_utils.py
@@ -53,6 +53,28 @@ def filter_eval_shape(fun: Callable, *args, **kwargs):
     return eqx.combine(dynamic_out, static_out.value)
 
 
+def filter_checkpoint(fun: Callable, *, prevent_cse: bool = True, policy: Optional[Callable[..., bool]] = None):
+    """As `jax.checkpoint`, but allows any Python object as inputs and outputs"""
+
+    @ft.wraps(fun)
+    def _fn(_static, _dynamic):
+        _args, _kwargs = eqx.combine(_static, _dynamic)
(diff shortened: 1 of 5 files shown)


dlwh:
this is pretty in the weeds but might interest @Ivan-Zhou . Also happy to just merge.

Issue #99, opened 2023-03-29
On-The-Fly Caching (Tracking Issue)


dlwh:
Currently, our data preprocessing story is something like:

* Get a corpus (either hf datasets or, better, jsonl files)
* ideally shuffle the corpus (yourself)
* Run a single machine process to pretokenize the data into shards

The last step can take many hours to days, and is a pretty annoying bottleneck that hits users right at the start of using levanter. No good.

We need an alternative. In this issue, I focus on the last step. Dealing with the second is covered in #34.

# Roadmap

- [x] Ray Cluster for tokenizing across multiple machines, without support for simultaneous writing and reading #112 
- [x] introduce a coordinator for simultaneous reads and writes #112 
- [x] Add resumption for writing #112 
- [x] auto-discovery of cluster on Slurm/TPU (can maybe hijack Jax's stuff) #118
- [x] log metrics during preprocessing #117 
- [x] wire up on-the-fly tokenization during training #120
- [ ] data loading from the global order with perfect reproducibility #119

# Proposed Approach

(Design as of 2023-04-18)

## Goals

We want to support the following:
1) Deterministic batches, even for a changing number of readers (or writers). That is, for any cluster size
during training, we want the same batches to be generated in the same order.
2) Sharded reading and writing. We want to be able to read and write from multiple shards in parallel.
3) Simultaneous reading and writing of shards. We want to be able to start training while we are still building the cache.
4) Fast resumption without losing too much progress. This applies to both *writing* and *reading* the cache. That is, when we resume a training run, we want to finish producing the cache and also jump to the right place in the cache for reads.
5) (eventually) shuffling/random access
6) Takes advantage of the fact that we typically have idle, beefy CPUs on the machines where we're doing training
7) We want to be able to build the cache offline too.
8) We want to support batches that are composed of fragments of documents. In particular, we take a moving window of tokens from documents. This implies that the mapping from "documents" to "batches" is not 1:1, or easy to compute.
9) (eventually) ≈random access to tokens and not docs
10)  can handle a variable number of examples being generated per input doc

We want to support the following use cases:
1) We have a larger training dataset, and we want to draw samples from it more or less independently on a large number of machines.
   We don't really care about "epochs"/"passes", but we do want to be able to handle resumes and be deterministic. Ideally, each
   machine only reads from the chunks that it needs to read from.
2) We have a smaller validation dataset, and we want to do a single pass over it. We don't care about resuming, and it's ok if
[...]

PR #112, opened 2023-04-11
sharded reading/writing via Ray

### LINKED ISSUE #118: auto-discovery of Ray cluster on Slurm/TPU
in the case where there is no ray cluster already, we should auto-discover the cluster the same way that Jax does for Slurm and TPU. We should do this in cache_dataset and in on-the-fly

Part of #99

dlwh:
Switches to using Ray for writing the cache. Part of #99 , fixes #118.

Currently the guaranteed-determinism-for-changing-number-of-workers for reading isn't implemented, and the simultaneous streaming and writing isn't quite wired up, but this at least gets us to the new caching implementation which should allow us to use ray to do the cache building as an initial step.

Would also like to add a bit of monitoring (monitoring is in #127 )


src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+
+# What is the global ordering of data for epochal, unsharded reads? (for validation data)
+# When we want to do a single pass over the data, we don't cycle and we don't shuffle. We just read the data in order.
+
+# Coordination
+# We use Ray to coordinate writing. We instantiate one ray task per writer (typically one per machine), and each task
+# is given a list of shards to process. Each task then processes one chunk from each shard in round robin order.
+# When a chunk is finished, we write it to disk and notify that a new chunk is available. There is a single ray actor
+# that is responsible for handling which chunks are available and which are not. This actor exposes a method
+# to get a chunk by index
```
dlwh, line 99: this is what i was going for. I was starting to move to one task per shard and even each shard-task spawning off tasks per "chunk" (which would allow more parallelism after moving to defining chunk size in terms of number of rows)

yifanmai:
Ray on Slurm looks quite reasonable: https://docs.ray.io/en/latest/cluster/vms/user-guides/community/slurm.html

src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+            self._propagate_exceptions()
+
+    def _propagate_exceptions(self):
+        # send exceptions to all blocked readers, terminate all writers
+        for future in self.reader_promises.values():
+            future.set_exception(self._exception[0].with_traceback(self._exception[1].as_traceback()))
+        for task in self.writer_tasks:
+            ray.cancel(task)
+
+    async def get_chunk(self, global_index: int) -> Optional[ChunkMetadata]:
```
yifanmai, line 503: Whoa, I didn't know that [async works with Ray actors](https://docs.ray.io/en/master/ray-core/actors/async_api.html#asyncio-for-actors). That's pretty cool.

src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+        # TODO: probably better to read from a global list of finished chunks
+        i = 0
+        while i < len(writers):
+            writer = writers[i]
+            try:
+                for chunk in writer.written_chunk_metadata:
+                    manager.finished_chunk(writer.shard_name, chunk)
+
+                if writer.is_finished:
+                    manager.finished_shard(writer.shard_name)
```
yifanmai, line 573: An architectural suggestion would be to do all the source / writer / seek position state management in the actor, and have the remote task just be the write itself i.e. have the calls be unidirectional rather than having the manager and writer call each other. This makes it easier to reason about things like fault tolerance and recovery i.e. we don't have worry about the manager and writer states desynchronizing.

Then you'd have a manager main loop that looks like: wait for any remote write task to complete with [`concurrent.futures.wait(return_when=FIRST_COMPLETED)`](https://docs.python.org/3.3/library/concurrent.futures.html#concurrent.futures.wait), update the source and chunk state and kick off a new remote write task.

src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+    def __init__(self):
+        self.buffered_chunks = []
+        self.finished = False
+
+    def ready(self) -> bool:
+        return len(self.buffered_chunks) > 0 or self.finished
+
+
+@ray.remote
+class _ShardCacheManager(Generic[T]):
```
yifanmai, line 414: We _may_ want actor fault tolerance, with would require us to either (1) save actor checkpoints or (2) make the actor be able to rebuild the chunk / shard state from the external disk state upon initalization. See [here](https://docs.ray.io/en/master/ray-core/actors.html#fault-tolerance) and [here](https://docs.ray.io/en/master/ray-core/fault_tolerance/actors.html#actor-fault-tolerance) in the Ray docs.

But it might also be simpler to just crash the whole job and restart it.

src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+import ray
+import tblib
+from dataclasses_json import dataclass_json
+from tqdm import tqdm
+
+from levanter.data import Dataset
+
+
+# We want to support the following:
+#  1) Deterministic ordering over the data, even for a changing number of readers or writers
```
yifanmai, line 21: General feedback regarding the documentation and the design doc is that there is some ambiguous verbiage; it would be useful to have a definitions section for the terms, and then use those terms consistently. In particular:

- "Input" referring to the input to the preprocessors versus to the model trainer (I suppose the term "preprocess" could also be ambiguous because the model could do its own preprocessing...)
- "Readers" could refer to the preprocessors or the model trainer
- "Resumption" could refer to the trainer or the preprocessor (since they are now decoupled; the preprocessor can write chunks faster than the trainer can read them)
- "Shards" (which refers to the input to the preprocessor) and "chunks" (which refer to the input to the trainer)

src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+# deterministic
+# 2) We have a smaller validation dataset, and we want to do a single pass over it. We don't care about resuming.
+# 3) Like (1) but we want to jump around the dataset. We still care about resuming and determinism, but don't
+# care about epochs.
+#
+#
+# We focus on (1) and (2) for now.
+
+# What does our input look like?
+# We have a smallish number of input shards (though usually >= num machines), each of which is an iterator over documents.
```
yifanmai, line 40: Does this assume the shards are uniform i.e. that we have a single Hugging Face dataset, or Parquet / JSONL file glob? Would this support concatenating multiple datasets?

src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+# so we round-robin over a slice of the chunk_iterators.
+
+# For other cases (R and R* don't divide each other), there's no simple relationship between the reader and chunk iterators
+# and you end up reading from everywhere, but that's ok.
+
+# What is the global ordering of data for epochal, unsharded reads? (for validation data)
+# When we want to do a single pass over the data, we don't cycle and we don't shuffle. We just read the data in order.
+
+# Coordination
+# We use Ray to coordinate writing. We instantiate one ray task per writer (typically one per machine), and each task
```
yifanmai, line 95: We should just have Ray tasks > machines and let the Ray scheduler handle machine placement for us. This incurs more Ray scheduling overhead, but we can potentially get better task placement across workers in return..

src/levanter/data/shardcache.py
```diff
@@ -0,0 +1,712 @@
...
+# is given a list of shards to process. Each task then processes one chunk from each shard in round robin order.
```
yifanmai, line 96: Also I think each chunk should be a separate task... I have a longer explanation later.

dlwh:
> Ray on Slurm looks quite reasonable: https://docs.ray.io/en/latest/cluster/vms/user-guides/community/slurm.html

yeah i think i looked at this when I was evaluating and I think it what decided me on ray

src/levanter/data/shardcache.py
dlwh: atm i'm assuming either a single HF dataset, or a list of jsonl globs

dlwh: for multiple datasets, I think the right thing to do is cache them separately and do mixing on the fly, since that's a thing one might want to play around with

docs/design/Data-Loader-Design.md
```diff
@@ -1,3 +1,168 @@
...
+changes. We also want readers to only read from the chunks that they need to read from.
+We pretend the list of data is infinite by cycling. We cannot track epochs.
+
+NB Our goal is a deterministic ordering over examples, and not merely chunks or even documents.
+
+Given a list of chunks and the idealized number of readers R*, we define the global ordering over chunks as follows:
+First define R* iterators over chunks, with `chunk_iterators[r]` being defined as `loop(all_chunks)[r::R*]`.
+
+Next, define a function `mk_examples(chunk_iterator)` that takes a list of iterators over chunks and returns
+a list of examples. Define `chunk_examples[r] = mk_examples(chunk_examples[r])`.
```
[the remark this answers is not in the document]
dlwh: i do, thank you

docs/design/Data-Loader-Design.md
```diff
@@ -1,3 +1,168 @@
...
+`chunk_examples[0][0], chunk_examples[1][0], ..., chunk_examples[R*-1][0], ..., chunk_examples[0][1], chunk_examples[1][1], ..., chunk_examples[R*-1][1], ...`
+that is, `example[i] == chunk_examples[i % R*][i // R*]`
+
+If we have $R*$ readers, then each `reader_iterator[r][j] == chunk_examples[r][j] == example[j * R* + r]`.
+Moreover, if either R or R* is a multiple of the other, then we still get a nice property where
+each reader reads from a strided slice of the chunk_iterators:
+
+(Boring math)
+If we have R readers, then `reader_iterator[r][j] == example[j * R + r] == chunk_examples[(j * R + r) % R*][(j * R + r) // R*]`
+If we have `R == n * R*`, then `reader_iterator[r][j] == example[j * R + r] == chunk_examples[(j * R + r) % R*][(j * R + r) // R*]
```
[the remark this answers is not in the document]
dlwh: good point, thanks

Ivan-Zhou (commented):
Still in progress of reading changes in shard_cache.py and text.py. The design looks good to me.

yifanmai (approved):
Looks good!

Btw, is there a test for global chunk ordering?

src/levanter/data/shard_cache.py
```diff
@@ -0,0 +1,564 @@
...
+T_co = TypeVar("T_co", covariant=True)
+_ExcInfo = Tuple[Optional[BaseException], tblib.Traceback]
+
+logger = logging.getLogger(__name__)
+
+ROWS_PER_CHUNK = 32 * 1024  # if a doc produces ~1200 tokens, this is ~150MB chunks
+LEDGER_FILE_NAME = "cache_ledger.json"
+
+
+class BatchProcessor(Generic[T_contra], ABC):  # type: ignore
```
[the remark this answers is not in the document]
dlwh: mypy thought that T couldn't be contravariant, which is plainly wrong.

config/data/openwebtext_source.yaml
```diff
@@ -1,5 +1,2 @@
-train_urls:
-  - "gs://pubmed-mosaic/openwebtext-sharded/openwebtext_train.{1..128}-of-128.jsonl.gz"
-validation_urls:
-  - "gs://pubmed-mosaic/openwebtext-sharded/openwebtext_val.{1..8}-of-8.jsonl.gz"
-cache_dir: "gs://pubmed-mosaic/tokenized/openwebtext/"
+id: openwebtext
+cache_dir: "zzz_cache"
```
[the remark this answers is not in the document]
dlwh: nope, thanks

src/levanter/data/shard_cache.py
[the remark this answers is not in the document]
dlwh: now it's fine with it, i dunno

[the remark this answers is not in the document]
dlwh: for a while BatchProcessor was a Protocol, which might have been the issue

tests/test_tokenized_document_cache.py
```diff
@@ -1,96 +1,108 @@
...
+def teardown_module(module):
+    ray.shutdown()
+
 
 def test_index_empty_file():
     with tempfile.TemporaryDirectory() as tmpdir:
-        empty_dataset = [{"text": ""}]
-        tokenized = [tokenizer(x["text"]) for x in empty_dataset]
-        cache = TokenizedDocumentCache.build_or_load(tokenized, f"{tmpdir}/cache", num_shards=1, flatten_docs=False)
+        empty_dataset = [""]
```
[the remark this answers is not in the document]
dlwh: done. no issues

PR #124, opened 2023-04-26
Update configuration for cosine_decay_schedule

Ivan-Zhou:
Minor change to the configuration for `optax.cosine_decay_schedule`:
- decay_steps: the value currently is self.num_train_steps - warmup_steps - warmup_steps. The third term seems redundant. 
- alpha should be just `self.min_lr_ratio` (right now it is `self.learning_rate * self.min_lr_ratio / self.learning_rate`)


dlwh (approved):

dlwh:
thanks!

PR #128, opened 2023-04-28
Allow "string selectors" where possible in Haliax

dlwh:
Also fixes a hidden bug with gradient accumulation due to manipulation of the batch dim...


dlwh:
gpt2_example produces bitwise identical results

PR #129, opened 2023-04-28
Factor out "next_token_loss" and use it everywhere

dlwh:
Depends on #128 

Introduces ReductionFn as well for aggregation or not as appropriate


dlwh:
verified bitwise identical

PR #130, opened 2023-04-28
Simplify fsdp specification

dlwh:
previously I had users manually specify the mapping from logical to physical axes (for data parallel and model parallelism), but in practice it's always the same thing. So I made it simpler, with defaults so you mostly don't need to specify it.


Ivan-Zhou (approved):
LGTM

PR #135, opened 2023-05-06
Add documentation for training config

Ivan-Zhou:
WIP -- want to share the PR out to make sure we are aligned on the format and structure.


dlwh:
looks great, thank you!

docs/Training.md
```diff
@@ -0,0 +1,103 @@
...
+```
+python levanter/examples/gpt2_example.py \
+    --config_path config/gpt2_small.yaml \
+    --trainer.load_checkpoint_path checkpoints/gpt2/wandb_run_name  \
+    --trainer.checkpointer.base_path checkpoints/gpt2/ \
+    --trainer.checkpointer.save_interval 20m
+```
+
+This will overwrite the default checkpoint settings from the `TrainerConfig` and `CheckpointerConfig` in [config.py](src/levanter/config.py) to 
+save checkpoints every 20 minutes. The checkpoint will be saved to the directory `checkpoints/gpt2/` with the WandB name `wandb_run_name`.
```
dlwh, line 53: just flagging we'll need to change when we switch to run id

docs/Training.md
```diff
@@ -0,0 +1,103 @@
...
+
+Note that:
+- The `--trainer.load_checkpoint_path` argument is optional. You only need to specify it if you want to load a checkpoint from a previous 
+run. If you do not specify it, the trainer will start training from scratch.
```
dlwh, line 57: maybe mention that it will recursively search for the latest checkpoint inside that directory

dlwh (approved):

PR #144, opened 2023-05-14
Preprocess and Caching RedPajama

Ivan-Zhou:
1. Add data configs for RedPajama 1B and 1T. 
2. Update data and training pipeline to accommodate the fact that RedPajama doesn't have val set.


scripts/cache_dataset.py
```diff
@@ -25,10 +25,15 @@ def main(args: RayCachedLMDatasetConfig):
...
         # connect or start the actor
         batch_tokenizer = BatchTokenizer(tokenizer)
-        source = args.get_shard_source(split)
-
-        cache_dataset(os.path.join(args.cache_dir, split), source, batch_tokenizer)
-        print(f"Finished caching {split} to {args.cache_dir}/{split}.")
+        try:
+            source = args.get_shard_source(split)
+        except ValueError as e:
+            print(f"Skipping {split} because it doesn't exist: {e}")
```
dlwh, line 31: probably we should have has_split(x) on LMDatasetConfig

dlwh (approved):

scripts/cache_dataset.py
Ivan-Zhou: Done

PR #150, opened 2023-05-18
Separate initialization of nn Modules from ctor

dlwh:
This has been driving me crazy for a while. Equinox's idiom is to do initialization in the constructor, but if you need to do anything fancy you're stuck doing extra fancy things with jax pytree manipulation, which isn't great


ASEM000:
I have a [library](https://github.com/ASEM000/PyTreeClass) that attaches lens-like setters/getters to pytrees modules like Equinox.
Why don't you try it and tell me if it solves your problem? (This should work starting from Equinox 0.10.4).

```python
import equinox as eqx
import pytreeclass as pytc
import jax

class Tree(eqx.Module):
    weight: jax.Array = jax.numpy.array([-1, 2, 3])
    bias: jax.Array = jax.numpy.array([1])
    counter: int = 1

    @property
    def at(self):
        return pytc.AtIndexer(self, ())


tree = Tree()

tree = (
    tree.at["counter"]
    .set(1)  # set counter to 1
    .at[jax.tree_map(lambda x: x < 0, tree)]
    .set(0)  # set negative values to 0
    .at["bias"].set(100)  # set bias to 100
)

print(tree.weight)
# [0 2 3]
print(tree.bias)
# 100
print(tree.counter)
# 1

```

P.S., love the named array implementation.

dlwh:
Hi @ASEM000,

That's very cool and I'm very sympathetic to using lenses in general. For this deserialization case where I'm basically reconstructing the entire tree bottom up, I think just exposing normal dataclass constructors is the right thing...

I will check the library out soon though!

dlwh:
verified bitwise reproduction of current main [gallant-monkey-1767](https://wandb.ai/stanford-mercury/levanter/runs/4aof5q16)

PR #153, opened 2023-05-22
Add rows_per_chunk as a config

Ivan-Zhou:
Add `rows_per_chunk` as a dataset config, so that users can set their custom `rows_per_chunk` variable if needed. 
- [ ] This change passed the unit tests
- [x] This change has been tested on real dataset
CleanShot 2023-05-23 at 19 54 12

### DIFF
--- a/src/levanter/data/text.py
+++ b/src/levanter/data/text.py
@@ -18,7 +18,6 @@ from pyrallis import field
 
 # intercept the logging nonsense here
 from levanter.logging import silence_transformer_nag  # noqa
-from levanter.data.utils import safe_enumerate_iterable
 
 silence_transformer_nag()  # noqa
 from transformers import BatchEncoding, PreTrainedTokenizerBase, PreTrainedTokenizerFast  # noqa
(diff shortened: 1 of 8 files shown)


src/levanter/data/text.py
```diff
@@ -522,7 +523,7 @@ def open_shard_at_row(self, shard_name: str, row: int) -> Iterator[str]:
             shard = dataset
 
         idx = 0
-        for _, doc in shard:
+        for _, doc in safe_enumerate_iterable(shard):
```
dlwh, line 526: i think i don't like this.  is this to suppress the http error? If so, and those failures aren't deterministic, then this breaks reproducibility, which is important for this project

 i'd rather crash if the underlying data source is broken.

src/levanter/data/text.py
```diff
@@ -424,6 +424,7 @@ class LMDatasetConfig:
     enforce_eos: bool = True  # whether to append eos even if the tokenizer doesn't
 
     splits: List[str] = field(default_factory=lambda: ["train", "validation"])
+    rows_per_chunk: int = 2048 # number of rows to process and cache per chunk
```
dlwh, line 427: should this default be DEFAULT_ROWS_PER_CHUNK?

dlwh (changes requested):

dlwh (approved):

PR #154, opened 2023-05-22
rename named_pjit to named_jit, make in_axis_resources not default

dlwh:
rename named_pjit to named_jit, since Jax is going that way


dlwh:
Verified bitwise reproducible https://wandb.ai/stanford-mercury/levanter/runs/mpebo0im?workspace=user-dlwh

dlwh:
verified bitwise reproducible in https://wandb.ai/stanford-mercury/levanter/runs/09hr28o7/overview?workspace=user-dlwh

PR #156, opened 2023-05-23
Add Backpack Networks

Ivan-Zhou:
This PR adds the [backpack network](https://backpackmodels.science/) to Levanter codebase:
1. The main change is in `src/levanter/models/backpack.py`. It includes the building blocks of `BackpackLMHeadModel`. 
2. It includes `examples/backpack_example.py`, which is the execution of the training of Backpack. It is very much redundant to `gpt2_example.py`. I will try to clean them up. 
3. It includes two test files: a unit test to test loading backpack and run inference; a test to check consistency between levanter GPT/Backpack models vs. corresponding HF models. 

Action items:
- [x] Update code to reflect the latest changes in Levanter (up to May 28th) 
- [x] The unit tests have passed
- [x] Launch training with `examples/backpack_example.py` 
- [ ] Reproduce training jobs from levanter-backpacks repo
- [ ] [Maybe] Consolidate `examples/backpack_example.py` into `gpt2_example.py`


percyliang (commented):
In general, I wonder how much code can be shared between different architectures?  In theory, only the model definition should change and even the model has a lot of shared components (after all, part of the point of Backpacks is that it builds on Transformers as a blackbox), but it seems like training code should be the same?  Of course, in practice, I know things are more complex, but just wanted to get a sense of what we're thinking on this.

Also, given the pedagogical goals of levanter, does it make sense to have a really small core and everything else in `contrib` or something?

examples/backpack_example.py
```diff
@@ -0,0 +1,293 @@
...
+from levanter.trainer_hooks import StepInfo, TrainerHooks
+from levanter.utils.jax_utils import global_key_array, parameter_count
+from levanter.utils.py_utils import non_caching_cycle
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class TrainBackpackConfig:
```
dlwh, line 35: other than this class, is it total copypaste? (fine if so, just won't look as carefully)

tests/test_levanter_hf_consistency.py
```diff
@@ -0,0 +1,104 @@
...
+from test_utils import skip_if_no_torch
+
+
+HF_BACKPACK = "ivanzhouyq/levanter-backpacks"
+LEVANTER_BACKPACK_CHECKPOINT = "gs://levanter-checkpoints/backpacks/backpack_170M_100k_steps_run_0424-new/step-100000/"
+HF_GPT2 = "ivanzhouyq/levanter-gpt2-small"
+LEVANTER_GPT2_CHECKPOINT = "gs://levanter-checkpoints/dev/clean-snowball-990-new/step-10000"
+
+
+@skip_if_no_torch
```
dlwh, line 26: let's also skip this test if the checkpoint isn't available b/c it's a non-crfm person running it

dlwh: actually better still would be a version that runs like the gpt2 test that tests hf == hf->levanter import == levanter->hf export

dlwh (approved):
a few cleanup requests but otherwise lgtm!

examples/backpack_example.py
Ivan-Zhou: I pasted it into gpt2_example.py. It is mostly copypaste. We definitely merge these two files, to save cost of maintenance in the future. I might do it in a second PR.

src/levanter/compat/hf_checkpoints.py
```diff
@@ -33,6 +36,15 @@
 SAFE_TENSORS_MODEL = "model.safetensors"
 
 
+@dataclass
+class HFAutoMapConfig:
```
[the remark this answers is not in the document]
Ivan-Zhou: This is needed for conversion to HF model, but I didn't add that code yet.

src/levanter/compat/hf_checkpoints.py
```diff
@@ -159,6 +213,44 @@ def get_to_cpu(arr: Union[jnp.ndarray, np.ndarray]):
         safetensors.numpy.save_file(state_dict, f"{path}/{SAFE_TENSORS_MODEL}", metadata={"format": "pt"})
 
 
+def _save_hf_checkpoint_local(
```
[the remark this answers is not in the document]
Ivan-Zhou: That's the purpose of this generic name, but I haven't merged two function yet...

src/levanter/models/gpt2.py
```diff
@@ -277,8 +277,11 @@ def init(Vocab: Axis, config: Gpt2Config, *, key) -> "Gpt2Embeddings":
         return Gpt2Embeddings(Vocab, config, token_embeddings, position_embeddings, dropout)
 
     @named_call
-    def embed(self, input_ids, inference, *, key):
+    def embed(self, input_ids, inference, *, key, input_embeds_only=False):
```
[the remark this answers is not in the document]
Ivan-Zhou: Sure

tests/test_levanter_hf_consistency.py
Ivan-Zhou: will add

PR #161, opened 2023-05-25
WIP Better announcement post

dlwh:
Not done yet but it's coming along.


docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
+# ---
+layout: blog
+title: Levanter — Towards Scaling Reproducible Foundational Models with Jax
```
percyliang, line 3: How about: Levanter -- Scalable, Reproducible, Legigible Foundation Models with Jax?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+        and infrastructure for training reproducible, legible foundation models models using Jax. We also XXX checkpoints.
+    </strong>
+> </div>
+
+
+## Introduction
+
+The growth of artificial intelligence and machine learning has brought about the need for scalable and reproducible models.
+To address this, at the [Center for Research on Foundation Models](https://crfm.stanford.edu), we have created two new tools — [Levanter and Haliax](https://github.com/stanford-crfm/levanter).
+They form a new code base for training foundational models with the promise of flexibility, modularity, efficiency,
```
percyliang, line 23: foundation

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+    <strong> We introduce our projects
+        <a href="https://github.com/stanford-mercury/levanter" target="_blank">Haliax and Levanter</a>, our code
+        and infrastructure for training reproducible, legible foundation models models using Jax. We also XXX checkpoints.
+    </strong>
+> </div>
+
+
+## Introduction
+
+The growth of artificial intelligence and machine learning has brought about the need for scalable and reproducible models.
```
percyliang, line 21: Seems like we want stronger justification for reproducible, or just punt on the more detailed positioning (legibility, reproducibility, scalability) until later.

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+To address this, at the [Center for Research on Foundation Models](https://crfm.stanford.edu), we have created two new tools — [Levanter and Haliax](https://github.com/stanford-crfm/levanter).
+They form a new code base for training foundational models with the promise of flexibility, modularity, efficiency,
+and scale, as well as strong guarantees about reproducibility.
+
+Levanter is a [Jax](https://github.com/google/jax)-based codebase for training foundation models that is designed to be
```
percyliang, line 26: Make the three pillars explicit with bullet points early?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+hope that these libraries will be useful to the community, and we look forward to seeing what people do with them.
+
+## The Landscape of Foundation Model Training Frameworks
+
+Numerous foundation model training frameworks exist in the community, each with its strengths and focus.
+For large language models (LLMs) (the focus of this release),
+the most well-known in the open source community is probably NVIDIA's PyTorch-based [Megatron-LM](https://github.com/NVIDIA/Megatron-LM),
+and its many derivatives, including EleutherAI's [GPT-NeoX](https://github.com/EleutherAI/gpt-neox) codebase. MosaicML
+has recently released [LLM Foundry](https://github.com/mosaicml/llm-foundry), which we used to train
+[BioMedLM](https://crfm.stanford.edu/2022/12/15/biomedlm.html) last autumn.
```
percyliang, line 43: Anything else?  HF / mistral?  metaseq?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+released the Haiku-based [Jaxformer](https://github.com/salesforce/jaxformer).
+
+Despite the wide array of existing frameworks, when we started, we found that none of them fully addressed our needs.
+At CRFM, we wanted to achieve four goals:
+
+* **Legibility**: We wanted to be able to write code that was easy to read and understand, and that could be composed easily.
+* **Reproducibility**: We wanted to be able to reproduce our results exactly, even in the presence of preemption and
+  restarting from checkpoints.
+* **Scalability**: We wanted to be able to fully utilize the Cloud TPU resources we were given as part of Google's TPU Research Cloud program, as well as our own GPU resources.
+* **Efficiency**: We wanted to achieve the other three goals without sacrificing (much) efficiency.
```
percyliang, line 58: Can these be combined into one?  Rule of 3s...

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+
+We chose [Jax](https://github.com/google/jax/) as our framework because it is a powerful, flexible, and performant,
+and offers strong reproducibility guarantees. Jax works well on TPUs, while we found that PyTorch support was still uneven.
+Jax is also a natural choice because it allows you to focus on the "what" of your code, and not on the "how:" details of
```
percyliang, line 62: "how":

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+partitioning and communication can be left to the XLA compiler. Finally, Jax makes reproducibility easy, since it uses
+bitwise deterministic PRNGs by default, with careful control over the PRNG state.
+
+However, Jax is a low-level framework, and we found that, by itself, it did not provide the legibility that we wanted.
+
+For us, legibility is a top concern. We therefore created two new libraries: **Haliax** and **Levanter**. Haliax is a
```
percyliang, line 68: Maybe should say relationship of haliax to flax?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+named tensor library that focuses on improving the legibility and compositionality of deep learning code while still
+maintaining efficiency and scalability. Levanter is a library for training foundation models built on top of Haliax. In
+addition to the goals of legibility, efficiency, and scalability, Levanter further strives for bitwise reproducibility,
+meaning that the same code with the same data will produce the exact same result, even in the presence of preemption and
+restarts.
+
+## Haliax: Legibility via Named Tensors
+
+Haliax is a Jax library for building neural networks with named tensors, built on Jax and [Equinox](https://github.com/patrick-kidger/equinox),
```
percyliang, line 77: Seems like named tensors should be independent of neural networks?  Is there another layer of abstraction missing?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+Pos = hax.Axis("position", 1024)  # sequence
+KPos = Pos.alias("key_position")  # key sequence for attention
+Head = hax.Axis("head", 8)  # number of attention heads
+Key = hax.Axis("key", 64)  # key size
+Embed = hax.Axis("embed", 512)  # embedding size
+
+# Despite making no reference to batching or heads, this same implementation is also batch-capable and multi-headed
+# (or multi-query) and even supports attending to or from non-sequential keys (e.g. image patches)
+def attention(Key, KPos, query, key, value, mask):
+  # how similar is each query to each key
```
percyliang, line 112: Four space indent?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+
+* Named axes are more semantically meaningful and rely less on bitrot-prone comments.
+* Broadcasting leads to unreadable strings of `view`s and `squeeze`s. (To this, I'd add implicit broadcasting is a source of bugs.)
+* Named axes allow you to abstract over unreferenced dimensions, making code more flexible for things like multi-headed attention or flexible attention masking.
+
+Let's take a look at a practical example of how Haliax can be used.
+Here's a minimal attention module implementation in Haliax. For a more detailed introduction,
+please see the [Haliax tutorial](https://colab.research.google.com/drive/1TiTcQQ4V5mopbgCu1SVl-oqJtXn7rFnC).
+
+```python
```
percyliang, line 94: I wonder if this is too much code, and there's smaller (1-3 line) snippets that one can put to highlight particular aspects like the named axes...can put '...' for intervening code.

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+
+    answers = attention(Key, KPos, q, k, v, causal_mask)
+
+    x = self.proj_answer(answers)
+    return x
+```
+
+We use named axes both to improve legibility and to enable scale: named axes are the basis of our
+[Fully-Sharded Data Parallel](https://engineering.fb.com/2021/07/15/open-source/fsdp/) implementation. FSDP can
+be added to the training code with about 10 lines of code, enabling scale to at least 256 TPU cores (which is
```
percyliang, line 160: Would be cool to show those 10 lines

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+as many as we can get our hands on) and at least 65b parameters (which is way bigger than we have compute for).
+
+### Haliax Tutorials
+
+For more details, please see our interactive tutorials on Colab:
+
+* [Introduction to Haliax with Transformers](https://colab.research.google.com/drive/1TiTcQQ4V5mopbgCu1SVl-oqJtXn7rFnC?usp=sharing)
+* [Scaling Transformers in Haliax](https://colab.research.google.com/drive/1QX4yH3zRFF3Xiibf1aahETcSQ5nbcUMz?usp=sharing), including FSDP in Jax.
+
+## Levanter: Bitwise Determinism with Jax
```
percyliang, line 170: Seems a bit weird - breaks the flow a bit...it seems like haliax/levanter divide is incompatible with the 3 pillars (I'd say the latter is more important)?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+for training a GPT-2-like Transformer, complete with data preparation, logging, training, checkpointing, evaluation, and export,
+while maintaining bitwise reproducibility throughout.
+
+We have used Levanter to train models as large as 6.7b parameters on a v3-256, and have run experiments showing that it can scale
+up to least 65b parameters.
+
+### Bitwise Reproducibility
+
+One of the benefits of Jax is that it offers strong guarantees for reproducibility. In particular, Jax's fine-grained
+control over random number generation makes it easy to ensure bitwise reproducibility, especially when using TPUs.
```
percyliang, line 182: "makes it easy" - can we strengthen this?

percyliang: Can we just state clearly what the guarantee is?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+here is a screenshot of a training run being resumed multiple times, even on different TPU pod slices:
+
+![plot showing bitwise reproducibility with four training runs superimposed with the exact same loss curve](figures/bitwise_repro_curve.png)
+
+The fact that you can't make out the different lines is the point: the training runs are bitwise identical,
+a huge advantage for debugging and reproducibility.
+
+### Efficiency and Scale
+
+XXX something something v3-256 scaling numbers?
```
percyliang, line 193: Also utilization, and things you did to make it performant

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Other Features
+
+* **Preprocessing**: Levanter uses [Hugging Face Tokenizers](https://github.com/huggingface/tokenizers/) to preprocess text
```
percyliang, line 197: Can we advertise the upshot a bit more?  It can stream datasets and start training right away!

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+using distributed preprocessing backed by [Ray](https://www.ray.io/))
+* **Training**: In addition to Jax and Haliax, Levanter uses [Optax](https://github.com/deepmind/optax) for optimization,
```
percyliang, line 199: Can Sophia not be released as part of Levanter?

docs/Blog-Post.md
```diff
@@ -0,0 +1,276 @@
...
+Here is an example of the token probability visualization in action on a small, quick training run:
+
+![video showing heat map of token probabilities for a sample of the validation set evolving as training progresses](figures/token_probabilities.mov)
+
+The darker, more purple the color, the lower the probability of the token. The lighter, more yellow the color, the higher the probability.
+This visualization is logged to WandB as training progresses and can be viewed interactively. I have found this to be a
+nice alternative to just staring obsessively at the loss curve, which is what I usually do.
+
+
+#### On-Demand Preprocessing
```
percyliang, line 226: on-demand data preprocessing?

docs/Blog-Post.md
dlwh: i had an old version that had a "too many to list" comment but I left it out of this draft. i added the ones you mentioned plus a couple others.

docs/Blog-Post.md
dlwh: i'll try...

docs/Blog-Post.md
dlwh: to Equinox? It has no relationship to Flax.

docs/Blog-Post.md
dlwh: I went back and forth on this. The things in Haliax are: named tensor, NN Modules that work with named tensors, and versions of jax primitives that transform nad distribute named tensors and module hierachies thereof. I think it's a pretty natural grouping. The NN part is pretty small.

docs/Blog-Post.md
dlwh: i use four in general, but colab defaults to two and this is copy paste from colab.

docs/Blog-Post.md
dlwh: it's 10 lines of code that are distributed across 5-6 places :-(, they mostly look the same though

docs/Blog-Post.md
dlwh: Sophia-G isn't quite working in Jax yet. Still figuring it out.

docs/Blog-Post.md
dlwh: think i could just drop the module and just do the axes + attention function?

docs/Blog-Post.md
percyliang: I meant more as a related work since it seems like people will probably think of jax/flax, and we have jax/equinox/haliax...

docs/Blog-Post.md
dlwh: (turns out it was just a bad gamma number. Hong is training this)

docs/Blog-Post.md
dlwh: ok, still struggling with this part, and there are a few more todos, but RFAL

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
+# ---
```
percyliang, line 1: Assuming there will be multiple blog posts in the future, perhaps we should rename this file?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+To address this, at the [Center for Research on Foundation Models](https://crfm.stanford.edu), we have created two new tools — [Levanter and Haliax](https://github.com/stanford-crfm/levanter).
+They form a new code base for training foundation models with the promise of flexibility, modularity, efficiency,
+and scale, as well as strong guarantees about reproducibility.
+-->
+
+XXX Positioning?
+
+Today, we are excited to announce the release of two new libraries for training foundation models: [Levanter and Haliax](https://github.com/stanford-crfm/levanter).
+Levanter is a [JAX](https://github.com/google/jax)-based codebase for training foundation models that is designed to be
+flexible, modular, and accessible, while still being performant and scalable. In particular, Levanter is designed to be
```
percyliang, line 35: Can we just use the three terms (legible, scalable, and reproducible) consistently?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+able to train models on a variety of different hardware, including GPUs, TPUs, and TPU pods.
```
percyliang, line 36: Simplify: "including TPUs and GPUs"?  Maybe lightly suggest it's better for TPUs?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+
+The growth of artificial intelligence and machine learning has brought about the need for scalable and reproducible models.
+To address this, at the [Center for Research on Foundation Models](https://crfm.stanford.edu), we have created two new tools — [Levanter and Haliax](https://github.com/stanford-crfm/levanter).
+They form a new code base for training foundation models with the promise of flexibility, modularity, efficiency,
+and scale, as well as strong guarantees about reproducibility.
+-->
+
+XXX Positioning?
+
+Today, we are excited to announce the release of two new libraries for training foundation models: [Levanter and Haliax](https://github.com/stanford-crfm/levanter).
```
percyliang, line 33: I think cleaner to just say Levanter

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+Levanter is a [JAX](https://github.com/google/jax)-based codebase for training foundation models that is designed to be
+flexible, modular, and accessible, while still being performant and scalable. In particular, Levanter is designed to be
+able to train models on a variety of different hardware, including GPUs, TPUs, and TPU pods.
+Levanter also offers strong guarantees about reproducibility.
+Haliax is a named tensor library that we developed to make it easier to write legible, composable code while
+still maintaining efficiency and scalability.
+
+Today, we're releasing the first version of Levanter and Haliax, along with several models trained with Levanter. We
```
percyliang, line 41: Should we make bullets?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+hope that these libraries will be useful to the community, and we look forward to seeing what people do with them.
+
+## The Landscape of Foundation Model Training Frameworks
+
+Numerous foundation model training frameworks exist in the community, each with its strengths and focuses.
+For large language models (LLMs) (the focus of this release),
+the most well-known in the open source community is probably NVIDIA's PyTorch-based [Megatron-LM](https://github.com/NVIDIA/Megatron-LM),
```
percyliang, line 48: What about mingpt?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+released the Haiku-based [JAXformer](https://github.com/salesforce/jaxformer). Previously, Eleuther AI released
+[mesh-transformer-jax](https://github.com/kingoflolz/mesh-transformer-jax/), though it is mostly unmaintained now and uses
+older, quasi-deprecated JAX APIs for distributed training.
+
+## A New Codebase for Foundation Model Training
+
+Despite the wide array of existing frameworks, when we started, we found that none of them fully addressed our needs.
+At CRFM, we focused on three fundamental goals:
+
+* **Legibility and Composability**: We prioritized writing code that is easy to read, understand, and compose.
```
percyliang, line 70: Can we keep it 'legibility'?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+* **Reproducibility and Resilience**: We emphasized the ability to reproduce results *exactly*, even in the face of preemption and restarts from checkpoints.
```
percyliang, line 71: Can we keep it "reproducibility"?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+* **Scalability and Efficiency**: We aimed to fully utilize the Cloud TPU resources from Google's TPU Research Cloud program, as well as our own GPU resources, without compromising efficiency.
```
percyliang, line 72: "Scalability"

percyliang: I'd focus on the levanter codebase, not the TPU research program (which is specific to our setting)

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+
+We chose [JAX](https://github.com/google/jax/) as our framework because it is a powerful, flexible, and performant,
+and offers strong reproducibility guarantees. JAX also works well on TPUs, while we found that PyTorch support was still uneven.
+JAX is also a natural choice because it allows you to focus on the "what" of your code, and not on the "how": details of
+partitioning and communication can be left to the XLA compiler. Finally, JAX makes reproducibility easy, since it uses
+bitwise deterministic PRNGs by default, with careful control over the PRNG state.
```
percyliang, line 78: Seems like some of these points can be sprinkled into the three bullet points (scalability, reproducibility, legibility).

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+
+However, JAX is a low-level framework, and we found that, by itself, it did not provide the legibility that we wanted.
+
+For us, legibility is a top concern. We therefore created two new libraries: **Haliax** and **Levanter**. Haliax is a
```
percyliang, line 82: This blog post is already about levanter...so it should just be introducing haliax at this point?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+named tensor library that focuses on improving the legibility and compositionality of deep learning code while still
+maintaining efficiency and scalability. Levanter is a library for training foundation models built on top of Haliax. In
+addition to the goals of legibility, efficiency, and scalability, Levanter further strives for bitwise reproducibility,
+meaning that the same code with the same data will produce the exact same result, even in the presence of preemption and
+restarts.
```
percyliang, line 87: Some redundancy with what came before?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+
+# Haliax: Legibility via Named Tensors
```
percyliang, line 89: Can we name this "Desideratum 1 (legibility): named tensors via Haliax" or "Legibility: named tensors via Haliax"?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+
+To this, I'd add that the implicit broadcasting so common in deep learning code is a source of easy-to-miss bugs, and
+that named tensors eliminate many of these bugs.
+
+## A Quick Example: Attention in Haliax
+
+This blog post isn't the place for a full introduction to Haliax (please see the [Haliax tutorial](https://colab.research.google.com/drive/1TiTcQQ4V5mopbgCu1SVl-oqJtXn7rFnC)),
+but here's a quick example of a minimal, but full-featured, attention implementation in Haliax.
+
+```python
```
percyliang, line 116: Can we do a diff between what this would look like without named tensors?

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+
+    # weighted sum of values
+    return hax.dot(KPos, scores, value)
+```
+
+In this example, we've defined an axis for each dimension of our tensors. In Haliax, the named `Axis` is the basic
+building block of named tensors, pairing a name with a size. We can then use these axes to define our tensors, and use
+those axes to perform operations like `softmax` and tensor multiplication (`dot`).
+
+## Compositionality
```
percyliang, line 146: Not sure I'd call this composibility / modularity...  seems like more like automatic generalization or something...

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+
+This compositionality is possible because we've abstracted over the unreferenced dimensions of our tensors.
+In the first example, both the `Batch` and `Head` axes are unreferenced, so they are automatically "batched" over.
+Similarly, in the second example, we omit the `Head` axis from the `key` and `value` tensors, but attention still works.
+In the third example, we can use tuples of axes in many places where we would normally use a single axis.
+
+
+## Avoiding Bugs
+
+Earlier, I claimed that named tensors can help avoid common bugs. Here's an example of a bug that is easy to make
```
percyliang, line 185: Not sure we should use the first-person singular in general...

docs/Blog-Post.md
```diff
@@ -0,0 +1,525 @@
...
+# specifying "head" shards attention and "mlp" shards the feedforward
++tensor_parallel_mapping = {"head": "model", "mlp": "model"}
+# We store our parameters and optimizer states fully sharded along the embed axis
+-param_mapping = {"embed": "data"}
++param_mapping = {"embed": "data", **tensor_parallel_mapping}
+# During computation, we instead shard our data along the batch axis, and gather the parameters just-in-time
+-data_mapping = {"batch": "data"}
++data_mapping = {"batch": "data", **tensor_parallel_mapping}
+```
+
```
percyliang, line 304: This is cool!

docs/Blog-Post.md
dlwh: fair enough. i'll redo

docs/Blog-Post.md
dlwh: Thanks! I'm pretty sure I can get to pipeline parallelism this way too, but next release maybe.

docs/Blog-Post.md
dlwh: yeah it's more abstraction...

docs/Blog-Post.md
dlwh: i was explaining how the two components individually addressed the goals?

dlwh: (I'm reworking some)

docs/Blog-Post.md
dlwh: I dunno? I see it as "what does each component do"

docs/Blog-Post.md
dlwh: i dunno, as is, it's "we have goals" --> "jax helps address two of these goals so we chose Jax" --> "levanter and haliax get us the rest of the way"

docs/Blog-Post.md
dlwh: Sorry, I don't know what bullets you mean here.

docs/Blog-Post.md
dlwh: yeah sorry. it was originally announcement-post but i renamed it so you'd get the entire post instead of a small diff

docs/Blog-Post.md
dlwh: how about I literally quote the mingpt implementation, which is an implementation that claims to care about legibility, but looks deranged.

dlwh: (did this, line 238 ff)

docs/Blog-Post.md
dlwh: i'll rename right before mereg

docs/Blog-Post.md
```diff
@@ -0,0 +1,561 @@
...
+Today, we are excited to announce the release of [Levanter](https://github.com/stanford-crfm/levanter), a new codebase for training foundation models.
+Levanter is a [JAX](https://github.com/google/jax)-based codebase for training foundation models that is designed to be
+legible, scalable, and reproducible:
+
+1. Legible: Levanter has a named tensor module named Haliax that makes it easy to write legible, composable deep learning code, while still maintaining efficiency and scalability.
+2. Scalable: Levanter is designed to scale to large models, and to be able to train on a variety of hardware, including GPUs and TPUs.
+3. Reproducible: Levanter is bitwise deterministic, meaning that the same configuration will always produce the same results, even in the face of preemption and resumption.
+
+Today, we're releasing the first version of Levanter, along with several models trained with Levanter. We
```
percyliang, line 33: mix it up

docs/Blog-Post.md
```diff
@@ -0,0 +1,561 @@
...
+Key = hax.Axis("key", 64)  # key/query/value size
+Embed = hax.Axis("embed", 512)  # embedding size
+
+def attention(Key, KPos, query, key, value, mask):
+    # how similar is each query to each key
+    scores = hax.dot(Key, query, key) / jnp.sqrt(Key.size)
+
+    # mask out invalid positions
+    if mask is not None:
+      scores -= 1E9 * (1.0 - mask)
```
percyliang, line 130: 4 space indent

docs/Blog-Post.md
```diff
@@ -0,0 +1,561 @@
...
+This automatic generalization is possible because we've abstracted over the unreferenced dimensions of our tensors.
+In the first example, both the `Batch` and `Head` axes are unreferenced, so they are automatically "batched" over.
+Similarly, in the second example, we omit the `Head` axis from the `key` and `value` tensors, but attention still works.
+In the third example, we can use tuples of axes in many places where we would normally use a single axis.
+
+
+## Avoiding Bugs
+
+Earlier, we claimed that named tensors can help avoid common bugs. Here's an example of a bug that is easy to make
+and hard to spot in a traditional tensor library. Consider the following simple linear model:
```
percyliang, line 187: Ask: where's the bug?

docs/Blog-Post.md
```diff
@@ -0,0 +1,561 @@
...
+```python
+    y = y.transpose(1, 2).contiguous().view(B, T, C)
+```
+
+It would have been easy to miss this bug, and it would have been hard to spot in testing without testing
+against a reference implementation: the `view` suppresses any error we might get.
+
+With named tensors, this kind of reshaping juggling is obviated. We can just write the logic of our code
+and let the library (and XLA compiler) handle the reshaping for us.
+
```
percyliang, line 270: Reference our Levanter implementation again

docs/Blog-Post.md
```diff
@@ -0,0 +1,561 @@
...
+optimizer = init_optimizer(model)
+
++@hax.named_jit
+def train_step(model, opt_state, input_ids):
+  ... # elided for brevity
+
+  # ensure that intermediate states are sharded correctly
+-  loss, grads = grad_loss(model, input_ids)
++  with hax.axis_mapping(data_mapping):
++    loss, grads = grad_loss(model, input_ids)
```
percyliang, line 315: 4 spaces

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+# Introduction
+
+We are excited to announce the release of [Levanter](https://github.com/stanford-crfm/levanter), a new [JAX](https://github.com/google/jax)-based codebase for training foundation models.
+Levanter is designed to be legible, scalable, and reproducible:
+
+1. Legible: Levanter comes with a new named tensor library named Haliax that makes it easy to write legible, composable deep learning code, while still being high performance.
+2. Scalable: Levanter is designed to scale to large models, and to be able to train on a variety of hardware, including GPUs and TPUs.
+3. Reproducible: Levanter is bitwise deterministic, meaning that the same configuration will always produce the same results, even in the face of preemption and resumption.
+
+Today, we're releasing the first version of Levanter, along with [tutorials](https://colab.research.google.com/drive/1TiTcQQ4V5mopbgCu1SVl-oqJtXn7rFnC)
```
percyliang, line 29: Have explicit versioning?  Levanter 1.0?

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+    return jnp.mean((pred - target) * (pred - target))
+
+y_pred = x @ W
+mse(y_pred, y)
+```
+
+This code appears straightforward, but it's incorrect: the dimensions of `y_pred` and `y` are not the same.
+Because `y_pred` is a 2D array of shape `(128, 1)`, and `y` is a 1D array of shape `(128,)`, the `-` operator will broadcast `y` to shape `(128, 128)`.
+(This makes the subtraction an "outer product"-like operation rather than the intended elementwise subtraction.)
+But, you won't get an error at runtime; this is a silent bug. The `mean` call hides the bug by averaging over all values.
```
percyliang, line 134: But does this produce the same result?  If so, it's just a performance issue?

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+Batch = hax.Axis("batch", 128)
+Feature = hax.Axis("feature", 64)
+
+x = hax.random.uniform(PRNGKey(0), (Batch, Feature))
+y = hax.random.uniform(PRNGKey(1), Batch)
+
+def mse(pred, target):
+    return hax.mean((pred - target) * (pred - target), axis=Batch)
+
+W = hax.random.uniform(PRNGKey(2), (Feature,))
```
percyliang, line 154: Can we put W next to x, y just like in the pytorch code?  In general, make the code as isomorphic as possible.

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+## Haliax Tutorials
+
+This is just a taste of what Haliax can do. For more details, please see our interactive tutorials on Colab:
+
+* [Introduction to Haliax with Transformers](https://colab.research.google.com/drive/1TiTcQQ4V5mopbgCu1SVl-oqJtXn7rFnC?usp=sharing)
+* [Scaling Transformers in Haliax](https://colab.research.google.com/drive/1QX4yH3zRFF3Xiibf1aahETcSQ5nbcUMz?usp=sharing), including FSDP in JAX.
+
+# Levanter: Bitwise Reproducible Foundation Models with JAX
+
+Levanter is a library for training foundation models built on top of Haliax. It provides a complete pipeline
```
percyliang, line 378: Add transition: Beyond legibility, the second property is reproducibility...  Let's use the 3 terms (in their noun form) more prominently.

docs/Blog-Post.md
dlwh: no, it doesn't produce the same result.

The code as written computes `diff[i, j] = y_pred[i] - y[j]`and then `mean(diff, axis=(i, j))`, while the intent is `diff[i] = y_pred[i] - y[i]`. (That's what i mean by "outer product like")

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+
+## Bitwise Reproducibility
+
+One of the benefits of JAX is that it offers strong guarantees for reproducibility. In particular, JAX's fine-grained
+control over PRNG states makes it easy to ensure bitwise reproducibility, especially when using TPUs.
+Levanter takes advantage of this to offer bitwise reproducibility for training runs, even after preemption. In particular,
+the same run with the same code on the same set of hardware (e.g. a v3-32 or a v3-256) will produce the exact same loss curve, even if it is
+preempted and resumed multiple times. As an example, here is a screenshot of a training run being resumed multiple times, even on different TPU pod slices:
+
+![plot showing bitwise reproducibility with four training runs superimposed with the exact same loss curve](figures/bitwise_repro_curve.png)
```
percyliang, line 393: Can we show two Pytorch loss curves which are different but are the same code with the same random seed?  I don't think people might appreciate this difference.

Also, could we argue more passionately why reproducibility is nice.  For example, you get loss spikes, but they go away when you run it...it's like a Heissenbug or whack-a-hole - whatever analogy you want to use.

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+
+Levanter also logs everything necessary to exactly reproduce a run: the git SHA, code, configuration,
+and a pip-freeze of the environment. Checkpoints serialize the entire model state, including the optimizer state,
+as well as the "main" PRNG state, which is used to generate the other PRNG states. This means that you can
+exactly reproduce a run by simply checking out the git SHA, installing the dependencies, and running the code (on the same
+hardware configuration).
+
+## Efficiency and Scale
+
+XXX something something v3-256 scaling numbers?
```
percyliang, line 406: Make sure we transition: We talked about legibility, reproducibility.  Final property is scalability.

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
+# ---
+layout: blog
+title: Levanter —  Legible, Scalable, Reproducible Foundation Models with JAX
```
Ivan-Zhou, line 3: NIT - it seems a verb is missing in the title, like "Train Legible, Scalable, Reproducible Foundation Models with JAX"

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+mse(y_pred, y)
+```
+
+This code appears straightforward, but it's incorrect: the dimensions of `y_pred` and `y` are not the same.
+Because `y_pred` is a 2D array of shape `(128, 1)`, and `y` is a 1D array of shape `(128,)`, the `-` operator will broadcast `y` to shape `(128, 128)`.
+(This makes the subtraction an "outer product"-like operation rather than the intended elementwise subtraction.)
+But, you won't get an error at runtime; this is a silent bug. The `mean` call hides the bug by averaging over all values.
+
+This is a common bug in deep learning code, and it's easy to miss. We, and anecdotally many others we have spoken to, have
+individually lost multiple days to this exact bug, and it shows up in every deep learning framework.
```
Ivan-Zhou, line 137: This is not just a problem to deep learning, but also regular matrix computation. I am sure lots of numpy users share the same complaint :)

docs/Blog-Post.md
dlwh: hrm, I think as is is better. Name - Thing you get with Name is not an uncommon pattern for titles i think

docs/Blog-Post.md
dlwh: I'm always scared of 1.0, but ok

docs/Blog-Post.md
dlwh: oops, yeah

docs/Blog-Post.md
dlwh: i'll try to get a thing

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+
+Levanter works out of the box with either [Hugging Face Datasets](https://huggingface.co/datasets) (including streaming) or urls of (compressed)
+jsonl files. Caches can be stored in any fsspec-compatible file system, including GCS and local file systems. We use
+[Hugging Face Tokenizers](https://huggingface.co/docs/tokenizers/) for tokenization.
+
+### Live Visualization during Training
+
+Levanter also provides a feature for visualizing the probability of each token in a sample of the validation set during training.
+When training large models, it can be difficult to get a sense of how the model is learning. This is especially true
+when training on novel data sets. As an example, we have seen issues with early versions of new datasets where the
```
percyliang, line 452: nit: datasets

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+learn differently from Transformers.
+
+### Other Features
+
+* **Training**: Levanter uses [Optax](https://github.com/deepmind/optax) for optimization
+  (though our new optimizer, [Sofia](https://arxiv.org/abs/2305.14342), is coming to Levanter soon!)
+* **Logging**: Logging is done with [WandB](https://wandb.ai/), complete with a fancy online visualization of the validation set during training.
+* **Checkpointing**: Distributed checkpointing is supported via Google's [TensorStore](https://google.github.io/tensorstore/) library. Training can even be resumed on a different number of hosts, though this breaks reproducibility for now.
+* **Export**: We also support exporting models to the Hugging Face Hub, with export compatible with Pytorch and Transformers via [SafeTensors](https://github.com/huggingface/safetensors).
+* **Stability**: The GPT-2 implementation uses the [Mistral stability trick](https://crfm.stanford.edu/2021/08/26/mistral.html) to improve stability during training.
```
percyliang, line 477: Will this be no longer necessary with Sophia?

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+the [Hugging Face Hub](https://huggingface.co/stanford-crfm) and can be used with the Hugging Face Transformers library,
+in Pytorch, Tensorflow, and JAX. We have more in development and will release them as they become available.
+
+- We are release a suite of music models trained on the [Lakh MIDI](https://colinraffel.com/projects/lmd/) corpus. The largest, 750M parameter one is available [here](https://huggingface.co/stanford-crfm/music-large-100k).
+ Please see [John Thickstun](https://johnthickstun.com/)'s [blogpost](XXX) for more, and a cool demo page!
+- We also have a new 1.4 billion parameter of the [Backpack Model](http://backpackmodels.science/) architecture developed by [John Hewitt](https://nlp.stanford.edu/~johnhew/) and coauthors.
+  This model is available [here](https://huggingface.co/stanford-crfm/levanter-backpack-1b).
+- [Levanter GPT](https://huggingface.co/stanford-crfm/levanter-gpt) is a 1.5B parameter GPT-2 model trained on the
+  [OpenWebText](https://skylion007.github.io/OpenWebTextCorpus/) corpus.
+- Hot off the presses, we have a 1.5B model trained on the [Pile](https://pile.eleuther.ai/) corpus.
```
percyliang, line 553: "The Pile" is the name.

percyliang, line 553: Why is this hot of the press? sounds like it's an old model...

docs/Blog-Post.md
```diff
@@ -0,0 +1,578 @@
...
+
+## Released Models
+
+Along with the release of the code, we are releasing a few models trained using Levanter. These models are available on
+the [Hugging Face Hub](https://huggingface.co/stanford-crfm) and can be used with the Hugging Face Transformers library,
+in Pytorch, Tensorflow, and JAX. We have more in development and will release them as they become available.
+
+- We are release a suite of music models trained on the [Lakh MIDI](https://colinraffel.com/projects/lmd/) corpus. The largest, 750M parameter one is available [here](https://huggingface.co/stanford-crfm/music-large-100k).
+ Please see [John Thickstun](https://johnthickstun.com/)'s [blogpost](XXX) for more, and a cool demo page!
+- We also have a new 1.4 billion parameter of the [Backpack Model](http://backpackmodels.science/) architecture developed by [John Hewitt](https://nlp.stanford.edu/~johnhew/) and coauthors.
```
percyliang, line 549: 1.4B instead of 1.4 billion?  Be consistent since you say 1.5B later

percyliang (approved):
Just a few more comments, but this looks great!

docs/Blog-Post.md
dlwh: I don't see why, but I'm of course not optimization person. My understanding is that the issue is that the grad norm gets big and then numerics get wonky. Is your point that the hessian estimate will say "whoa this is steep, slow down"?

I think update clipping is helpful here, as is more layer norms. (always want more layer norms). I guess sophia and update clipping are doing similar things in this regard.

 it's also usually not a huge problem with bf16 (though we saw a few problems with backpacks even with the tweak...)

docs/Blog-Post.md
dlwh: (i clarified)

PR #163, opened 2023-05-25
Partitioned Random

dlwh:
Switch to the new threefry_partitioned rng for jax. breaks backwards bitwise repro, but worth it. I need to run some tests

Means we don't need to use generate_sharded anymore, which is great

### DIFF
--- a/src/haliax/nn/embedding.py
+++ b/src/haliax/nn/embedding.py
@@ -13,20 +13,11 @@ class Embedding(eqx.Module):
     Vocab: Axis = eqx.static_field()
     Embed: Axis = eqx.static_field()
 
-    def __init__(
-        self,
-        Vocab: Axis,
-        Embed: Axis,
-        initializer_range: float = 0.02,
-        *,
-        key,
-    ):
-        super().__init__()
-
-        self.Vocab = Vocab
-        self.Embed = Embed
-
-        self.weight = hax.random.generate_sharded(hax.random.normal)(key, (Vocab, Embed)) * initializer_range
+    # TODO: should allow axisspec for Embed
+    @staticmethod
+    def init(Vocab: Axis, Embed: Axis, initializer_range: float = 0.02, *, key):
+        weight = hax.random.normal(key, (Vocab, Embed)) * initializer_range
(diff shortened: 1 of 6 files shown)


src/levanter/config.py
```diff
@@ -271,19 +277,9 @@ class TrainerConfig:
 
     # config related to partitioning
 
-    batch_axis: Optional[
```
dlwh, line 274: just formatting

src/levanter/config.py
```diff
@@ -450,7 +446,7 @@ def _optimizer(learning_rate):
 
         return optimizer
 
-    def maybe_load_checkpoint(self, model: M, training_state: S) -> typing.Tuple[M, S, Optional[int]]:
```
dlwh, line 453: cosmetic

src/levanter/config.py
```diff
@@ -415,9 +412,8 @@ def parameter_axis_mapping(self) -> ResourceMapping:
         return mapping
 
     def _initialize_jax_config(self):
-        """Initialize global jax config with settings we like, based on config"""
-        jax_utils.set_hardware_rng_ops(self.use_hardware_rng)
-        jax.config.update("jax_array", self.use_jax_array)
```
dlwh, line 420: real change

src/levanter/models/gpt2.py
```diff
@@ -23,10 +23,6 @@
 )
 
 
-# we use sharded_normal here so that the random initialization can be split across devices
-sharded_normal = hax.random.generate_sharded(hax.random.normal)
```
dlwh, line 27: jax's new RNG does a version of this

src/haliax/nn/embedding.py
```diff
@@ -13,20 +13,11 @@ class Embedding(eqx.Module):
     Vocab: Axis = eqx.static_field()
     Embed: Axis = eqx.static_field()
 
-    def __init__(
```
dlwh, line 16: missed switching this one to a staticmethod ctor in previous PR

dlwh:
my quick tests seem ok

PR #172, opened 2023-06-01
[EXP][DO NOT MERGE] Add changes for new backpack exp

Ivan-Zhou:
- 100k step gpt2-Large with 48 senses
- We can make sure we have a layernorm at the end of the contextualization network.
- LN the sense vectors everywhere maybe
- Remove scale by num of senses

### DIFF
--- a/src/levanter/models/backpack.py
+++ b/src/levanter/models/backpack.py
@@ -363,10 +363,6 @@ class BackpackLMHeadModel(StateDictSerializationMixin, eqx.Module):
         hidden_states = hax.dot(self.config.KeyPos, contextualization_weights, sense_vectors)  # (seq, senses, embed)
         hidden_states = hax.sum(hidden_states, axis=self.config.Senses)
 
-        # Rescale - this is important for large num_senses
-        scale = self.config.Senses.size
-        hidden_states = hidden_states / scale
-
         lm_logits = self.embeddings.unembed(hidden_states)
 
         return lm_logits
--- a/src/levanter/models/backpack.py
+++ b/src/levanter/models/backpack.py
@@ -115,7 +115,7 @@ class WeightsOnlyAttention(StateDictSerializationMixin, eqx.Module):
         k_c, _ = jrandom.split(key, 2)
         c_attn = hnn.Linear.init(In=Embed, Out=(Qk, config.Senses, config.SenseHeadDim), key=k_c, use_bias=use_bias)
         dropout = hnn.Dropout(config.attn_pdrop)
-        ln = hnn.LayerNorm.init(config.Embed, eps=config.layer_norm_epsilon)
+        ln = hnn.LayerNorm.init((config.Pos, config.KeyPos), eps=config.layer_norm_epsilon)
 
         return WeightsOnlyAttention(config, c_attn, dropout, ln)
 
--- a/src/levanter/models/backpack.py
+++ b/src/levanter/models/backpack.py
@@ -332,7 +332,7 @@ class BackpackLMHeadModel(StateDictSerializationMixin, eqx.Module):
             config=config,
             key=k_attn,
         )
-        rescaler = jnp.array( 1.0 / config.Senses.size)
+        rescaler = jnp.array(config.Senses.size)
 
         return BackpackLMHeadModel(
             transformer=transformer,
@@ -364,10 +364,10 @@ class BackpackLMHeadModel(StateDictSerializationMixin, eqx.Module):
 
         ## Weight-and-sum
         hidden_states = hax.dot(self.config.KeyPos, contextualization_weights, sense_vectors)  # (seq, senses, embed)
-        hidden_states = hax.sum(hidden_states, axis=self.config.Senses)
+        hidden_states = hax.sum(hidden_states, axis=self.config.Senses) # (seq, embed)
 
         # Rescale 
-        hidden_states = hidden_states * self.rescaler
+        hidden_states = hidden_states / self.rescaler
 
         lm_logits = self.embeddings.unembed(hidden_states)
 
--- a/src/levanter/models/backpack.py
+++ b/src/levanter/models/backpack.py
@@ -290,6 +290,7 @@ class BackpackLMHeadModel(StateDictSerializationMixin, eqx.Module):
     embeddings: BackpackGpt2Embeddings
     sense_net: BackpackSenses
     kq_selfattention: WeightsOnlyAttention
+    rescaler
 
     @property
     def config(self):
@@ -331,12 +332,14 @@ class BackpackLMHeadModel(StateDictSerializationMixin, eqx.Module):
             config=config,
             key=k_attn,
         )
+        rescaler = jnp.array( 1.0 / config.Senses.size)
 
         return BackpackLMHeadModel(
             transformer=transformer,
             embeddings=embeddings,
             sense_net=sense_net,
             kq_selfattention=kq_selfattention,
+            rescaler=rescaler,
         )
 
     def __call__(self, input_ids: NamedArray, attn_mask: Optional[NamedArray], *, inference, key):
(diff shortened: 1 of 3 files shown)


src/levanter/models/backpack.py
```diff
@@ -114,8 +115,9 @@ def init(config: Gpt2Config, *, key) -> "WeightsOnlyAttention":
         k_c, _ = jrandom.split(key, 2)
         c_attn = hnn.Linear.init(In=Embed, Out=(Qk, config.Senses, config.SenseHeadDim), key=k_c, use_bias=use_bias)
         dropout = hnn.Dropout(config.attn_pdrop)
+        ln = hnn.LayerNorm.init(config.Senses, eps=config.layer_norm_epsilon)
```
Ivan-Zhou, line 118: Do we need this LayerNorm to be on the entire output axises, `(Qk, config.Senses, config.SenseHeadDim)`?

src/levanter/models/backpack.py
```diff
@@ -241,21 +245,23 @@ def init(
             key=k_mlp,
             use_bias=config.use_bias,
         )
+        ln_2 = hnn.LayerNorm.init(config.Senses, eps=config.layer_norm_epsilon)
```
Ivan-Zhou, line 248: Same here, do we need it to be `(config.Senses, config.Embed)`?

dlwh, line 248: i think this should be Embed?

dlwh: (you want to normalize over the embed dimension, sort of like softmax(x, axis=Pos) rather than softmax(x, axis=Head) or something)

dlwh: i think just config.Embed. I think we just want to ensure that each sense's vector is normalized, which means the embedding dim should be normalized.

src/levanter/models/backpack.py
dlwh: also Embed for the same reason i think

Ivan-Zhou: If I use Embed, I would get an error:
```
FAILED tests/test_backpack.py::test_backpack_predict - ValueError: axis (Axis(name='embed', size=768),) is not in (Axis(name='position', size=512), Axis(name='senses', size=16), Axis(name='key_position', size=512))
```

So based on the principle that we want to apply LayerNorm per senses, it should be normalized along the axises of Position and KeyPosition:
```
ln = hnn.LayerNorm.init((config.Pos, config.KeyPos), eps=config.layer_norm_epsilon)
```

dlwh: i don't think this is right either (in particular, this lets information leak from the future into the past). I now think putting a layer norm in after the final attention isn't going to work. I think we should aggressively layer norm the sense vectors as they come out of the sense network. The issue is how to make them have large enough magnitude, but i think in high dimension it maybe not doesn't matter.

Let's defer to JohnH.

john-hewitt: Pasting from slack for reference

Layer Norm after the Q and K in the final attention seems fine to me. We can also Layer Norm the sense vector outputs; shouldn't the bias and gain help with the magnitude?

Ivan-Zhou: If I understand correctly, I think this is what we need for LN on Q and K in the final attention: https://github.com/stanford-crfm/levanter/pull/172/commits/03e0fcfad33b60b2a154041c277c6114be2d5fa3

john-hewitt: Seems right.

Ivan-Zhou:
One of the change is causing the following issue at model sharding:

```
The above exception was the direct cause of the following exception:

  File "/home/ivanzhou/levanter/examples/backpack_example.py", line 293, in <module>
    main()
  File "/home/ivanzhou/levanter/src/levanter/config.py", line 585, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "/home/ivanzhou/levanter/examples/backpack_example.py", line 111, in main
    model = init_model()
  File "/home/ivanzhou/levanter/src/haliax/partitioning.py", line 283, in f
    return cached_pjitted_fun(dynamic_donated, dynamic_reserved, static)
TypeError: 'NoneType' object is not iterable
wandb: Waiting for W&B process to finish... (failed 1). Press Control-C to abort syncing.
  File "/home/ivanzhou/levanter/examples/backpack_example.py", line 293, in <module>
    main()
  File "/home/ivanzhou/levanter/src/levanter/config.py", line 585, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "/home/ivanzhou/levanter/examples/backpack_example.py", line 111, in main
    model = init_model()
  File "/home/ivanzhou/levanter/src/haliax/partitioning.py", line 283, in f
    return cached_pjitted_fun(dynamic_donated, dynamic_reserved, static)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/traceback_util.py", line 166, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/pjit.py", line 208, in cache_miss
    outs, out_flat, out_tree, args_flat = _python_pjit_helper(
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/pjit.py", line 155, in _python_pjit_helper
    out_flat = pjit_p.bind(*args_flat, **params)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/core.py", line 2633, in bind
    return self.bind_with_trace(top_trace, args, params)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/core.py", line 383, in bind_with_trace
    out = trace.process_primitive(self, map(trace.full_raise, args), params)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/core.py", line 790, in process_primitive
    return primitive.impl(*tracers, **params)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/pjit.py", line 1088, in _pjit_call_impl
    always_lower=False, lowering_platform=None).compile()
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/interpreters/pxla.py", line 2313, in compile
    executable = UnloadedMeshExecutable.from_hlo(
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/interpreters/pxla.py", line 2651, in from_hlo
    in_shardings_xla, out_shardings_xla = _get_mesh_pspec_shardings_from_executable(
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/interpreters/pxla.py", line 2408, in _get_mesh_pspec_shardings_from_executable
    in_pspec, out_pspec = pjit._get_pspec_from_executable(xla_executable, mesh)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/pjit.py", line 1963, in _get_pspec_from_executable
    in_ppspec, out_ppspec = _get_ppspec_from_executable(executable, mesh)
  File "/home/ivanzhou/venv310/lib/python3.10/site-packages/jax/_src/pjit.py", line 1954, in _get_ppspec_from_executable
    for s in input_op_shardings:
jax._src.traceback_util.UnfilteredStackTrace: TypeError: 'NoneType' object is not iterable
```

dlwh:
That’s gotta be the scalar. Can you open an issue and I’ll fix this
weekend? I’m the mean time could you try without?

Ivan-Zhou:
Yep, after [this commit](https://github.com/stanford-crfm/levanter/pull/172/commits/872c4eb58ca242335fcc8325822dc820e874254c) to remove the scalar variable, I was able to avoid this error and launched my training successfully. 

Here's the issue: https://github.com/stanford-crfm/levanter/issues/175

Ivan-Zhou:
This PR has been replaced with this: https://github.com/stanford-crfm/levanter/pull/190/

PR #180, opened 2023-06-05
Update Getting-Started-CUDA.md

dlwh:
Taking a guess at what needed to change. @Ivan-Zhou and @julian-q please lmk.

### DIFF
--- a/docs/Getting-Started-CUDA.md
+++ b/docs/Getting-Started-CUDA.md
@@ -1,31 +1,44 @@
 # Getting Started with CUDA
 
-**Note** that right now CUDA support is very preliminary and Levanter does not work on a multi-machine setup.
+**Note** that right now CUDA support is very preliminary and Levanter does not work on a multi-machine setup. (We think this is just a dependency/environment issue.)
 
-## Setting up dependencies
+We only test on Ampere (e.g. A100s or 30xx series)  GPUs. If it works with Jax it should work, though.
+
+## Installation
+### Setting up Conda Environment
 ```bash
 # 3.11 is too new for tensorstore
 # this doesn't actually install pytorch, but it bundles cuda dependencies nicely
 conda create --name levanter -c pytorch -c nvidia pytorch-cuda=11.7 python~=3.10
 conda activate levanter
-pip install "jax[cuda11_cudnn82]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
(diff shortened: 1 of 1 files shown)


docs/Getting-Started-CUDA.md
```diff
@@ -1,31 +1,44 @@
...
 
 ## Running a job on Slurm
 
 ### Single Node
 
+Simple example, customize to your needs:
+
 ```bash
-srun --cpus-per-task=128 --gres=gpu:8 --job-name=levanter-multi-1 --mem=1000G --nodelist=sphinx[7-8] --open-mode=append --partition=sphinx --time=14-0 ~/src/levanter/scripts/run-slurm.sh python examples/gpt2_example.py --config_path config/gpt2_small.yaml
+srun --cpus-per-task=128 --gres=gpu:8 --job-name=levanter-multi-1 --mem=1000G --nodelist=sphinx7 --open-mode=append --partition=sphinx --time=14-0 ~/src/levanter/scripts/run-slurm.sh python examples/gpt2_example.py --config_path config/gpt2_small.yaml
```
Ivan-Zhou, line 41: Note that I just tested in sphin6, which used CUDA 11.7. I installed jax 0.4.11. 

I had this issue when running Levanter. This is consistent with what I encountered on jagupard (P6000):

```
FAILED tests/test_backpack.py::test_backpack_predict - jaxlib.xla_extension.XlaRuntimeError: INTERNAL: nvlink exited with non-zero error code 256, output: nvlink fatal   : Input file '/tmp/ivanzhou/tempfile-sphinx6-efa86375-3475856-5fd80f1d5370f.cubin' newer than toolkit (118 vs 113)
```

Ivan-Zhou (approved):
The change to the doc LGTM. I just have a separate issue that is specific to the slurm environment.

docs/Getting-Started-CUDA.md
Ivan-Zhou: I don't think this is an issue of Levanter. This is more likely something between CUDA toolkit and Jax.

dlwh: image

Yeah looks like we need to increase the support version range and/or require people to build form source on sphinx and/or get the sphinxes upgraded. We should probably do all three.

dlwh: (that's a screenshot of the current jax readme)

PR #181, opened 2023-06-05
First steps to pypi package

dlwh:
Followed https://packaging.python.org/en/latest/tutorials/packaging-projects/ , ended up removing requirements.txt because that's deprecated


pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: Apache Software License 2.0",
+    "Operating System :: Linux",
+    "Operating System :: MacOS :: MacOS X",
+    "Development Status :: 3 - Alpha",
+    "Intended Audience :: Science/Research",
+]
+
+dependencies = [
+    # we'll require that you install cuda yourself, since the extras vary by system.
```
Ivan-Zhou, line 25: You meant jax instead of cuda?

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+]
+
+[project.urls]
+"Homepage" = "https://github.com/stanford-crfm/levanter"
+"Bug Tracker" = "https://github.com/stanford-crfm/levanter/issues"
+
+
 [tool.black]
 line-length = 119
 target-version = ["py36", "py37", "py38"]
```
Ivan-Zhou, line 58: According to the conversation [here](https://github.com/psf/black/issues/751), 3.6 through 3.8 are currently the same as far as Black is concerned, so either variant will do the same.
According to [this PR](https://github.com/mopidy/mopidy/pull/2042/files), Python 10 should be `py310`.

So maybe we can do: `["py38", "py39", "py310"]`?

pyproject.toml
```diff
@@ -1,3 +1,58 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "levanter"
+version = "0.0.1"
+authors = [
+  { name="David Hall", email="dlwh@cs.stanford.edu" },
+    {name="Ivan Zhou", email="ivanz@stanford.edu"}
```
Ivan-Zhou, line 10: Feel honored 😄 !

Ivan-Zhou (approved):

pyproject.toml
dlwh: i think we only support python 310+ so i'll just require that

pyproject.toml
dlwh: super glad to be working with you! 😃

PR #189, opened 2023-06-07
Backpack Roundtrip Test

dlwh:
adds a test for roundtripping backpacks. doesn't quite save a hf-loadable checkpoint yet, but everything else is in place.

I'm trying to get us to a state where we can have a clean OO packaging of everything needed to "port" a model, so that the CLI scripts can be nice

PR #191, opened 2023-06-08
introduce HFCheckpointConverter to unify hf serialization

dlwh:
(open to bikeshedding on the name)

Supersedes #189 

could still be cleaned up, but i think overall this is an improvement. introduces a single class that handles the serialization for GPT2, MPT, and Backpacks in a pluggable way.

Haven't actually tested HF upload yet.

Next up is a registry, and maybe unifying the various _example and _eval scripts

### DIFF
--- a/src/levanter/compat/hf_checkpoints.py
+++ b/src/levanter/compat/hf_checkpoints.py
@@ -134,7 +134,7 @@ _GLOBAL_SAVE_COUNT = 0
 
 
 @dataclass
-class HFCheckpointConverter(abc.ABC, Generic[LevConfig]):
+class HFCheckpointConverter(Generic[LevConfig]):
     LevConfigClass: Type[LevConfig]
     reference_checkpoint: Optional[Union[str, RepoRef]] = None
     "A reference HF Hub checkpoint to extract non-parameter files (like model code an config from)"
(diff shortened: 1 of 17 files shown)


dlwh:
@Ivan-Zhou after merging this into your other branch, #190 works

src/levanter/compat/hf_checkpoints.py
```diff
@@ -55,269 +135,498 @@ def to_dict(self) -> dict:
...
+        dict_config = config.to_dict()
+
+        # copy over the default keys
+        for k in KEYS_TO_COPY_FROM_BASE_CONFIG:
+            attr = getattr(self.default_hf_config, k, None)
+            if attr is not None:
+                dict_config[k] = attr
+
+        if self.config_overrides:
+            dict_config.update(self.config_overrides)
```
Ivan-Zhou, line 444: Instead of overriding config here, can we pass `config_overrides` in `to_hf_config()`?

Ivan-Zhou (commented):
Add a patch stacked to this PR: https://github.com/stanford-crfm/levanter/pull/193

src/levanter/compat/hf_checkpoints.py
dlwh: sure, i was actually intending on removing config_overrides from to_hf_config but happy to go the other way

PR #193, opened 2023-06-09
Fix converter for backpack

Ivan-Zhou:
A few minor changes:
1. Remove `HFAutoMapConfig` class as it is no longer used. Instead, I can put them in `config_overides` (for example [here](https://github.com/stanford-crfm/levanter/blob/671422ed82ffb269d3fe4d297a632d54d572516b/config/local_backpack_hf.yaml#L13-L17))
2. Pass `self.config_overrides` to `to_hf_config`
3. Add `backpack.` in front of `word_embeddings.weight` and `position_embeddings.weight`, see error log below:

```
Some weights of the model checkpoint at /sailhome/ivanzhou/dev/levanter-backpack-gpt2-large-ln/ were not used when initializing BackpackGPT2LMHeadModel: ['position_embeddings.weight', 'word_embeddings.weight']
- This IS expected if you are initializing BackpackGPT2LMHeadModel from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing BackpackGPT2LMHeadModel from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of BackpackGPT2LMHeadModel were not initialized from the model checkpoint at /sailhome/ivanzhou/dev/levanter-backpack-gpt2-large-ln/ and are newly initialized: ['backpack.word_embeddings.weight', 'backpack.position_embeddings.weight']
```

### DIFF
--- a/src/levanter/compat/hf_checkpoints.py
+++ b/src/levanter/compat/hf_checkpoints.py
@@ -116,25 +116,6 @@ def _coerce_to_rr(s: Union[str, RepoRef]) -> RepoRef:
         return RepoRef.from_string(s)
 
 
-@dataclass
-class HFAutoMapConfig:
-    """
-    To create a custom AutoModel class, in your model's config.json,
-    you will need to add a field like this:
-    "auto_map": {
-        "AutoConfig": "backpack_config.BackpackGPT2Config",
-        "AutoModelForCausalLM": "backpack_model.BackpackGPT2LMHeadModel"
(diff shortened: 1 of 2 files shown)


dlwh:
thanks, i appreciate it!

PR #227, opened 2023-07-02
Fix string interpolation for joining paths

### LINKED ISSUE #139: We shouldn't use string interpolation for joining paths
it doesn't work well with gcsfs paths. We should use os.path.join instead

mkly:
Fix string interpolation for joining paths

* Replace string interpolation with `os.path.join` for file and directory paths

Resolves #139

### DIFF
--- a/src/levanter/compat/hf_checkpoints.py
+++ b/src/levanter/compat/hf_checkpoints.py
@@ -333,13 +333,13 @@ class HFCheckpointConverter(Generic[LevConfig]):
 
         id, rev = self._get_ref(ref)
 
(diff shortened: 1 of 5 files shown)


mkly:
This was basically just a grep to see what would be easy to find. Possible I've missed some. Let me know if you'd like some others addressed and if there is some test coverage you'd like to see in regards to these changes. :tropical_drink:

dlwh:
thanks for jumping in and contributing! Let me know if you'd like to talk through what you're thinking of using levanter for!

mkly:
To be honest I'm just poking around looking for stuff to do after some time away. If you have anything else you think you could use some help with, definitely let me know.

PR #228, opened 2023-07-02
Fix flake8 repo in pre-commit

mkly:
Fix flake8 repo in pre-commit

flake8 fails to run in pre-commit due to gitlab repo no longer exitsing at https://gitlab.com/pycqa/flake8 (returns 404). This is following the Contibuting documentation which indicates the following command:

```
pre-commit run --all-files
```

* Update flake8 to https://github.com/PyCQA/flake8 which is the recommended location based on the current pre-commit docs: https://pre-commit.com/hooks.html

_Note_: While this does fix the installation of flake8, running `pre-commit run --all-files` causes approximately 32 files to be changed on the current *main* branch. This may indicate that if this is now applied, that linting should also be added to CI to protect newer commits with divergent formatting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
     hooks:
     -   id: isort
 
--   repo: https://gitlab.com/pycqa/flake8
+-   repo: https://github.com/PyCQA/flake8
     rev: 3.9.2
     hooks:
     -   id: flake8


mkly:
Following the contribution docs I realized `pre-commit` could not successfully run. I updated flake8 to the valid github repo to fix the issue. With that said, I'm not sure this is followed  much anymore as it does make quite a few changes when `pre-commit run --all` is run. Not sure if it is better to remove it altogether. Another alternative might be one big ugly lint commit and then adding some linting to CI.

dlwh (approved):

dlwh:
Hi Mike,

Thanks for noticing/fixing! It's probably a good idea to add to CI/do the big ugly commit. I'll do the big ugly commit at a minimum now.

dlwh:
looks like it's mostly from my (very recent!) extraction of haliax from the levanter repo, which means that flake now treats haliax as an "external"/3p lib for import purposes rather than as part of the project

Issue #218, opened 2023-06-28
Support Weighted Data Mixtures


dlwh:
* Dive into the data loader code
* At each batch, you roll alpha die B times, and get domains, and then take the next example from the domain for each
    * Sharding: fold_in shard index into random key
* Test: unit test with alpha = 1; 10K examples with <1 alpha, test that sharding does what you want

Issue #119, opened 2023-04-20
Preprocessing: Reproducible batch order even when number of workers changes


dlwh:
Part of #99 

A bit tricky to pull off. The design doc has this:

## Sharded Reading

We say there are K input shards, W writers, R readers. We assume K >= W (though typically K is not too large), and W ≈ R.
We produce N chunks. We also define an idealized number of readers R*, which defines the global ordering over the data.
Typically R* should be the maximum number of readers we expect to actually use.

We want to be able to read from the cache in a way that is deterministic and reproducible, even if the number of readers
changes. We also want readers to only read from the chunks that they need to read from.
We pretend the list of data is infinite by cycling. We cannot track epochs.

NB Our goal is a deterministic ordering over examples, and not merely chunks or even documents.

Given a list of chunks and the idealized number of readers R*, we define the global ordering over chunks as follows:
First define R* iterators over chunks, with `chunk_iterators[r]` being defined as `loop(all_chunks)[r::R*]`.

Next, define a function `mk_examples(chunk_iterator)` that takes a list of iterators over chunks and returns
a list of examples. Define `chunk_examples[r] = mk_examples(chunk_examples[r])`.
This function depends on our sequence length, etc. Then the ordering over examples is:

`chunk_examples[0][0], chunk_examples[1][0], ..., chunk_examples[R*-1][0], ..., chunk_examples[0][1], chunk_examples[1][1], ..., chunk_examples[R*-1][1], ...`
that is, `example[i] == chunk_examples[i % R*][i // R*]`

If we have $R*$ readers, then each `reader_iterator[r][j] == chunk_examples[r][j] == example[j * R* + r]`.
Moreover, if either R or R* is a multiple of the other, then we still get a nice property where
each reader reads from a strided slice of the chunk_iterators:

(Boring math)
If we have R readers, then `reader_iterator[r][j] == example[j * R + r] == chunk_examples[(j * R + r) % R*][(j * R + r) // R*]`
If we have `R == n * R*`, then `reader_iterator[r][j] == example[j * R + r] == chunk_examples[(j * R + r) % R*][(j * R + r) // R*]
 == chunk_examples[r % R*][(j * n * R* + r) // R*] == chunk_examples[r % R*][j * n + r // R*],` so each reader reads from
a strided slice (specifically `islice(..., r//R*, None, n)`)
[...]

PR #229, opened 2023-07-03
Weighted Mixture of Datasets for Training

Ivan-Zhou:
This PR implements https://github.com/stanford-crfm/levanter/issues/218. 
- [x] Design doc
- [x] Implement LMMixtureDatasetConfig 
- [x] Implement MixtureDataset 
- [x] Testing


docs/design/Multiple-Data-Mixture.md
```diff
@@ -0,0 +1,92 @@
...
+a for loop over `one_item_generator`, then at every step, we slice from `one_item_generator` to get 
+B sequence, what do we want to achieve with the first for loop?
+
+### Weighted Sampling
+At every step of `ShardedBatchLoader.__iter__()`, we need to sample B sequences from N datasets. 
+The weights of the datasets are used to sample from N datasets. Specifically, we will sample
+B sequences from the i-th dataset with probability `weight[i] / sum(weight)`. 
+
+### Deterministic Batches
+I am not sure if we can still achieve deterministic batches with weighted sampling.
```
dlwh, line 86: as long as the random numbers are deterministic it's fine.

supporting this with #119 is hard though

docs/design/Multiple-Data-Mixture.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Validation Set
+We will apply the same logic to `ReplicatedBatchLoader` for validation sets.
```
dlwh, line 89: i think it's better to keep validation sets split and report separately

docs/design/Multiple-Data-Mixture.md
```diff
@@ -0,0 +1,92 @@
...
+B sequences from the i-th dataset with probability `weight[i] / sum(weight)`. 
+
+### Deterministic Batches
+I am not sure if we can still achieve deterministic batches with weighted sampling.
+
+### Validation Set
+We will apply the same logic to `ReplicatedBatchLoader` for validation sets. 
+
+*TODO: remove the following question*
+> Question: right now does Levanter supports a dataset that only has a training set but no validation set?
```
dlwh, line 92: not explicitly, though I think it might just work if you use 0 urls

docs/design/Multiple-Data-Mixture.md
```diff
@@ -0,0 +1,92 @@
...
+1. Users can specify the weight of each domain in the dataset, through the configuration file.
+2. At every batch, the weight of each domain is used to sample the number of tokens from the 
+corresponding domain.
+
+Ideally, we still want to preserve the reproducibility and deterministic batches of Levaner, 
+as specified in the [Data Loader design](Data-Loader-Design.md).
+
+
+## Design and Implementation
+### Configuration
```
dlwh, line 30: this section looks great

docs/design/Multiple-Data-Mixture.md
```diff
@@ -0,0 +1,92 @@
...
+      cache_dir: "gs://pubmed-mosaic/tokenized/reddit/"
+```
+
+Note that this design is backward compatible, as users can still specify a single dataset in the
+configuration file, and the weight of the dataset will be 1.0 by default.
+
+### Mixing at Batch Level
+Currently at training, `ShardedBatchLoader` takes in a single dataset of TokenSeqDataset class, 
+and implements the batching and sharding logic. Therefore, the mixing of multiple datasets should 
+be implemented at ShardedBatchLoader.
```
dlwh, line 70: I'd rather we do this at the dataset level: have a MixtureDataset that takes a list of datasets and weights and does the logic you describe. feels like a better separation of concerns

docs/design/Multiple-Data-Mixture.md
```diff
@@ -0,0 +1,92 @@
...
+- The [Pile dataset](https://pile.eleuther.ai/) is a mixture of 22 different datasets, including
+web data (24%), Wikipedia (9%), GitHub (4%), and more;
+- The [RedPajama dataset](https://huggingface.co/datasets/togethercomputer/RedPajama-Data-1T) is 
+a mixture of 1.2 tokens, from domains such as CommonCrawl (73%), GitHub (5%), Books (2.4%), 
+Wikipedia (2.4%), etc. 
+- The [Falcon 40B](https://huggingface.co/tiiuae/falcon-40b) model is trained on 1T tokens from 
+sources including RedfinedWeb English (75%), Books (6%), Conversations (5%), etc. 
+
+The common approach at training, used in LLaMA, MPT, and Falcon models, is to sample from each domain in 
+the dataset uniformly, based on the percentage of tokens in each domain. Recently, we have seen a new line
```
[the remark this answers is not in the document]
Ivan-Zhou: You are right: 
image

src/levanter/data/text.py
```diff
@@ -107,6 +126,60 @@ def load(pos: Axis, cache_dir: str, stride: Optional[int] = None) -> "TokenSeqDa
...
+        docs from doc_cache and yield batches of token sequences. We leverage this
+        implementation of iteration. This function mainly samples a dataset according to
+        the weights and call __iter__() function from corresponding TokenSeqDataset.
+
+        Yields:
+            Iterator[NamedArray]: _description_
+        """
+        while True:
+            dataset_index = self.sample_index(self.weights)
+            yield from self.token_seq_datasets[dataset_index]
```
Ivan-Zhou, line 167: Instead of yield from, we should call from a list of iterators

src/levanter/data/text.py
```diff
@@ -107,6 +126,60 @@ def load(pos: Axis, cache_dir: str, stride: Optional[int] = None) -> "TokenSeqDa
...
+    their associated weights.
+    We leverage the implementation of TokenSeqDataset on sharding and iterating over the data.
+
+    :param doc_cache_list: a list of TokenizedDocumentCache to draw from
+    :param Pos: the axis to use for the sequences. Sequences will be a NamedArray with axis Pos
+    :param weight_list: a list of weights for each dataset
+    """
+
+    def __init__(self, doc_caches: List, Pos: Axis, weights: List[float]):
+        self.token_seq_datasets = [TokenSeqDataset(doc_cache, Pos) for doc_cache in doc_caches]
```
Ivan-Zhou, line 141: We can make this more abstract

src/levanter/data/text.py
```diff
@@ -502,8 +562,14 @@ class LMDatasetConfig:
     def the_tokenizer(self) -> PreTrainedTokenizerFast:
         return load_tokenizer(self.tokenizer)
 
-    def build_or_load_cache(self, split: str, monitors: Union[bool, List[MetricsMonitor]] = True):
-        source = self.get_shard_source(split)
+    def build_or_load_cache(
+        self, split: str, monitors: Union[bool, List[MetricsMonitor]] = True
+    ) -> Optional[TokenizedDocumentCache]:
```
Ivan-Zhou, line 567: This is for handling cases when the dev split is not available (RP dataset, for example)

Ivan-Zhou:
@dlwh While I work on fixing the validation loss = 0, maybe you can start reviewing and let me know if there's any comment

src/levanter/data/text.py
```diff
@@ -175,6 +174,75 @@ def load(seq_len: int, cache_dir: str, stride: Optional[int] = None) -> "TokenSe
...
+        return MixtureDataset(sharded_doc_caches, self.seq_len, self.weights)
+
+    def __iter__(self) -> Iterator[np.ndarray]:
+        """TokenSeqDataset has a non-trivial implementation of __iter__() that iterates
+        docs from doc_cache and yield batches of token sequences. We leverage this
+        implementation of iteration. This function mainly samples a dataset according to
+        the weights and call __iter__() function from corresponding TokenSeqDataset.
+
+        Edge cases to consider:
+        1. Some of the iterators have been exhausted. We need to drop them and corresponding weights.
```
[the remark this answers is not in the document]
dlwh: we should make take inspiration from https://huggingface.co/docs/datasets/v2.14.0/en/package_reference/main_classes#datasets.interleave_datasets

src/levanter/data/text.py
```diff
@@ -580,6 +657,44 @@ def get_shard_source(self, split) -> ShardedDataSource[str]:
...
+    datasets: List[LMDatasetConfig]
+
+    def __post_init__(self):
+        if len(self.datasets) == 0:
+            raise ValueError("Must provide at least one dataset")
+
+        self.weights = [d.weight for d in self.datasets]
+
+        # ensure all datasets use the same tokenizer
+        if len(set([d.tokenizer for d in self.datasets])) != 1:
```
dlwh, line 675: maybe we shouldn't use LMDatasetConfig then.

What about LMDatasetSourceConfig that is just (splits of) URLs or a hf dataset it, then LMDatasetConfig takes a list of those (or map) along with weights as well as a tokenizer?

I'm not super concerned about making sure the configs are backwards compatible. Our user base is low enough ATM that we can break without making too many people mad, so I'd rather get the right thing while we still can

config/gpt2_data_mix.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - train_urls:
+          - "gs://pubmed-mosaic/openwebtext-sharded/openwebtext_train.{1..128}-of-128.jsonl.gz"
+      validation_urls:
+        - "gs://pubmed-mosaic/openwebtext-sharded/openwebtext_val.{1..8}-of-8.jsonl.gz"
+      cache_dir: "gs://levanter-data/tokenized/openwebtext/"
+      tokenizer: "gpt2"
+      weight: 0.5
+    - id: dlwh/wikitext_103_detokenized
+      cache_dir: "gs://levanter-data/tokenized/wikitext"
+      weight: 0.5
```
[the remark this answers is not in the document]
Ivan-Zhou: Yes I do agree that such interface would be nicer for tuning dataset weights. Is there any particular reason that you assign names for each task? 

Maybe an equivalent config for data mixture looks like this:

```
data-mixture:
  data_configs:
    pile:
      train_urls:
        - gs://levanter-data/pile/train/{00..29}.jsonl.zst
      validation_urls:
        - gs://levanter-data/pile/val.jsonl.zst
      cache_dir: "gs://levanter-data/tokenized/pile-old/"
      tokenizer: "EleutherAI/gpt-neox-20b"
    redpajama:
      id: togethercomputer/RedPajama-Data-1T
      cache_dir: gs://levanter-data/tokenized/redpajama/
      tokenizer: EleutherAI/gpt-neox-20b
      splits:
        - train
      rows_per_chunk: 4096
  data_weights:
    pile: 0.6
    redpajama: 0.4
```

I think this could be more flexible and cleaner than the current approach, which is trying to modify the existing `data` field while ensure compatibility (referring back to your comment about LMDatasetConfig).

src/levanter/data/text.py
```diff
@@ -175,6 +174,75 @@ def load(seq_len: int, cache_dir: str, stride: Optional[int] = None) -> "TokenSe
...
+                yield item
+            except StopIteration:
+                # if the iterator is exhausted, we need to drop the iterator and its weight
+                del token_seq_iterators[dataset_index]
+                del weights[dataset_index]
+
+    @staticmethod
+    def sample_index(weights):
+        """Sample a dataset according to the weights"""
+        return np.random.choice(len(weights), p=weights)
```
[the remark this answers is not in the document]
Ivan-Zhou: Acknowledge.

src/levanter/data/text.py
```diff
@@ -175,6 +174,75 @@ def load(seq_len: int, cache_dir: str, stride: Optional[int] = None) -> "TokenSe
...
+    """
+
+    def __init__(self, doc_caches: List, seq_len: int, weights: List[float]):
+        doc_caches, weights = self._check_doc_cache(doc_caches, weights)
+        self.doc_caches = doc_caches
+        self.token_seq_datasets = [TokenSeqDataset(doc_cache, seq_len) for doc_cache in doc_caches]
+        self.seq_len = seq_len
+        self.weights = self.normalize_weights(weights)
+
+    def _check_doc_cache(self, doc_caches, weights):
```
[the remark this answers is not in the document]
Ivan-Zhou: Acknowledge.

Ivan-Zhou:
It is time for me to re-pick up this PR. Given the amount of changes that have been made to the dataset and text, I might need to proceed forward with this PR or make a new one from a clean start. 

I summarized TODOs based on David's comments :
- Make `LMDatasetSourceConfig` or `LMDatasetMixtureConfig` that takes in a list of dataset configs as well as weights. It should support more than one tokenizer and provide a clean interface for tuning dataset weights. 
- Borrow `oversampling` argument and logic from datasets.interleave_datasets. I might need to take a closer look at their implementation. 
- Move `_check_doc_cache` outside of MixtureDataset and expect caches as input to MixtureDataset
- Create an instance rng as a member variable and a seed argument as well.

dlwh (approved):

PR #263, opened 2023-07-27
introduce a BackgroundIterable to load data in background thread

### LINKED ISSUE #252: Load data in a separate thread/process
Currently data loading happens on the main thread, it should happen in a separate thread/process.

For large models it's not a bottleneck, but it starts to matter on GPU for small models. Will also matter if we go multimodal.

We're also getting perf numbers high enough (on TPU) where one of the main easy wins is this.

dlwh:
fixes #252 

(still need to test on machines)


dlwh:
knocks 10-15% off step time at gpt2-small on a v3-32. almost certainly less substantial on larger models, but will be even more significant on GPU.

PR #270, opened 2023-08-02
Llama

Ivan-Zhou:
Implement Llama based on HF implementation and the paper. 

A few noteable difference from Gpt2:
- Rotary Positional Embedding;
- No dropout being used (therefore, many of the `call()` doesn't need key as input). 

Tasks
- [x] Add Llama code based on HF implementation
- [x] Refactor with Haliax
- [x] Match Levanter's state dict with HF's
- [x] Roundtrip test
- [ ] Ensure perf roughly matches GPT-2 implementation at scale


src/levanter/models/llama.py
```diff
@@ -0,0 +1,279 @@
...
+class LlamaRotaryEmbedding(eqx.Module):
+    dim: int
+    max_position_embeddings: int = 2048
+    base: float = 10000
+    inv_freq: jnp.ndarray = eqx.static_field()
+    cos_cached: jnp.ndarray = eqx.static_field()
+    sin_cached: jnp.ndarray = eqx.static_field()
+    max_seq_len_cached: int = eqx.static_field()
+
+    def __init__(self, dim, max_position_embeddings=2048, base=10000):
```
dlwh, line 116: this isn't quite as nice as i'd like.

```
def __init__(self, Embed: Axis, Pos: Axis, base=10000):
    ...
    self.inv_freq =  1.0 / (self.base ** (hax.arange(Embed.resize(Embed.size // 2), step=2)/Embed.size)
    self.cos_cached, self.sin_cached = self._set_cos_sin_cache(Pos)
```

src/levanter/models/llama.py
```diff
@@ -0,0 +1,279 @@
...
+        self.max_position_embeddings = max_position_embeddings
+        self.base = base
+        self.inv_freq = 1.0 / (self.base ** (jnp.arange(0, self.dim, 2) / self.dim))
+        self.cos_cached, self.sin_cached = self._set_cos_sin_cache(seq_len=self.max_position_embeddings)
+
+    def _get_positional_ids(self):
+        """A helper function for the convenience of extending to two sub-classes
+        Here we use a standard positional encoding function, which was described in `Attention is all you need`.
+        """
+        return jnp.arange(self.max_seq_len_cached)
```
dlwh, line 127: return hax.arange(self.Pos)

src/levanter/models/llama.py
```diff
@@ -0,0 +1,279 @@
...
+        t = self._get_positional_ids()
+
+        # Evaluates the Einstein summation convention on the operands.
+        freqs = jnp.einsum("i,j->ij", t, self.inv_freq)
+        # Different from paper but following HF implementation
+        # It uses a different permutation in order to obtain the same calculation
+        emb = jnp.concatenate((freqs, freqs), axis=-1)
+        cos_cached = jnp.cos(emb)[None, None, :, :]
+        sin_cached = jnp.sin(emb)[None, None, :, :]
+        return cos_cached, sin_cached
```
dlwh, line 140:
```python
freqs = t * self.inv_freq.broadcast_axis(self.Pos)
# or
freqs = hax.dot( (), t, self.inf_freq)

emb = hax.concatenate(self.Embed, (freqs, freqs))
cos_cached = hax.cos(emb)
sin_cached = hax.sin(emb)
```

src/levanter/models/llama.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def __call__(self, x, seq_len: int):
+        # x: [bs, num_attention_heads, seq_len, head_size]
+        if seq_len > self.max_seq_len_cached:
+            self.cos_cached, self.sin_cached = self._set_cos_sin_cache(seq_len=seq_len)
+
+        return (
+            self.cos_cached[:, :, :seq_len, ...],
```
dlwh, line 148: if you still do things this way (that is, supporting variable seq length with a cache), then:

`self.cos_cached[self.Pos, :seq_len]`

src/levanter/models/llama.py
```diff
@@ -0,0 +1,279 @@
...
+    output = jnp.concatenate((-x2, x1), axis=-1)
+    return hax.named(output, x.axes)
+
+
+def _apply_rotary_pos_emb(
+    q: NamedArray,  # [batch, seq_len, heads, head_size]
+    k: NamedArray,  # [batch, seq_len, kv_heads, head_size]
+    cos: jnp.ndarray,  # [1, 1, seq_len, head_size]
+    sin: jnp.ndarray,  # [1, 1, seq_len, head_size]
+    position_ids: jnp.ndarray,  # [bs, seq_len]
```
dlwh, line 268: ignore this for now and just always assume it's 0...seq_len

src/levanter/models/llama.py
```diff
@@ -0,0 +1,279 @@
...
+) -> Tuple[NamedArray, NamedArray]:
+    """Applies rotary position embedding to q and k."""
+    cos = jnp.squeeze(jnp.squeeze(cos, axis=1), axis=0)  # from [1, 1, seq_len, dim] to [seq_len, dim]
+    sin = jnp.squeeze(jnp.squeeze(sin, axis=1), axis=0)
+    cos = cos[position_ids]  # [batch, seq_len, head_size]
+    sin = sin[position_ids]  # [batch, seq_len, head_size]
+    cos = hax.named(cos, ("batch", "position", "head_size"))
+    sin = hax.named(sin, ("batch", "position", "head_size"))
+    q_embed = hax.multiply(q, cos) + hax.multiply(_rotate_half(q), sin)
+    k_embed = hax.multiply(k, cos) + hax.multiply(_rotate_half(k), sin)
```
dlwh, line 278: once cos and sin are named, replace the whole function with

q_embed = (q * cos + _rotate_half(q) * sin)
k_embed = (k * cos + _rotate_half(k) * sin)

if you still use position ids, then do 

cos = cos[Pos, position_ids]
sin = sin[Pos, position_ids]

first

src/levanter/models/llama.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def _rotate_half(x: NamedArray) -> NamedArray:
+    """Rotates half of the hidden dims of the input and concatenates them.
+    This is difficult to Haliax, so we do it in regular array ops.
+    """
+    x_array = x.array
+    x1 = x_array[..., : x_array.shape[-1] // 2]
+    x2 = x_array[..., x_array.shape[-1] // 2 :]
+    output = jnp.concatenate((-x2, x1), axis=-1)
```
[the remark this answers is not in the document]
Ivan-Zhou: Nice. This works perfectly well:

```
def _rotate_half(x: NamedArray) -> NamedArray:
    """Rotates half of the hidden dims of the input and concatenates them.
    """
    HeadSize = x.axes[-1]
    x1 = x[HeadSize, :HeadSize.size // 2]
    x2 = x[HeadSize, HeadSize.size // 2:]
    out = hax.concatenate(HeadSize, (-x2, x1))
    return out
```

src/levanter/models/llama.py
```diff
@@ -0,0 +1,404 @@
...
+from levanter.models.lm_model import LmConfig
+
+
+@LmConfig.register_subclass("llama")
+@dataclass(frozen=True)
+class LlamaConfig:
+    """Config for LlamaModel
+
+    Args:
+        seq_len (int, optional): maximum length of the input sequence. Defaults to 2048.
```
raisin, line 26: I believe max_position_embeddings is the maximum length of the input sequence in the LLamaConfig.

src/levanter/models/llama.py
```diff
@@ -0,0 +1,405 @@
...
+
+class LlamaTransformer(StateDictSerializationMixin, eqx.Module):
+    config: LlamaConfig = eqx.static_field()
+    layers: Stacked[LlamaDecoderLayer]
+    ln_f: hnn.LayerNorm
+
+    @staticmethod
+    def init(config: LlamaConfig, *, key) -> "LlamaTransformer":
+        # TODO: here it reports an error that is related to _get_rotary_emb() in LlamaAttention
+        # TypeError: Output from batched function Axis(name='head_size', size=4) with type <class 'haliax.axis.Axis'> is not a valid JAX type
```
Ivan-Zhou, line 276: `Stacked.init` reports an error that is related to _get_rotary_emb() in LlamaAttention. Here's the full stacktrace:

```
def test_llama_transformer():
        llama_config = _get_llama_config()
        key = random.PRNGKey(0)
>       llama_transformer = LlamaTransformer.init(config=llama_config, key=key)

tests/test_llama.py:156: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/levanter/models/llama.py:275: in init
    layers = Stacked.init(config.Layers, LlamaDecoderLayer, gradient_checkpointing=config.gradient_checkpointing)(
/nlp/scr/ivanzhou/miniconda3/envs/levanter/lib/python3.10/site-packages/haliax/nn/scan.py:69: in fn
    stacked = haliax.vmap(module.init, Block)(*args, **kwargs)
/nlp/scr/ivanzhou/miniconda3/envs/levanter/lib/python3.10/site-packages/haliax/hof.py:270: in wrapped_vmap_fn
    result = jax.vmap(
/nlp/scr/ivanzhou/miniconda3/envs/levanter/lib/python3.10/site-packages/jax/_src/api.py:1240: in vmap_f
    out_flat = batching.batch(
/nlp/scr/ivanzhou/miniconda3/envs/levanter/lib/python3.10/site-packages/jax/_src/linear_util.py:203: in call_wrapped
    ans = gen.send(ans)
/nlp/scr/ivanzhou/miniconda3/envs/levanter/lib/python3.10/site-packages/jax/_src/interpreters/batching.py:600: in _batch_inner
    out_vals = map(partial(from_elt, trace, axis_size), outs, out_dim_dests)
/nlp/scr/ivanzhou/miniconda3/envs/levanter/lib/python3.10/site-packages/jax/_src/interpreters/batching.py:237: in from_elt
    return matchaxis(trace.axis_name, axis_size, x_.batch_dim, spec, x_.val)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

axis_name = <object object at 0x7f22a3423e10>, sz = 32, src = None, dst = 0, x = Axis(name='head_size', size=4), sum_match = False

    def matchaxis(axis_name, sz, src, dst, x, sum_match=False):
      if dst == pile_axis:
        x = bdim_at_front(x, src, sz)
        elt_ty = x.aval.update(shape=x.shape[1:])
        aval = PileTy(core.Var(0, '', core.ShapedArray((), np.dtype('int32'))),
                      x.shape[0], elt_ty)
        return Pile(aval, x)
      try:
        _ = core.get_aval(x)
      except TypeError as e:
>       raise TypeError(f"Output from batched function {repr(x)} with type "
                        f"{type(x)} is not a valid JAX type") from e
E       jax._src.traceback_util.UnfilteredStackTrace: TypeError: Output from batched function Axis(name='head_size', size=4) with type <class 'haliax.axis.Axis'> is not a valid JAX type
E       
E       The stack trace below excludes JAX-internal frames.
E       The preceding is the original exception that occurred, unmodified.
```

dlwh: this is a new one for me. is this changed pushed?

dlwh: pretty sure it's this (or perhaps similar problems):

https://github.com/stanford-crfm/levanter/blob/6e222b7297250126cc17a8ab5c834c12ded592bf/src/levanter/models/llama.py#L97C3-L98C14

those need to be static fields

Ivan-Zhou: Will make a try

dlwh:
btw i fixed the problem in latest haliax. You should still make them static
fields, but you won't get an error there

Ivan-Zhou:
I've taken the time to address all the comments and made several refactoring changes to the code. Here are the notable updates:
- I've removed the Linear and Dynamic variants since they aren't required for V1. This allowed me to simplify the implementation of the LlamaRotaryEmbedding class.
- Following the suggestions from @dlwh, I refactored the Rotary Position Embedding implementation and re-wrote the get_cos_sin_cache and _apply_rotary_pos_emb methods in Haliax. All operations now utilize Haliax's NamedArray.
- I've removed position_ids, operating under the assumption that it is always [1, ... seq_len]. This eliminates the need for this argument in all classes and simplifies the _apply_rotary_pos_emb function considerably.
- All end-to-end tests and checks have passed. 

I'll conduct a round-trip test tomorrow, and this PR should be ready for review soon after.

Ivan-Zhou:
The following weights are present in HF’s state_dict, but not in Levanter’s. 

### `rotary_emb.inv_freq`
```python
model.layers.0.self_attn.rotary_emb.inv_freq: torch.Size([2])
model.layers.1.self_attn.rotary_emb.inv_freq: torch.Size([2])
model.layers.2.self_attn.rotary_emb.inv_freq: torch.Size([2])
model.layers.3.self_attn.rotary_emb.inv_freq: torch.Size([2])
```

`inv_freq` is present in the rotary positional embedding. According to the definition [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L99C15-L100), it doesn’t seem to be trainable weights. It is present in state_dict might because of register_buffer. 

```python
inv_freq = 1.0 / (self.base ** (torch.arange(0, self.dim, 2).float().to(device) / self.dim))
self.register_buffer("inv_freq", inv_freq, persistent=False)
```

### `lm_head.weight`
The `lm_head.weight` is present as the last linear layer in Llama Head Model in HF implementation. It is marked as `_tied_weights_keys`. It should be shared the weight with the text embedding to unembed and get logits. 

```c
lm_head.weight: torch.Size([1000, 16])
```

----
Other than these, the state_dict of Llama in Levanter matches with HF's.

Ivan-Zhou:
Actually, there is one issue: In linear layers, the weights in torch models have shape `(out_features,in_features); where as in Levanter, it seems the shape is `(in_features, out_features)`:

```
Key model.layers.0.mlp.gate_proj.weight has different values. First dictionary: torch.Size([11008, 8]), Second dictionary: (8, 11008)
```

This is a very standard Haliax operation, though. I don't quite understand why it shows mismatch here but not in GPT2 or other models.

Ivan-Zhou:
Oh I found the magic to do so:

```
my_dict.update(flatten_linear_layers(apply_prefix(prefix, "down_proj"), self.down_proj, out_dims_first_in_dict=True))
```

Ivan-Zhou:
The output of Levanter and HF models don't match. I will need to check module by module to find the source of inconsistency.

dlwh:
Ugh, yeah, this is one the hardest things to get right. I typically step through with a debugger side by side (turning off jit)

dlwh:
> This is a very standard Haliax operation, though. I don't quite understand why it shows mismatch here but not in GPT2 or other models.

gpt2 uses Conv1d modules which are (In, Out) rather than (Out, In). Knowing what I know now I'd change Haliax's Linear default order to be (Out, In) but I'm worried about subtle breakage at this point

src/levanter/models/llama.py
```diff
@@ -0,0 +1,475 @@
...
+        q, k = _apply_rotary_pos_emb(q, k, cos, sin)
+
+        scale = jax.lax.rsqrt(float(self.config.HeadSize.size))
+
+        # do this first to help keep FP values small
+        q = q * scale
+
+        q = q.astype(jnp.float32)
+        k = k.astype(jnp.float32)
+        k = k.rename({"position": "key_position"})
```
dlwh, line 249: you need to rename v's position too

dlwh: i think you can replace line s 240-258 with `haliax.nn.attention.dot_product_attention(config.Pos, config.Pos, config.HeadDim, q, k, v, mask)`

src/levanter/models/llama.py
```diff
@@ -0,0 +1,521 @@
...
+        seq_len (int, optional): maximum length of the input sequence. Defaults to 2048.
+        hidden_dim (int, optional): dimension of the hidden state. Defaults to 4096.
+        intermediate_dim (int, optional): dimension of the intermediate state. Defaults to 11008.
+        num_layers (int, optional): number of hidden layers in the Transformer encoder. Defaults to 32.
+        num_heads (int, optional): number of attention heads for each attention layer. Defaults to 32.
+        activation_function (str, optional): activation function for the hidden layer. Defaults to "silu".
+        rope_scaling (Dict, optional): dict containing the scaling configuration for the Rotary Positional Embedding.
+    """
+
+    # TODO(ivan): add tying of embeddings, default it to false to match the original model
```
dlwh, line 47: @Ivan-Zhou this is critical. We probably need to add a roundtrip test starting from an HF model to catch this.

src/levanter/models/llama.py
```diff
@@ -0,0 +1,521 @@
...
+    HeadSize = property(lambda self: Axis(name="head_size", size=self.hidden_dim // self.num_heads))
+    Intermediate = property(lambda self: Axis(name="intermediate", size=self.intermediate_dim))
+
+    @cached_classproperty
+    def default_hf_checkpoint_converter(cls) -> HFCheckpointConverter["LlamaConfig"]:  # type: ignore
+        return HFCheckpointConverter(
+            cls,  # type: ignore
+            "meta-llama/Llama-2-7b-hf",
+            trust_remote_code=True,
+            config_overrides={"tie_word_embeddings": True},
```
dlwh, line 80: (revert when we have the tests in place)

tests/test_llama.py
```diff
@@ -0,0 +1,289 @@
...
+    input_ids = hax.random.randint(random.PRNGKey(0), (Batch, Pos), 0, Vocab.size)
+    mask = hax.nn.attention.causal_mask(Pos, llama_config.KeyPos)
+
+    llama_model = LlamaLMHeadModel.init(Vocab=Vocab, config=llama_config, key=random.PRNGKey(0))
+    out = llama_model(input_ids, mask)
+    assert out.array.shape == (Batch.size, Pos.size, Vocab.size)
+
+
+@skip_if_no_torch
+def test_llama_roundtrip():
```
dlwh, line 223: make it start from an HF model (or add one that does)

dlwh (changes requested):

Ivan-Zhou:
Untie word_embeddings at LMHead is done. The only pending issue is Jax leakage when loading HF weight to Levanter's model.

tests/test_llama.py
```diff
@@ -0,0 +1,298 @@
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        torch_model.save_pretrained(f"{tmpdir}/torch_model")
+
+        model = converter.load_pretrained(LlamaLMHeadModel, f"{tmpdir}/torch_model")
+
+        def compute(input):
+            model_output = model(input, attn_mask=attn_mask)
+            return hax.nn.softmax(model_output, axis=model.Vocab)
+
+        # compute = jax.jit(compute)
```
dlwh, line 261:
```suggestion
        compute = jax.jit(compute)
```

dlwh (approved):

src/levanter/models/llama.py
```diff
@@ -480,7 +486,7 @@ def init(cls, Vocab: Axis, config: LlamaConfig, *, key) -> "LlamaLMHeadModel":
         k_t, k_emb = jrandom.split(key, 2)
         transformer = LlamaTransformer.init(config, key=k_t)
         embeddings = LlamaEmbedding.init(Vocab, config, key=k_emb)
-        lm_head = hnn.Linear.init(In=config.Embed, Out=Vocab, key=k_emb, use_bias=config.use_bias)
+        lm_head = hnn.Linear.init(In=config.Embed, Out=Vocab, key=k_emb, use_bias=False, out_first=True)
```
Ivan-Zhou, line 468: If we specify `out_first=True` at initializing lm_head (good to know such argument exists), do we still need the logic in `from_state_dict` and `update_state_dict`? I made those to revert the order in the weight.

tests/test_llama.py
```diff
@@ -0,0 +1,300 @@
...
+
+    torch_out = torch_model(input_torch)
+    torch_out = torch_out.logits[0].detach().cpu().numpy()
+    torch_out = jax.nn.softmax(torch_out, axis=-1)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        torch_model.save_pretrained(f"{tmpdir}/torch_model")
+
+        model = converter.load_pretrained(
+            LlamaLMHeadModel, f"{tmpdir}/torch_model", resize_vocab_to_match_tokenizer=False
```
Ivan-Zhou, line 256: Good to know that we can load from local checkpoint in this way.

src/levanter/compat/hf_checkpoints.py
```diff
@@ -477,10 +478,15 @@ def load_pretrained(
 
             # Vocab: next, we resize the desired actual size
             if Vocab.size != tokenizer_Vocab.size:
-                logger.info(
-                    f"Resizing model from {Vocab.size} to {tokenizer_Vocab.size} to match tokenizer vocab size"
-                )
-                lev_model = lev_model.resize_vocab(tokenizer_Vocab.size)
+                if resize_vocab_to_match_tokenizer:
```
Ivan-Zhou, line 481: In which case do we want to keep the vocab size different from tokenizer? @dlwh

dlwh: this is a good point. I was thinking it would mean keep whatever's in the model itself? WDYT?

PR #291, opened 2023-08-29
Alpaca script

dlwh:
script that basically replicates https://github.com/tatsu-lab/stanford_alpaca

I decided to try to keep the "engineering" for it pretty minimal to keep it approachable. It doesn't use the sophisticated preprocessing or data loading machinery, for example. Decided to punt on prefix lm.

Gonna wait on Llama to merge.

TODOs:
- [x] merge llama
- [ ] run alpaca on a v3-32 and make sure we can reproduce
- [x] save pretrained hf (and upload?)
- [ ] tutorial

### DIFF
--- a/examples/alpaca.py
+++ b/examples/alpaca.py
@@ -1,0 +1,253 @@
+# levanter version of https://github.com/tatsu-lab/stanford_alpaca/blob/main/train.py
+
+# We attempt to stick fairly close the original code, but there are some differences:
+# - We use the huggingface dataset version of alpaca rather than checking it in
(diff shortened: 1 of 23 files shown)


Ivan-Zhou:
The json file is signficantly large, shall we host on GCS instead?

dlwh:
i thought about it. alpaca checks it in which is why i did, but I agree with you... looks like it's on HF https://huggingface.co/datasets/tatsu-lab/alpaca so I'll do that

dlwh:
(currently relies on #298 )

PR #298, opened 2023-09-02
Add Llama training configs

Ivan-Zhou:
## Main changes
- Add Llama-2-7b (currently initialized from scratch, till when we fixed the resource exhaustion issue);
- Add Llama-2 nano for quick debugging and testing;
- Materialize the input mask, as I found that the mask from the training pipeline is of class `AttentionMask` instead of NameArray. 
----
### Training Throughput
The following profiling results are measured on TPU V4-64:

| Model Size | Tokens per Sec | Job Link |
|--------|--------|--------|
| Llama2-nano (4M) | 1,111,420 | [W&B](https://wandb.ai/stanford-mercury/levanter/runs/7vaq2yf2) |
| Llama2-1B | 123,860 | [W&B](https://wandb.ai/stanford-mercury/levanter/runs/6yoqcqpx/) |
| Llama2-3B | 66,913 | [W&B](https://wandb.ai/stanford-mercury/levanter/runs/r1b9yc9p) |
| Llama2-7B (batch size 32) | 33,142| [W&B](https://wandb.ai/stanford-mercury/levanter/runs/4gw2odfn) |
| Llama2-7B (batch size 256) | 63,655 | [W&B](https://wandb.ai/stanford-mercury/levanter/runs/xjl4dqb2/) |

Note:
1. Except for the last training job, all the training jobs have batch size of 32. 

### Training Loss Curve
Below is the training loss curve of Llama-7B that was initialized randomly. 
image


src/levanter/models/llama.py
```diff
@@ -471,18 +472,21 @@ def init(cls, Vocab: Axis, config: LlamaConfig, *, key) -> "LlamaLMHeadModel":
...
             input_ids (NamedArray): [batch, position]
                 Indices of input sequence tokens in the vocabulary.
-            attn_mask (NamedArray, optional): [batch, position, seq_len]
+            attn_mask (Union[NamedArray, AttentionMask], optional): [batch, position]
                 Mask to avoid performing attention on the padding token indices of the encoder input.
+                The attn_mask from training pipeline may be an AttentionMask object instead of NamedArray
         """
         x = self.embeddings.embed(input_ids)
+        if isinstance(attn_mask, AttentionMask):
+            attn_mask = attn_mask.materialize()
```
Ivan-Zhou, line 489: The attn_mask has to be a NamedArray; otherwise, it would cause an AssertionError in `hnn.attention.dot_product_attention`:

```
File "/home/ivan/levanter/src/levanter/models/llama.py", line 269, in __call__
    attn_output = hnn.attention.dot_product_attention(c.Pos, c.KeyPos, c.HeadSize, q, k, v, mask)
  File "/home/ivan/haliax/src/haliax/nn/attention.py", line 111, in dot_product_attention
    weights = dot_product_attention_weights(KeySize, KPos, query, key, mask, bias, attention_dtype, precision)
  File "/home/ivan/haliax/src/haliax/nn/attention.py", line 65, in dot_product_attention_weights
    weights = haliax.where(mask, weights, -1e9)
  File "/home/ivan/haliax/src/haliax/ops.py", line 41, in where
    assert isinstance(condition, NamedArray)
jax._src.traceback_util.UnfilteredStackTrace: AssertionError

The stack trace below excludes JAX-internal frames.
```

config/llama2_7b.yaml
```diff
@@ -0,0 +1,22 @@
+data: !include data/openwebtext_source.yaml
```
dlwh, line 1: can't use this b/c the tokenizer is different

src/levanter/models/llama.py
```diff
@@ -471,18 +472,21 @@ def init(cls, Vocab: Axis, config: LlamaConfig, *, key) -> "LlamaLMHeadModel":
     def __call__(
         self,
         input_ids: NamedArray,
-        attn_mask: Optional[NamedArray] = None,
+        attn_mask: Optional[Union[NamedArray, CausalMask]] = None,
```
dlwh, line 475: `AttnMask` please

src/levanter/models/llama.py
dlwh: probably should be a better error, but these "structured" masks are pretty important for ram usage when dealing with flash attention...

dlwh (commented):
we need a different dataset default for llama. otherwise lgtm

src/levanter/models/llama.py
Ivan-Zhou: @dlwh What would you recommend? Shall we pass the CausalMask from LMHead to LlamaAttention and materialize there? Or handle them in Haliax.attention?

config/llama2_7b.yaml
Ivan-Zhou: Yes, I will switch to a new dataset once I finished tokenization and caching.

src/levanter/models/llama.py
Ivan-Zhou: I have switched from CasualMask type to AttentionMask type. 
I also move the step of materialize inside LlamaAttention module, following the implementation at test_flash_attention [here](https://github.com/stanford-crfm/levanter/blob/f72a4c9018b90344dce56c25b39b69ee100f63a8/tests/test_flash_attention.py#L42).

dlwh: sounds good, thanks!

config/llama2_7b.yaml
```diff
@@ -0,0 +1,28 @@
+data:
+  train_urls:
+    - "gs://pubmed-mosaic/openwebtext-sharded/openwebtext_train.{1..128}-of-128.jsonl.gz"
+  validation_urls:
+    - "gs://pubmed-mosaic/openwebtext-sharded/openwebtext_val.{1..8}-of-8.jsonl.gz"
+  cache_dir: "gs://levanter-data/tokenized/openwebtext_llama/"
+  tokenizer: "meta-llama/Llama-2-70b-hf"
+model:
+  type: llama
+# TODO: uncomment this once we resolve the resource exhaustion issue
```
dlwh, line 10: this should be fixed now

dlwh (approved):

PR #304, opened 2023-09-11
remove device check that was breaking hf export

dlwh:


dlwh:
(wait on me to run itest)

PR #314, opened 2023-09-17
Improve Llama's training throughput

Ivan-Zhou:
This PR adds two changes to improve Llama's training throughput:
1. Not upcast q, k and attention outputs to float 32 by default. This is consistent with the implementation in HF. As a result of this, we are able to fit a larger batch size and therefore increase training throughput. We still keep an argument for upcasting if needed. 
2. reordering the position and heads in qkv matmul for better training throughput.

## Benchmarking
Below is the comparison of the training throughput before/after the changes, measured on v3-32:
- Before changes: 23.5K
- Turns off upcasting: 26.7K
- Reorder qkv: 27.0K

image

Below is the comparison of Llama's training throuhgput after the changes, vs GPT2-7B and MPT-7B:
- GPT-2 7B: 28.3K
- Llama-2 7B: 27.1K
- MPT 7B: 26.4K

image

## Testing
- [x] I have validated the changes with all 8 tests in the test_llama.py

### DIFF
--- a/config/llama2_7b.yaml
+++ b/config/llama2_7b.yaml
@@ -7,7 +7,6 @@ data:
   tokenizer: "meta-llama/Llama-2-70b-hf"
 model:
   type: llama
-  upcast_attn: false
 # TODO: uncomment this once we resolve the resource exhaustion issue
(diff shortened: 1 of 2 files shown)


dlwh (approved):
🎉

PR #321, opened 2023-09-24
Port-Models doc

Ivan-Zhou:
Add a documentation on how to add new models to Levanter. 

Screenshot 2023-09-24 at 3 33 48 PM


docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
+# Add New Models to Levanter
```
dlwh, line 1: maybe let's call it Porting Models to Levanter?

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+This guide outlines the process of adding new models into Levanter. While we emphasize adding models that are implemented in Hugging Face, these steps are applicable to models from other sources.
+
+We'll start with a detailed walkthrough on implementing your model and testing it. Subsequently, we'll describe how to configure and run a training job with your model. To conclude, we'll share insights and recommendations to enhance your model's training efficiency.
+
+## Write Your Model
+Writing a new model in Levanter is very similar to Hugging Face. You start by adding a new file in [models/](https://github.com/stanford-crfm/levanter/tree/main/src/levanter/models). You first create a config class to register the key hyperparameters and axes of your model. Then, you write the model class that includes key layers and components of your model.
```
dlwh, line 8: Maybe 'At a high level,' ? Writing with Haliax is pretty different...

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+**Hyperparameters** should be declared with their type and a default value, illustrated below:
+
+```python
+seq_len: int = 2048
+hidden_dim: int = 4096
+intermediate_dim: int = 11008
+```
+
+**Model axies** are used for parallelization. An Axis is registered with its name and size. The size of an Axis is normally associated with a hyperparameter. For example:
```
dlwh, line 23:
```suggestion
**Model axes** are used for parallelization. An Axis is registered with its name and size. The size of an Axis is normally associated with a hyperparameter. For example:
```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+This guide outlines the process of adding new models into Levanter. While we emphasize adding models that are implemented in Hugging Face, these steps are applicable to models from other sources.
+
+We'll start with a detailed walkthrough on implementing your model and testing it. Subsequently, we'll describe how to configure and run a training job with your model. To conclude, we'll share insights and recommendations to enhance your model's training efficiency.
+
+## Write Your Model
+Writing a new model in Levanter is very similar to Hugging Face. You start by adding a new file in [models/](https://github.com/stanford-crfm/levanter/tree/main/src/levanter/models). You first create a config class to register the key hyperparameters and axes of your model. Then, you write the model class that includes key layers and components of your model.
+
+### Write Config
+We start by writing your model config class. This class will register all the hyperparameters and axes of your models. We want to define them as the first step because they will be used immidately in the next step that implements your model.
```
dlwh, line 11: Somewhere we should mention registration with LmConfig for LM Models. We should also mention that configs should be a dataclass to work with the config parsing stuff.

Maybe we should just provide a template?

```python
@LmConfig.register_subclass("gpt2") # if implementing a Causal Language model
@dataclass(frozen=True)
class MyConfig(HFCompatConfig):  # Extend HFCompatConfig if you want to support HF import/export
```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+```python
+Pos = property(lambda self: Axis(name="position", size=self.seq_len))
+Embed = property(lambda self: Axis(name="embed", size=self.hidden_dim))
+Mlp = property(lambda self: Axis(name="mlp", size=self.intermediate_dim))
+```
+
+For real examples and deeper understanding, check out `Gpt2Config` in [gpt2.py](https://github.com/stanford-crfm/levanter/blob/main/src/levanter/models/gpt2.py) and `LlamaConfig` in [llama.py](https://github.com/stanford-crfm/levanter/blob/main/src/levanter/models/llama.py).
+
+To convert your config class to and from Hugging Face config class, you can write class functions `to_hf_config()` and `from_hf_config()` that maps the input parameters to the corresponding Hugging Face parameters. For example, in Llama, we have the following:
```
dlwh, line 33: probably here talk about HFCompatConfig

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+        num_heads=hf_config.num_attention_heads,
+        activation_function=hf_config.hidden_act,
+        initializer_range=hf_config.initializer_range,
+        layer_norm_epsilon=hf_config.rms_norm_eps,
+        rope_scaling=hf_config.rope_scaling,
+    )
+```
+
+Lastly, you should register your head model's class name as a class property. This step can be deferred until the head class is constructed.
+This class property would make it easier to call your model with the config class. For example:
```
dlwh, line 56: It's more that it makes it work with the automatic model selection stuff in train_lm and friends

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+```python
+@property
+def model_type(cls) -> Type["LlamaLMHeadModel"]:
+    return LlamaLMHeadModel
+```
+
+### Write Model
+After you have defined your config class, you can start writing your model.
+You can follow the breakdown of the model in your Hugging Face implmentation. This would make it easier to validate the implementation through unit tests.
```
dlwh, line 66:
```suggestion
You can follow the breakdown of the model in your Hugging Face implementation. This would make it easier to validate the implementation through unit tests.
```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+- `Gpt2Block`: a block of Gpt2Mlp and Gpt2Attention
+- `Gpt2Transformer`: a stack of Gpt2Block
+- `Gpt2Embeddings`: token and position embeddings
+- `Gpt2LMHead`: a complete GPT2 model with embedding, transformer, and LM head.
+
+We follow the same breakdown in the implementation of Llama in Levanter.
+
+#### Note on the Implementation Format
+- Each class will have its key layers and components defined as attributes and be initialized with a static method `init()`.
+- Each class will inherit from `StateDictSerializationMixin` from `torch_serialization` and Equinox's `Module` class.
```
dlwh, line 80: classes don't need to inherit `StateDictSerializationMixin` unless they have custom logic

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Mapping weights from/to Hugging Face
+To load weights from Hugging Face, you will need to write a class function `from_hf_state_dict()` in each of your model class. It takes in a Hugging Face state_dict and returns a Levanter state_dict.
+
+For implementation, there are a few helper classes from `torch_serialization` that you can use:
+- To add specific prefix to the keys of Hugging Face state_dict, you can use the helper function `apply_prefix()`. The prefix comes from the name of attributes defined at the beginning of your model class.
+- To unflatten the linear layers of Hugging Face, you can use the helper function `unflatten_linear_params()`.
+- To unstack the transformer blocks of Hugging Face, you can use the helper function `unstack_transformer_blocks()`.
```
dlwh, line 88: this isn't really right: `Gpt2MLP` doesn't need custom serialization logic.


```suggestion
### Serialization to/from State Dicts

PyTorch and Hugging Face Transformers use "state dicts" as their preferred serialization format, either as pickles or as the new [safetensors](https://github.com/huggingface/safetensors) format. A state dict is a Python `dict` with string keys and tensor values. The keys of the dict are json-ish "key paths" like `model.blocks.0.mlp.c_proj` and the values are the corresponding parameters for that key path.

You can read more about [PyTorch State Dicts here](https://pytorch.org/tutorials/recipes/recipes/what_is_state_dict.html).

Levanter has machinery for (de)serializing to and from state dicts. Simple cases are handled automatically, but often times custom logic is needed.

#### Easy Case: Identical Module Structure

If your module has exactly the same fields with the same names and same shapes as the Hugging Face state dict (e.g. Gpt2Mlp), you don't need to do anything.

#### Medium Case: Different Names

If for some reason you want to use different names from the HF implementation (e.g. because the names from HF aren't clear...), you can implement `StateDictSerializationMixin` and use `_state_dict_key_map` to rename keys. For instance, the `Gpt2Transformer` class has this method:

```python
class Gpt2Transformer(StateDictSerializationMixin, eqx.Module):

    ...
    
    def _state_dict_key_map(self) -> Dict[str, Optional[str]]:
        return {"blocks": "h"}
```

This says that the field called `blocks` in this class should be (de)serialized as `h`, because the HF GPT-2 implementation uses `h`, which in our opinion isn't very clear. You can also "flatten" the submodules of a field by using `None` or even include `.`s in the name if needed.

#### Hard Case: Custom Serialization

If your modules need special logic, you'll need to implement class `StateDictSerializationMixin` and override `to_state_dict` and `from_state_dict`. It takes in a Hugging Face state dict and returns a Levanter state_dict.

For implementation, there are a few helper classes from `torch_serialization` that you can use:
- To add specific prefix to the keys of a state dict, you can use the helper function `apply_prefix()`. The prefix comes from the name of attributes defined at the beginning of your model class.
- To unflatten the linear layers of a state dict, you can use the helper function `unflatten_linear_params()`.
- To unstack the transformer blocks of a state dict, you can use the helper function `unstack_transformer_blocks()`.
```

File: docs/Add-Models.md [hunk 10]
```diff
@@ -0,0 +1,270 @@
...
+
+For example, below is the implementation of `from_hf_state_dict()` in LlamaAttention:
```
dlwh, line 90:
```suggestion
For example, below is the implementation of `from_state_dict()` in `LlamaAttention`. `LLamaAttention`, like most attention layers in Levanter projects the embeddings (with shape `(Pos, Embed)`) to a tensor of shape `(Pos, Head, HeadSize)` rather than the convention in Hugging Face Transformers to use `(Pos, Head * HeadSize)` and then reshaping (because PyTorch Linear doesn't support multiple input or output dimensions).

This difference in convention means that we have to flatten linear layers when converting to a Hugging Face Transformers-compatible state dict and unflatten them when they are read in.
```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+    d = {}
+    d.update(unflatten_linear_layers(apply_prefix(prefix, "q_proj"), state_dict, self.q_proj, True))
+    d.update(unflatten_linear_layers(apply_prefix(prefix, "k_proj"), state_dict, self.k_proj, True))
+    d.update(unflatten_linear_layers(apply_prefix(prefix, "v_proj"), state_dict, self.v_proj, True))
+    d.update(unflatten_linear_layers(apply_prefix(prefix, "o_proj"), state_dict, self.o_proj, True))
+
+    return super().from_state_dict(d, prefix)
+```
+
+Similarly, to save weights to Hugging Face, you will need to write a class function `to_hf_state_dict()` in each of your model class.
```
dlwh, line 104:
```suggestion
Similarly, to save weights to Hugging Face, you will need to write a class function `to_state_dict()` in each of your model class.
```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+The correctness of your implementation can be validated through serialization tests, which will be discussed in the next section.
+
+## Write Tests
+There are two types of tests that you should write for your model: unit tests and serialization tests.
+
+### Unit Tests
+Unit tests are very useful for testing the correctness of your model implementation.
+It is recommended to have at least one test for each of your modules, so that you can make sure that each module is working as expected and capture any surprises early on, before you test them end-to-end.
```
dlwh, line 113: I honestly don't think this is necessary but it's useful to do if you get stuck to minimize.

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+In a unit test, you test your module in the following aspects:
+1. The forward pass is successful.
+2. The output shape is correct.
+3. The output value is correct.
+
+The point 3 is the most important one, but it is also the most difficult one to implement. If you have a reference implementation in another framework, like HuggingFace, you can use it to test your implementation on the output consistency.
```
dlwh, line 120:
```suggestion
The point 3 is the most important one, but it is also the most difficult one to implement. If you have a reference implementation in another framework, like Hugging Face Transformers, you can use it to test your implementation on the output consistency.
```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+Here is a piece of example code to test a same module of Levanter against HuggingFace:
```
dlwh, line 121:
```suggestion
Here is a piece of example code to test a same module of Levanter against Hugging Face Transformers:
```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+levanter_output = levanter_rope(seq_len=seq_len)
+hf_rope = HFLlamaRotaryEmbedding(dim=hidden_dim, max_position_embeddings=seq_len, device="cpu")
+hf_output = hf_rope(x_torch, seq_len=seq_len)
+
+# compare the output values
+for jax_out, torch_out in zip(levanter_output, hf_output):
+    torch_out = torch_out.numpy()
+    assert np.isclose(torch_out, np.array(jax_out.array), rtol=1e-4, atol=1e-4).all(), f"{torch_out} != {jax_out}"
+```
+
```
dlwh, line 137:
```suggestion

Be sure to use `np.isclose` with reasonably loose tolerances to account for numerical differences.

```

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+The serialization tests are very useful for testing the correctness of your implementation and make sure you can load your pretrained HuggingFace model into Levanter.
+
+## Training
+After you have implemented your model and validated it through tests, you can start training your model with Levanter.
+
+### Write Training Configuration
+To launch a training job, you will need to write a training configuration file in yaml. It includes the dataset, model, and trainer specifications for the training job. You can find many examples in [configs/](https://github.com/stanford-crfm/levanter/tree/main/config).
+
+Under the model section, you will need to specify the model name as `type` and modify the hyperparameters that you would like to change. For parameters that are not specified, the default values will be used.
```
dlwh, line 215: the `type` is needed if registered as an `LmConfig`

docs/Add-Models.md
```diff
@@ -0,0 +1,270 @@
...
+
+Check out [Training on Your Own Data](./Training-On-Your-Data.md) for more detailed guide on how to spin off a training cluster and launch a training job.
+
+### Profile Your Model
+If you are interested in profiling the training throughput of your model, good news is that it comes for free with automatic job monitoring in Levanter, powered through Weights & Biases.
+
+Once you run a training job, on the corresponding job page on Weights & Biases, you will be able to find a section named "Throughput". It reports metrics like `examples_per_second` and `tokens_per_second` across the training time.
+
+## Tips for Optimization
+1. Avoid upcasting to float32. Levanter uses float16 by default, which is more memory efficient and faster for training. You should avoid upcasting to float32 unless it is necessary.
```
dlwh, line 263:
```suggestion
1. Avoid upcasting to float32. Levanter uses bfloat16 by default, which is more memory efficient and faster for training. You should avoid upcasting to float32 unless it is necessary for stability or accuracy.
```

mkdocs.yml
```diff
@@ -84,6 +84,7 @@ nav:
       - 'Getting-Started-CUDA.md'
   - 'User Guide':
       - 'Getting-Started-Training.md'
+      - 'Add-Models.md'
```
dlwh, line 87: let's move this to the end, or maybe add a "Developer Guide" section

dlwh:
Thanks! a few suggestions. I think it would be good to think about three slightly different uses cases:

1. Porting a new model of any sort (we might do vision soon maybe?)
2. Porting some causal LM model (the LmConfig stuff)
3. Porting an HF causal LM model (the HfCompatConfig stuff and serialization stuff)

docs/Add-Models.md
Ivan-Zhou: What're the examples of the custom logic?

Ivan-Zhou: Oh I found it below

Ivan-Zhou:
Thanks for your comments and detailed suggestions, @dlwh! Please take another look when you have time. 

This is very detailed instruction. I am bit concerned that people might get lost and this may become difficult to maintain over the long run. But I am sure people who want to port a model today will find it very helpful.

docs/Port-Models.md
```diff
@@ -0,0 +1,318 @@
+# Porting Models to Levanter
+
+## Overview
+This guide outlines the process of porting new models into Levanter. While we emphasize adding a causal language model (LM) that is implemented in Hugging Face, these steps are applicable to general models from multiple sources.
```
Ivan-Zhou, line 4:
```suggestion
This guide outlines the process of porting new models into Levanter. While we emphasize the scenario of adding a causal language model (LM) that is implemented in Hugging Face, this guide is applicable to general models from all sources.
```

docs/Port-Models.md
```diff
@@ -0,0 +1,318 @@
...
+            activation_function=hf_config.hidden_act,
+            initializer_range=hf_config.initializer_range,
+            layer_norm_epsilon=hf_config.rms_norm_eps,
+            rope_scaling=hf_config.rope_scaling,
+        )
+```
+
+#### Register Your Model
+Lastly, there are a few steps to register a model in Levanter and make it tightly integrated with the training pipeline:
+1. For language models, You can register your config class with `LmConfig.register_subclass("ModelName")`. By doing so, the trainer can use the name as the `model.type` field from the training configuration file to identify your config and model.
```
Ivan-Zhou, line 67:
```suggestion
1. For language models, You can register your config class with `LmConfig.register_subclass("ModelName")`. By doing so, the trainer can use the name from the `model.type` field in the training configuration file to identify your config and model.
```

docs/Port-Models.md
```diff
@@ -0,0 +1,318 @@
...
+- `Gpt2Block`: a block of Gpt2Mlp and Gpt2Attention
+- `Gpt2Transformer`: a stack of Gpt2Block
+- `Gpt2Embeddings`: token and position embeddings
+- `Gpt2LMHead`: a complete GPT2 model with embedding, transformer, and LM head.
+
+We follow the same breakdown in the implementation of Llama in Levanter.
+
+#### Note on the Implementation Format
+- Each class will have its key layers and components defined as attributes and be initialized with a static method `init()`.
+- If Each class will be extended from Equinox's `Module` class.
```
Ivan-Zhou, line 99:
```suggestion
- Each class will be extended from Equinox's `Module` class.
```

docs/Port-Models.md
```diff
@@ -0,0 +1,318 @@
...
+
+```python
+class Gpt2Transformer(StateDictSerializationMixin, eqx.Module):
+    ...
+
+    def _state_dict_key_map(self) -> Dict[str, Optional[str]]:
+        return {"blocks": "h"}
+```
+
+This says that the field called `blocks`` in this class should be (de)serialized as `h``, because the Hugging Face GPT-2 implementation uses `h``, which is not very clear. You can also "flatten" the submodules of a field by using `None`` or even include `.s` in the name if needed.
```
Ivan-Zhou, line 123:
```suggestion
This says that the field called `blocks` in this class should be (de)serialized as `h`, because the Hugging Face GPT-2 implementation uses `h`, which is not very clear. You can also "flatten" the submodules of a field by using `None` or even include `.s` in the name if needed.
```

docs/Port-Models.md
```diff
@@ -0,0 +1,318 @@
...
+
+Similarly, to save weights to Hugging Face, you will need to write a class function `to_state_dict()` in each of your model class.
+
+The correctness of your implementation can be validated through serialization tests, which will be discussed in the next section.
+
+## Write Tests
+There are two types of tests that you should write for your model: unit tests and serialization tests.
+
+### [Optional] Unit Tests
+Unit tests are very useful for testing the correctness of your model implementation.
```
Ivan-Zhou, line 158:
```suggestion
```

docs/Port-Models.md
```diff
@@ -0,0 +1,318 @@
...
+    mask = hax.nn.attention.causal_mask(config.Pos, config.KeyPos)
+    return x, mask
+```
+
+If your module contains model weights, such consistency tests would require you to load weight from your module in Levanter to the reference module in HuggingFace. We will discuss how to do that in the next section.
+
+### Serialization Tests
+
+In order to validate the implementation on Levanter matches with the reference implementation in HuggingFace, we would like to pass the same input to both implementations and compare the output.
+However, if the model are initialized with different weights, such comparison would not be meaningful. Therefore, we will need to serialize the weights from HuggingFace and load them into Levanter, or vice versa. We call such test "serialization test".
```
Ivan-Zhou, line 204:
```suggestion
However, if the model are initialized with different weights, such comparison would not be meaningful. Therefore, we will need to serialize the weights from Hugging Face transformers and load them into Levanter, or vice versa. We call such test "serialization test".
```

docs/Port-Models.md
```diff
@@ -0,0 +1,318 @@
...
+
+For modules like Attention, Mlp, and Embeddings, you can read the weight from Levanter in memory and load into corresponding modules in HuggingFace. For example, in Llama, we did the following:
```
Ivan-Zhou, line 206:
```suggestion
For modules like Attention, Mlp, and Embeddings, you can read the weight from Levanter as state dict in memory and load into corresponding modules in HuggingFace. For example, in Llama, we did the following:
```

dlwh:
> This is very detailed instruction. I am bit concerned that people might get lost and this may become difficult to maintain over the long run. But I am sure people who want to port a model today will find it very helpful.

That's a fair point, and it's also kind of intimidating as a result.

What do you think we should do?

dlwh:
maybe we should add high level guidance that you should pattern match off of llama and mpt?

dlwh (approved):

Ivan-Zhou:
Let's keep iterating on both the doc as well as the related codebase. 

With GPT-2, MPT, and Llama serve as examples and this doc as a guide, developers should be able to figure out a way to port general LM models. If we see frequent needs of 3rd party developers on porting new models, we can look after their pain points and address them accordingly.

Issue #330, opened 2023-09-27
Guide for running Alpaca/Alpaca LORA on GPU


dlwh:
Mostly we just want to be sure the code runs and the configs are not embarrassing.

Ideally we'd have Alpaca running on a full sphinx and Alpaca LORA running on the smallest GPU that works (single 3090?)

PR #338, opened 2023-10-04
Ray Init num_cpus bug. 

vadam5:
When alpaca training is run on a sphinx node, the following error occurs during ray initialization:

```
2023-09-29 10:59:14,261	INFO worker.py:1458 -- Connecting to existing Ray cluster at address: sphinx4:64053...
2023-09-29 10:59:14,276	INFO worker.py:1633 -- Connected to Ray cluster. View the dashboard at http://127.0.0.1:8265 
[2023-09-29 11:00:15,624 E 2740790 2740790] core_worker.cc:203: Failed to register worker 01000000ffffffffffffffffffffffffffffffffffffffffffffffff to Raylet. IOError: [RayletClient] Unable to register worker with raylet. No such file or directory
```

The solution is to explicitly set the number of CPUs when initializing ray, then training begins without issue. 

Internally related issue: https://github.com/stanford-crfm/levanter/issues/330

Related external issues: 
- https://discuss.ray.io/t/ray-on-slurm-problems-with-initialization/6361
- https://github.com/ray-project/ray/issues/21479

### DIFF
--- a/src/levanter/distributed.py
+++ b/src/levanter/distributed.py
@@ -150,7 +150,6 @@ def auto_ray_cluster(
     We don't use that because it's more geared towards submitting jobs to a ray cluster that is backed by slurm.
     Instead, we have our machines already.
     """
-    print("\n\nENTERED AUTO RAY CLUSTER\n\n")
     global _already_initialized
 
     if _already_initialized:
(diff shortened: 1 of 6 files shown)


src/levanter/distributed.py
```diff
@@ -190,8 +191,15 @@ def _munge_address_port(address: str):
                 address = f"{host}:{ray_port}"
 
                 if cluster_type.get_process_id() == 0:
+                    # Explicitly setting the number of CPUs on ray init stops init errors
+                    try:
+                        num_cpus = os.getenv("SLURM_CPUS_PER_TASK")
+                    except:
+                        num_cpus = logical_cpu_core_count()
```
dlwh, line 198: i think you should also set this for the other `ray start` on line 210

dlwh (approved):
thanks!

PR #350, opened 2023-10-17
Factor out an attention method everyone can call, expose flash attention

dlwh:


dlwh:
@vadam5 merging so i can cut release.

Issue #360, opened 2023-11-02
Gracefully handle case when there's no validation set


Ivan-Zhou:
RPv1 and v2 only have train set, which caused the following error. 
```
Traceback (most recent call last):
  File "/home/ivan/levanter/src/levanter/main/train_lm.py", line 186, in <module>
    levanter.config.main(main)()
  File "/home/ivan/levanter/src/levanter/config.py", line 84, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "/home/ivan/levanter/src/levanter/main/train_lm.py", line 103, in main
    eval_dataset = CausalLmDataset(config.data.token_seq_dataset("validation", Pos.size), Pos, KeyPos)
  File "/home/ivan/levanter/src/levanter/data/text.py", line 549, in token_seq_dataset
    cache = self.build_or_load_cache(split, monitors=monitors)
  File "/home/ivan/levanter/src/levanter/data/text.py", line 555, in build_or_load_cache
    source = self.get_shard_source(split)
  File "/home/ivan/levanter/src/levanter/data/text.py", line 476, in get_shard_source
    ds = WrappedHFDataset(self.id, split=split, name=self.name, streaming=self.stream)
  File "/home/ivan/levanter/src/levanter/data/sharded_dataset.py", line 150, in __init__
    self._shard_names = self._compute_shard_names()
  File "/home/ivan/levanter/src/levanter/data/sharded_dataset.py", line 157, in _compute_shard_names
    dataset = self._load_dataset()
  File "/home/ivan/levanter/src/levanter/data/sharded_dataset.py", line 183, in _load_dataset
    return datasets.load_dataset(self.id, split=self.split, **self.kwargs)
  File "/home/ivan/venv310/lib/python3.10/site-packages/datasets/load.py", line 1784, in load_dataset
    return builder_instance.as_streaming_dataset(split=split)
  File "/home/ivan/venv310/lib/python3.10/site-packages/datasets/builder.py", line 1271, in as_streaming_dataset
    raise ValueError(f"Bad split: {split}. Available splits: {list(splits_generators)}")
ValueError: Bad split: validation. Available splits: ['train']
```

We need to think of a way of handling this. 
1. Not set val set and skip validation loss
2. Allocate 1K train examples to val set.

dlwh:
My preference would be to tackle this as part of #217. Basically we should go through each named source (e.g. webtext, wiki) and add a validation callback for each source that has a validation set.

PR #364, opened 2023-11-07
add support for multiple eval sets (including n = 0)

### LINKED ISSUE #217: Support Multiple Validation Sets
Should be able to plug in an arbitrary number of validation sets

dlwh:
Fixes #217 #360

### DIFF
--- a/config/gpt2_nano_mixture.yaml
+++ b/config/gpt2_nano_mixture.yaml
@@ -1,0 +1,30 @@
+data:
+  configs:
+    wikitext:
+      id: dlwh/wikitext_103_detokenized
+    w2:
+      id: dlwh/wikitext_103_detokenized
+  train_weights:
+    wikitext: 1.0
+    w2: 0
+model:
+  type: gpt2
+  hidden_dim: 32
+  num_heads: 4
+  num_layers: 2
+trainer:
+  mp: f32
+  num_train_steps: 100
+
+  checkpointer:
+    keep:
+      - every: 50
+    save_interval: 5m
+
+  per_device_eval_parallelism: 1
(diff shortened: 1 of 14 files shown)


src/levanter/callbacks.py
```diff
@@ -23,11 +23,16 @@
 logger = logging.getLogger(__name__)
 
 
-def eval_loss_loop(loss_fn, model, dataset, max_batches: Optional[int] = None):
+def eval_loss_loop(loss_fn, model, dataset, max_batches: Optional[int] = None, name: Optional[str] = None):
     total_loss = 0.0
     n = 0
 
-    pbar = tqdm(dataset, desc="eval", position=1, leave=False)
+    if name is not None:
```
Ivan-Zhou, line 30: Not sure if I capture the logic correctly, but based on [here](https://github.com/stanford-crfm/levanter/blob/1f597337c079e7960ac3ac6ba9f1b33d3c1d6030/src/levanter/data/text.py#L567), it seems that for single dataset, the eval set name will be `""` instead of None, right?

config/gpt2_nano_mixture.yaml
```diff
@@ -0,0 +1,30 @@
+data:
+  configs:
+    wikitext:
+      id: dlwh/wikitext_103_detokenized
+    w2:
+      id: dlwh/wikitext_103_detokenized
```
Ivan-Zhou, line 6: If we use different names, this will make two eval loss curves, despite that these are the same datasets. Nice trick.

src/levanter/callbacks.py
dlwh: thanks!

PR #381, opened 2023-12-07
support callbacks inside JIT boundary, include grad histograms

dlwh:


dlwh:
add  in https://github.com/stanford-crfm/levanter/pull/641

PR #389, opened 2023-12-18
Sniff format from file content.

tjingrant:
This patch double checks whether URLs ending in ".json" are actually json; sometimes they can be jsonl, in which case I adjusted the sniffed format accordingly.


dlwh:
Hi @tjingrant . Thanks! I thought about this a bit and I don't love always parsing the entire file just to figure out the format. WDYT about the somewhat more heuristic approach in #394 ?

tjingrant:
Makes sense, and confirm that c4 formatting follows the heuristic! Closing.

PR #405, opened 2024-01-04
Add weight decay masking for optimizer

anh-tong:
Hi,

Thanks for the awesome library.

I've been training GPT models using Levanter. As noted from nanoGPT repo, weight decay masking is applied on the weights of transformers and embeddings excepts layer norms and biases. 

Since Levanter does not support this yet, I try to make one in this PR.  I added mask parameter `weight_decay_mask` for `OptimizerConfig.build` that allows to customize maskings. 

Currently, I use `equinox.tree_at` to select which nodes of PyTree to apply the weight decay. I wonder if there is a better way.

### DIFF
--- a/src/levanter/trainer.py
+++ b/src/levanter/trainer.py
@@ -708,10 +708,11 @@ class OptimizerConfig:
     """fraction of training steps to use as cooldown, or steps to use. 0.0 means no cooldown"""
     lr_schedule: str = "cosine"  # constant, cosine, linear
 
-    def build(self, num_train_steps: int) -> GradientTransformation:
+    def build(self, num_train_steps: int, weight_decay_mask=None) -> GradientTransformation:
         """Creates the optimizer"""
+
         # indirection makes it work with optax.inject_hyperparams so we can log the learning rate
-        def _optimizer(learning_rate):
+        def _optimizer(learning_rate, weight_decay_mask):
             components = []
 
             if self.max_grad_norm:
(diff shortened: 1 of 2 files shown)


dlwh:
HI @anh-tong, thanks for putting this together! It would be good to get some kind of weight decay masking in.

That said, I don't love coupling the trainer/optimizer with the specifics of the gpt2 model.

I could see a few paths forward:
1. add a spec that lets you add weight decay based on a regex on the paths ([similar to what we do for lora](https://github.com/stanford-crfm/levanter/blob/15e223d54ccffe8bbb36af943cfd9f19bc03187d/src/levanter/lora.py#L11-L13))
2. default to weight decaying based on number of dimensions which is what nanogpt actually does. This gets a bit more subtle for us because of layer stacking, which I increasingly regret.
3. add some kind of registry for different layer types/allow layers to declare how they want to do biasing.
 
I think tree_at is the best we can do, though you can see how we do things for the lora masking.

I'm kind of inclined toward (2), maybe with the ability to do (1) (eventually). WDYT?

anh-tong:
Thank you for your insight, @dlwh .  The regex approach seems to be a very good way that can configure masks. I will try to implement a similar version of Lora for the case of weight decay here. 

It seems like masking based on dimension like nanoGPT can be done like lora case by checking if the module is instance of `hnn.Stacked` as well.

I don't know much about registry but look forward seeing it in lora later.

anh-tong:
@dlwh  From your suggestion, I have updated the code that can work for the case 1. Unfortunately, for the default case that might use dimensions, the dimension of `c_proj.weight` and `c_attn.bias` in `Gpt2Attention` are the same. So we can not distinguish these two, making it not possible to implement this case.

dlwh (approved):
i like this! Probably we could do something simple like make the weight field of Linear and Embeddings be decayed by default or something, but this is a good feature for now!

dlwh:
hi anh, the test doesn't pass. can you fix please?

dlwh (changes requested):
please make the test pass.

anh-tong:
Oop sorry! I've corrected the regex and cleaned up the code a bit.

dlwh (approved):
thanks! really appreciate it!

PR #406, opened 2024-01-10
Slurm cluster fix

dlwh:


dlwh:
lol it's apparently in merged into main already (cc @DwarKapex )

PR #410, opened 2024-01-12
Add Feature: Group Query Attention

blahBlahhhJ:
Adds support for GQA which is required by llama2-70B.

This PR is still WIP. Code is not yet tested.

Since this is my first PR, I'm not so sure on some conventions in this project stack (haliax, jax, etc.)
* I think when specifying axis, we can either use the axis object or the axis name (string)? What's the convention here? Should I always aim to use the axis object if possible?

Feel free to give any suggestions :)

### DIFF
--- a/src/levanter/models/llama.py
+++ b/src/levanter/models/llama.py
@@ -49,7 +49,7 @@ class LlamaConfig(HFCompatConfig):
         num_heads (int, optional): number of attention heads for each attention layer. Defaults to 32.
         num_kv_heads (int, optional): number of attention heads for keys and values in each attention layer.
             Setting to 1 means MQA. Setting to num_heads means MHA. Otherwise GQA.
-            Note that this value must be divisible by num_heads. Defaults to 32.
+            Note that num_heads must be divisible by this number. Defaults to 32.
         activation_function (str, optional): activation function for the hidden layer. Defaults to "silu".
         rope_scaling (Dict, optional): dict containing the scaling configuration for the Rotary Positional Embedding.
     """
--- a/src/levanter/models/llama.py
+++ b/src/levanter/models/llama.py
@@ -47,8 +47,8 @@ class LlamaConfig(HFCompatConfig):
         intermediate_dim (int, optional): dimension of the intermediate state. Defaults to 11008.
         num_layers (int, optional): number of hidden layers in the Transformer encoder. Defaults to 32.
         num_heads (int, optional): number of attention heads for each attention layer. Defaults to 32.
-        num_kv_heads (int, optional): number of attention heads for keys and values in each attention layer. 
-            Setting to 1 means MQA. Setting to num_heads means MHA. Otherwise GQA. 
+        num_kv_heads (int, optional): number of attention heads for keys and values in each attention layer.
+            Setting to 1 means MQA. Setting to num_heads means MHA. Otherwise GQA.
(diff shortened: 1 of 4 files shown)


src/levanter/models/llama.py
```diff
@@ -47,6 +47,9 @@ class LlamaConfig(HFCompatConfig):
         intermediate_dim (int, optional): dimension of the intermediate state. Defaults to 11008.
         num_layers (int, optional): number of hidden layers in the Transformer encoder. Defaults to 32.
         num_heads (int, optional): number of attention heads for each attention layer. Defaults to 32.
+        num_kv_heads (int, optional): number of attention heads for keys and values in each attention layer.
+            Setting to 1 means MQA. Setting to num_heads means MHA. Otherwise GQA.
+            Note that num_heads must be divisible by this number. Defaults to 32.
```
dlwh, line 52: can you add a __post_init__ that checks this?

src/levanter/models/llama.py
```diff
@@ -288,8 +298,8 @@ def __call__(self, x: NamedArray, mask: Optional[NamedArray], *, key=None) -> Na
         if self.config.upcast_attn:
             q = q.astype(jnp.float32)
             k = k.astype(jnp.float32)
-        k = k.rename({"position": "key_position"})
-        v = v.rename({"position": "key_position"})
+        k = k.rename({"position": "key_position"}).broadcast_axis(self.config.QHeadsPerGroup)
+        v = v.rename({"position": "key_position"}).broadcast_axis(self.config.QHeadsPerGroup)
```
dlwh, line 302: i think you can get away without this, but I could be wrong?

dlwh (commented):
style all seems fine. I think there's going to be some pain with de/serialization with HF checkpoints so we should make sure that roundtrips.

i think the broadcasts can all be removed? I *think* dot product attention should go through unmodified since q_heads_per_group is effectively a batch dimension. Let me know if I'm wrong

src/levanter/models/llama.py
blahBlahhhJ: you're right. in that case, should I define the `QHeadsPerGroup` axis in the attention's `init` function instead of in the config cuz we don't need to reference this axis object other than in the `init`.

dlwh: sure that makes senes

blahBlahhhJ: I guess the difference is that we'll end up defining this axes using the same name on every layer. Is this considered bad? should the axis be somewhat a "global" thing as it's defined in the config?

dlwh: i think it's fine either way. Up to you what seems cleaner

src/levanter/models/llama.py
blahBlahhhJ: done.

src/levanter/models/llama.py
blahBlahhhJ: done.

dlwh (approved):
lgtm!

dlwh:
> I think when specifying axis, we can either use the axis object or the axis name (string)? What's the convention here? Should I always aim to use the axis object if possible?

I think the convention is evolving. Originally you always had to use the Axis object, and a lot of older code reflects that. I now tend to use the Axis if it's "easy" to get, but if it involves a lot of threading I avoid it.

PR #411, opened 2024-01-12
Fixes bug in LoraConfig

blahBlahhhJ:
Removes two unnecessary keys when converting into hf lora config.
* `enable_lora`
* `merge_weights`

I don't think these two keys exists in the huggingface PEFT package, so having these keys will cause errors.

Reference:
* https://github.com/huggingface/peft/blob/v0.7.1/src/peft/config.py#L83
* https://huggingface.co/docs/peft/v0.7.1/en/package_reference/config#peft.PeftConfig
* https://huggingface.co/docs/peft/package_reference/lora#peft.LoraConfig


dlwh:
thanks! those definitely used to be there!

PR #441, opened 2024-01-30
Dependabot bumps

dlwh:
putting these all in one place so i don't have to run several different tests


dlwh:
https://wandb.ai/stanford-mercury/levanter/runs/149gvmgx?workspace=user-dlwh

PR #445, opened 2024-02-02
bring over the improvement to ray preproc from dev

dlwh:


dlwh:
cc @Ivan-Zhou and @blahBlahhhJ merging this into main

Issue #443, opened 2024-02-01
GQA in Levanter | TypeError: body_fun output and input must have same type structure


Ivan-Zhou:
After I rebased my branch to the latest master, in my training with Llama2-3B, I ran into this TypeError between body_fun output and input. After I reverted the change from https://github.com/stanford-crfm/levanter/pull/410, the error did not happen again in my training. 

Is this related to some parameter in my config that needs to be updated? Here is my config: https://github.com/stanford-crfm/levanter/blob/1dec0b5c09eb9dd0a17a23787c111655d70cf9f5/config/llama2_3b_pretrain.yaml#L2-L10

```
Traceback (most recent call last):
  File "/home/ivan/levanter/src/levanter/main/train_lm.py", line 195, in <module>
    levanter.config.main(main)()
  File "/home/ivan/levanter/src/levanter/config.py", line 84, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "/home/ivan/levanter/src/levanter/main/train_lm.py", line 190, in main
    trainer.train(state, train_loader)
  File "/home/ivan/levanter/src/levanter/trainer.py", line 326, in train
    for info in self.training_steps(state, train_loader, run_hooks=run_hooks):
  File "/home/ivan/levanter/src/levanter/trainer.py", line 311, in training_steps
    info = self.train_step(state, example)
  File "/home/ivan/levanter/src/levanter/trainer.py", line 288, in train_step
    loss, new_model, new_optstate = self._train_step_fn(
  File "/home/ivan/venv310/lib/python3.10/site-packages/haliax/partitioning.py", line 259, in __call__
    return self._call(False, *args, **kwargs)
  File "/home/ivan/venv310/lib/python3.10/site-packages/equinox/_module.py", line 935, in __call__
    return self.__func__(self.__self__, *args, **kwargs)
  File "/home/ivan/venv310/lib/python3.10/site-packages/haliax/partitioning.py", line 312, in _call
    output_shape = _cached_filter_eval_shape(self._fn, *args, **kwargs)
  File "/home/ivan/venv310/lib/python3.10/site-packages/haliax/partitioning.py", line 517, in _cached_filter_eval_shape
    _eval_shape_cache[static] = eqx.filter_eval_shape(fun, *args, **kwargs)
  File "/home/ivan/venv310/lib/python3.10/site-packages/equinox/_eval_shape.py", line 38, in filter_eval_shape
[...]

PR #449, opened 2024-02-02
Bugfix: GQA with FA

blahBlahhhJ:
Related issue #443 

1. Added new unit tests for llama & mistral to test backward pass for the LMHead model.
2. Newer version of huggingface transformer has a new required argument for `LlamaDecoderLayer`.

Cause of the error in FA backward:
* K,V are broadcasted over Q's axis "q_heads_per_group" during forward pass but the "un-broadcast" cannot be automatically done by the FA backward kernel.

How I fix this issue:
* Since K,V are repeated over this axis, the gradient should simply be a summation over this axis. So for dK and dV, I sum over all the axis that doesn't belong to K and V (but are present in dK_j and dV_j)

### DIFF
--- a/src/levanter/models/flash_attention.py
+++ b/src/levanter/models/flash_attention.py
@@ -304,13 +304,16 @@ def _flash_attention_backward(
             dAttn_ij = p_ij * (dP_ij - D_i)
             dAttn_ij = dAttn_ij.astype(dQ_i.dtype)
 
-            dV_j = dV_j + hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
-            dK_j = dK_j + hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
+            pdO = hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
+            dAq = hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
 
             # GQA-specific: eliminate unnecessary axes (e.g. 'q_heads_per_group')
-            unnecessary_axes = hax.eliminate_axes(dV_j.axes, v.axes)
-            dV_j = hax.sum(dV_j, unnecessary_axes)
-            dK_j = hax.sum(dK_j, unnecessary_axes)
+            unnecessary_axes = hax.eliminate_axes(pdO.axes, v.axes)
+            pdO = hax.sum(pdO, unnecessary_axes)
+            dAq = hax.sum(dAq, unnecessary_axes)
+
+            dV_j = dV_j + pdO
+            dK_j = dK_j + dAq
 
             dQ_i = dQ_i + hax.dot(KPos.name, dAttn_ij, k_j).astype(dQ.dtype)
             # dQ[i*block_size:(i+1)*block_size] = dQi
--- a/src/levanter/models/flash_attention.py
+++ b/src/levanter/models/flash_attention.py
@@ -304,16 +304,16 @@ def _flash_attention_backward(
             dAttn_ij = p_ij * (dP_ij - D_i)
             dAttn_ij = dAttn_ij.astype(dQ_i.dtype)
 
-            pdO = hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
-            dAq = hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
+            dV_ji = hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
+            dK_ji = hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
 
             # GQA-specific: eliminate unnecessary axes (e.g. 'q_heads_per_group')
-            unnecessary_axes = hax.eliminate_axes(pdO.axes, v.axes)
-            pdO = hax.sum(pdO, unnecessary_axes)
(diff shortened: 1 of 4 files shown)


src/levanter/models/flash_attention.py
```diff
@@ -307,6 +307,11 @@ def do_inner_block(state):
             dV_j = dV_j + hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
             dK_j = dK_j + hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
 
+            # GQA-specific: eliminate unnecessary axes (e.g. 'q_heads_per_group')
```
blahBlahhhJ, line 310: I'm not sure if this is a good way to do this summation. If there's a better place, please let me know. I'm guessing the placement of this summation logic would affect performance. @dlwh

src/levanter/models/flash_attention.py
```diff
@@ -307,6 +307,11 @@ def do_inner_block(state):
...
+            unnecessary_axes = hax.eliminate_axes(dV_j.axes, v.axes)
+            dV_j = hax.sum(dV_j, unnecessary_axes)
+            dK_j = hax.sum(dK_j, unnecessary_axes)
```
blahBlahhhJ, line 313: I'm not sure if this is a good place to do the summation. If there's a better place please let me know. I guess the placement of this logic would affect performance of FA. @dlwh

dlwh: wait, is this the right semantics? I think, if nothing else, you need to do this reduction before you add it to the prior dV_j/d_K_j

tests/test_llama.py
```diff
@@ -224,6 +225,25 @@ def test_llama_lm_head_model(num_kv_heads):
...
+    input_ids = hax.random.randint(random.PRNGKey(0), (Batch, Pos), 0, Vocab.size)
+    mask = hax.nn.attention.causal_mask(Pos, llama_config.KeyPos)
+
+    llama_model = LlamaLMHeadModel.init(Vocab=Vocab, config=llama_config, key=random.PRNGKey(0))
+
+    def f(llama_model, input_ids, mask):
+        out = llama_model(input_ids, mask)
+        return hax.sum(out).scalar()
+
+    _, grads = eqx.filter_value_and_grad(f)(llama_model, input_ids, mask)
```
dlwh, line 244: can we add in test_fa that checks for all_close in the FA-vs-non-FA case for GQA? I suspect this code is incorrect

dlwh (changes requested):
i think this is incorrect? let's add an explicit test in test_fa that checks equivalence of grad in the GQA case

tests/test_llama.py
blahBlahhhJ: sorry I did made a mistake there.
Added a grad test in test_fa and now it passes.

src/levanter/models/flash_attention.py
blahBlahhhJ: You're right. My newest revision should work. I did just randomly pick the variable names for the intermediate dot product result. let me know if you want some other meaningful variable names for them.

src/levanter/models/flash_attention.py
```diff
@@ -304,8 +304,16 @@ def do_inner_block(state):
             dAttn_ij = p_ij * (dP_ij - D_i)
             dAttn_ij = dAttn_ij.astype(dQ_i.dtype)
 
-            dV_j = dV_j + hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
-            dK_j = dK_j + hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
+            pdO = hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
+            dAq = hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
```
dlwh, line 308:
```suggestion
            dV_ji = hax.dot(QPos.name, p_ij, dO_i).astype(dV_j.dtype)
            dK_ji = hax.dot(QPos.name, dAttn_ij, q_i).astype(dK_j.dtype)
```

src/levanter/models/flash_attention.py
```diff
@@ -304,8 +304,16 @@ def do_inner_block(state):
...
+
+            # GQA-specific: eliminate unnecessary axes (e.g. 'q_heads_per_group')
+            unnecessary_axes = hax.eliminate_axes(pdO.axes, v.axes)
+            pdO = hax.sum(pdO, unnecessary_axes)
+            dAq = hax.sum(dAq, unnecessary_axes)
```
dlwh, line 313:
```suggestion
            dV_ji = hax.sum(dV_ji, unnecessary_axes)
            dK_ji = hax.sum(dK_ji, unnecessary_axes)
```

src/levanter/models/flash_attention.py
```diff
@@ -304,8 +304,16 @@ def do_inner_block(state):
...
+
+            dV_j = dV_j + pdO
+            dK_j = dK_j + dAq
```
dlwh, line 316:
```suggestion
            dV_j = dV_j + dV_ji
            dK_j = dK_j + dK_ji
```

dlwh (approved):

dlwh:
i feel like there should be something cleaner for the dot --> sum. In particular I'm pretty sure that combining the dot and the sum together will be faster.

PR #468, opened 2024-02-14
TransformerEngine Fused Attention

vadam5:
Adds FlashAttention implemented by TransformerEngine (they call it fused attention). Performance improvements can be found here: https://docs.google.com/spreadsheets/d/1RHxOxuEE5OeqlVs2fhDvRNm74CzNxDvvTYq9A_4NfGg/edit#gid=1593888450


src/levanter/models/attention.py
```diff
@@ -57,27 +59,77 @@ def dot_product_attention(
...
+            )
+        elif accelerator_type == "gpu":
+            from transformer_engine.jax.fused_attn import AttnBiasType, AttnMaskType, self_fused_attn
+
+            # TransformerEngine looks for qkv shape:
+            # batch_shape, max_seqlen, nqkv, num_heads, head_dim; where nqkv = 3
+            q_ = haliax.rearrange(query, ("batch", "position", "heads", "head_size")).array
+            k_ = haliax.rearrange(key, ("batch", "key_position", "heads", "head_size")).array
+            v_ = haliax.rearrange(value, ("batch", "key_position", "heads", "head_size")).array
+
```
dlwh, line 95: I'd prefer these not be hard coded. At a minimum, could we do something like:
 
```suggestion
            
            q_ = haliax.rearrange(query, ("batch", QPos, "heads", Key)).array
            k_ = haliax.rearrange(key, ("batch", KPos, "heads", Key)).array
            v_ = haliax.rearrange(value, ("batch", KPos, "heads", Key)).array
```

Ideally we wouldn't hardcode the batch and heads dim either, but we'll need a heuristic for that. (If it were just one batch-y dim we could flatten them all)

src/levanter/models/attention.py
```diff
@@ -57,27 +59,77 @@ def dot_product_attention(
...
+            qkv = jnp.stack((q_, k_, v_), axis=2)
```
dlwh, line 96: this mode doesn't work for cross attention or GQA. Could we detect that (the number of heads is different) and do the right thing? at least for GQA? this is what flax does: https://github.com/NVIDIA/TransformerEngine/blob/a174985b91db023181465a3ec9ad161a68787c57/transformer_engine/jax/flax/transformer.py#L742-L751

probably equally easy to check cross and GQA: just check if either size differs

src/levanter/models/attention.py
```diff
@@ -57,27 +59,77 @@ def dot_product_attention(
...
+
+            # scaling_factor = jax.lax.rsqrt(float(query.axis_size(Key)))
+            scaling_factor = 1 / math.sqrt(float(query.axis_size(Key)))
+            is_training = not inference
+
+            # TODO: bias type is probably also configurable
+            attn_bias_type = AttnBiasType.NO_BIAS
+            fused_attn_bias = None
+            if bias:
+                raise Exception(
```
dlwh, line 106: NotImplementedError

src/levanter/models/attention.py
```diff
@@ -57,27 +59,77 @@ def dot_product_attention(
...
+                    "Using bias with flash attention on GPU is not currently implemented. Try setting"
+                    " model.use_bias=False"
+                )
+                # fused_attn_bias = bias.array
+
+            # TODO: We have a mask type we can use to configure this
+            attn_mask_type = AttnMaskType.CAUSAL_MASK
+            batch_size = query.shape["batch"]
+            seq_len = query.shape["position"]
+            mask = jnp.tril(jnp.ones((batch_size, seq_len, seq_len)))
```
dlwh, line 116: either raise if the `AttentionMask.explicit_mask` is not None or use it please

src/levanter/models/attention.py
```diff
@@ -57,27 +59,77 @@ def dot_product_attention(
...
+            attn_bias_type = AttnBiasType.NO_BIAS
+            fused_attn_bias = None
+            if bias:
+                raise Exception(
+                    "Using bias with flash attention on GPU is not currently implemented. Try setting"
+                    " model.use_bias=False"
+                )
+                # fused_attn_bias = bias.array
+
+            # TODO: We have a mask type we can use to configure this
```
dlwh, line 112: let's either handle this todo or raise (preference to handle it)

src/levanter/models/attention.py
```diff
@@ -57,27 +59,77 @@ def dot_product_attention(
...
+                query,
+                key,
+                value,
+                block_size=flash_block_size,
+                mask=mask,
+                bias=bias,
+                dropout=dropout,
+                inference=inference,
+                key=prng,
+                dtype=attention_dtype,
```
dlwh, line 84: leave a TODO about what to do for dtype and precision in fused_attn. (For precision I think it's just warn if not default, but I'm not sure)

src/levanter/models/attention.py
dlwh: probably the thing to do is to choose the latest non-pos/non-key dim as a heads dim and then convert the rest to batch dims, but let's not worry about that right now

dlwh (commented):
my main comment is to detect and raise on any inputs that aren't things we can handle, and to at least handle GQA on top of what we handle now.

Could you also make sure the full test suite passes on a sphinx?

dlwh (approved):
lgtm assuming tests pass on nvidia!

dlwh (approved):
a few small things, but i'm gonna merge now and just do a quick pass so we can be done!

Yay!

PR #484, opened 2024-02-23
Docs for setup on TPUv4s

rtaori:
- allow specifying custom subnetwork
- allow specifying alpha mode
- faq on ray issue when ports aren't open


dlwh (approved):
Thanks!

PR #489, opened 2024-02-27
Update background training job's configuration

Ivan-Zhou:
1. Add RP V1 Dataset as a data mixture for better shuffling;
2. Increase training efficiency with flash attention;
3. Configure data mixture's stop_strategy in the training config.


config/llama2_3b_pretrain.yaml
```diff
@@ -1,27 +1,33 @@
...
   wandb:
     project: "levanter"
     tags: ["redpajama", "llama"]
 
   mp: p=f32,c=bfloat16
   train_batch_size: 1024
-  per_device_parallelism: 4  # set for v3 TPU
-  per_device_eval_parallelism: 8 # set a larger batch size for eval
-  num_train_steps: 250000
+  per_device_parallelism: 8
```
dlwh, line 18: do we still need this? Does -1 work?

dlwh (approved):
my preference would be to set per_device_parallelism to -1 and lower the batch size if we need to

config/llama2_3b_pretrain.yaml
Ivan-Zhou: This would allow us to switch the same training job between v3-64 and v3-128 seamlessly, without need to modify any parameter. I think that's quite useful especially for background run. What do you think?

dlwh: fair enough that makes sense to me

PR #495, opened 2024-02-28
Adds config Prefix to all train commands in GPU doc

Helw150:
Very minor documentation tweak to point to the training configs in their home in the `config` folder so that they run with copy paste!


dlwh:
thanks! they're supposed to work without the `config/` if levanter has been pip installed, though I think that only works without -e. 

I don't really think people use it without -e... Probably better to support the common case.

Helw150:
Ah - I see. Yeah, I used `-e` following the setup in the same doc!

PR #502, opened 2024-02-29
k_cross unused

Helw150:
Not sure what the original `k_cross` use was, but this key now is unused in the current code. I assume it was for another model that had cross attention, but not sure! Apologies if I'm missing a nuance of how this gets implicitly passed somewhere.

(For context @dlwh, I'm currently implementing Whisper in Levanter and these mini-PRs are just different things I notice as I get familiar with the code base. Let me know if you'd prefer I just leave these or leave them to merge as part of a larger PR with a Whisper implementation to come!)

### DIFF
--- a/src/levanter/models/gpt2.py
+++ b/src/levanter/models/gpt2.py
@@ -235,7 +235,7 @@ class Gpt2Block(StateDictSerializationMixin, eqx.Module):
 
     @staticmethod
     def init(config: Gpt2Config, *, key) -> "Gpt2Block":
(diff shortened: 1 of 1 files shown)


dlwh:
Thanks! Yeah just missed it in a refactor. It breaks bitwise repro, but i do that from time to time anyhow.

dlwh:
and let me know if you'd like to chat! Sounds like a good project!

Small refactors are great. I appreciate them

PR #503, opened 2024-03-01
Whisper Implementation in Levanter

Helw150:
Hi @dlwh!

Have all the basic functionality implemented. I've confirmed that the code loads from the HuggingFace config and at least forward passes without error when provided with a real input!

Next steps
1. Implement the HuggingFace conversion scripts
2. Properly test parity in a variety of situations with the HuggingFace implementation

Before I take next steps, I want to make sure that this implementation looks at least reasonable from Haliax and Levanter best practices, since this is my first time working with either! 

Not planning on merging this until I get the full HF loading completed, but since that code all requires the architecture to be static, I'm hoping to get a preliminary review that the Haliax code looks reasonable (most of it is adapted from the GPT-2 implementation, so hopefully it is!)


src/levanter/models/whisper.py
```diff
@@ -0,0 +1,497 @@
...
+        inv_timescales = hax.exp(-log_timescale_increment * hax.arange(ChannelHalfSize))
+
+        position_ids: NamedArray = hax.arange(SourcePos)
+
+        freqs = position_ids * inv_timescales.broadcast_axis(SourcePos)
+        emb = hax.concatenate(Channels, (hax.sin(freqs), hax.cos(freqs)))
+        return emb
+
+    def __call__(self, seq_len: int) -> NamedArray:
+        return jax.lax.stop_gradient(self.pos_emb[self.Pos, :seq_len])
```
dlwh, line 311: this is unfortunately not enough (by itself) to stop the embedding from being modified (namely by weight decay). This being JAX, you're best off just making this a def and using compite_time_eval in the jit block. (I just realized this today) thttps://github.com/stanford-crfm/levanter/pull/503/files

dlwh (commented):
generally looks good. One bug that I noticed (and noticed in our llama today!) but otherwise seems right!

src/levanter/models/whisper.py
Helw150: Perfect, adapted my version to use the compile_time_eval block modeled after the LLama one!

Helw150:
Alright @dlwh! I think this is ready for a real review (esp if you have suggestions for tests other than the simple check I have here for parity with a model loaded from HF)

Fixed the comments you mentioned before, so LMK if there's anything that seems off!

A few points that I had to touch Levanter code outside of Whisper:
- In order to have an optional code block work with checkpointing, I added a check that omits the field from the PyTree if it was None in both the serialized model and the model in memory.
- Since the call signature wasn't consistent with the LmModel class, I separated out the required parts for serialization and created a subclass for models that have the same LmModel call signature.

src/levanter/compat/torch_serialization.py
```diff
@@ -151,6 +151,9 @@ def default_eqx_module_from_state_dict(mod: Mod, state_dict: StateDict, prefix:
         value = getattr(mod, field.name)
         # TODO: might want to add a flag that allows missing keys?
         new = jax_tree_from_state_dict(value, state_dict, apply_prefix(prefix, key))
+        # Do not try to update parameters that are never defined
+        if not value and not new:
```
dlwh, line 155: this should be an explicit check again None yeah?

src/levanter/models/whisper.py
```diff
@@ -0,0 +1,562 @@
...
+
+        state_dict.update(my_dict)
+        return state_dict
+
+
+class WhisperAttention(StateDictSerializationMixin, eqx.Module):
+    config: WhisperConfig = eqx.static_field()
+
+    q_proj: hnn.Linear  # input projection from [embed] -> [q, heads, head_dim]
+    # Unlike other attention blocks in Levanter, splits query to support cross-attn
```
dlwh, line 175: it's really only gpt2 that does it this way still i think fwiw

src/levanter/models/whisper.py
```diff
@@ -0,0 +1,562 @@
...
+        my_state_dict: StateDict = {}
+        super().update_state_dict(my_state_dict, prefix=prefix)
+
+        stacked_dict = unstack_state_dict(my_state_dict, prefix=apply_prefix(prefix, "layers"))
+        state_dict.update(stacked_dict)
+
+        return state_dict
+
+    def _state_dict_key_map(self) -> Dict[str, Optional[str]]:
+        return {"ln_f": "layer_norm"}
```
dlwh, line 366: you can also just call your member variable layer_norm and not need this, entirely your call

dlwh (commented):
couple of small questions/suggestions, otherwise lgtm! thanks for taking this on!

tests/whisper_test.py
```diff
@@ -0,0 +1,103 @@
...
+import haliax as hax
+from haliax import Axis
+
+from levanter.compat.hf_checkpoints import RepoRef
+from levanter.models.whisper import WhisperConfig, WhisperModel
+from levanter.utils.tree_utils import inference_mode
+from test_utils import skip_if_no_torch
+
+
+def test_basic_forward_whisper():
```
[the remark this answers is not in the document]
Helw150: It doesn't! the WhisperProcessor and WhisperConfig both work without PyTorch (since I'm specifying the processor should return NP arrays rather than PT tensors)

src/levanter/models/whisper.py
Helw150: I'm actually somewhat curious if grouping all the linears has perf advantages with Jax. Grouping the linear layers has perf advantages in Pytorch, but does the compiler of jax recognize these can be run in parallel and do so?

Helw150: I guess the cost of these linear layers is relatively minimal v.s. the MLP for most models so it might not matter much.

src/levanter/compat/torch_serialization.py
Helw150: Yes! Done.

src/levanter/models/whisper.py
Helw150: Shifted it here! In general, I like the shorter form names but it does seem silly to have the full map function just for the tiny name difference.

src/levanter/models/whisper.py
Helw150: Removed the comment regardless though!

dlwh: I would be shocked if it mattered on TPU especially, but JAX is pretty good at fusing and overlapping things. I wonder if torch.compile also can make most of the difference go away?

Helw150:
Ok - comments addressed and did one cleanup on an unused config item!

src/levanter/models/whisper.py
```diff
@@ -0,0 +1,572 @@
...
+    def __call__(
+        self,
+        input_ids: NamedArray,
+        audio_embeds: NamedArray,
+        attn_mask: Optional[AttentionMask | NamedArray] = None,
+        *,
+        key=None,
+    ) -> NamedArray:
+        Pos = input_ids.resolve_axis("position")
+        causal_mask = hax.nn.attention.causal_mask(Pos, Pos.alias("key_position"))
```
dlwh, line 499: sorry another suggestion based on changes in @vadam5 's branch: can you use AttentionMask.causal? It'll play better with TransformerEngine's flash attention thing... I'm happy to merge as is, it'll just future proof a bit better

dlwh (approved):

dlwh:
Exciting! Thanks!

dlwh:
@Helw150 oops, can you fix tests in CI? I think it's mostly just missing some imports? (please instlal them in CI or add them as extra or something to pyproject.toml)

Helw150:
Ah - yes. Sorry! Forgot transformers doesn't install the soundfile dependency itself.

PR #504, opened 2024-03-01
don't cache rotary_embeddings so we don't inadvertently weight_decay

dlwh:
we were stopping grad on them, but that doesn't stop weight decay from  doing stuff to them. We should really add a buffer to eqx.field....

Oof.


blahBlahhhJ:
This means the regularization loss has a component of `L2norm(ROPE sin&cos)`? It should only affect the loss magnitude but nothing else right? cuz ROPE doesn't update.

dlwh:
right that's the problem. b/c it's stored it gets weight decayed

blahBlahhhJ (approved):
LGTM :)

PR #522, opened 2024-03-12
Working Code to Train an ASR Model from Librispeech

Helw150:


Helw150:
One thing I do need to figure out is how to support bf16 correctly - right now if I drop the model to bf16 it hits a mismatch with the type of the MelSpectrogram

Helw150:
Hmm, spoke to soon on this (though I think the error is on HF side). The dataloader process seems to time out since librispeech_asr isn't a parquet dataset.

dlwh:
> One thing I do need to figure out is how to support bf16 correctly - right now if I drop the model to bf16 it hits a mismatch with the type of the MelSpectrogram

What's the error?

dlwh:
> Hmm, spoke to soon on this (though I think the error is on HF side). The dataloader process seems to time out since librispeech_asr isn't a parquet dataset.

timing out on our end? if so, that's a bug even if HF datasets is throwing

Helw150:
> > Hmm, spoke to soon on this (though I think the error is on HF side). The dataloader process seems to time out since librispeech_asr isn't a parquet dataset.
> 
> timing out on our end? if so, that's a bug even if HF datasets is throwing

Yeah, "librispeech_asr" breaks in streaming mode for some reason. Error goes away on parquet version: `WillHeld/librispeech_parquet`

dlwh:
ok probably we're not catching an exception properly in an actor constructor somewhere... if you have a repro I can take a look

dlwh:
OK, just to be clear: are we getting a crash or is it just hanging? I'm ok with a crash since it's HF's bug (though maybe we should be retrying if that's transient?). hanging isn't ok

Helw150:
Ok - on the two issues:

1) Loader: We're hanging right now. The bug on HF end kills the ShardCache but gets caught inside of `_monitor_metrics`. Then, the job just starts showing `logs/xfxf7t8y.log.log:2024-03-12T16:12:04 - 0 - ShardCache.librispeech/train.100 - shard_cache.py:1636 - WARNING :: Waiting for chunk 2 for 80 seconds`

2) BF16 Support: The Mel Spectrogram (inputs for Whisper) are float32, so when I drop the model to bf16 I'm seeing the following error from the first convolution: `TypeError: lax.conv_general_dilated requires arguments to have the same dtypes, got float32, bfloat16.`

dlwh:
I see. Ok, well for (2) do you want the conv to be in fp32 or in bf16?

Helw150:
Perhaps I'm misunderstanding the `mp` argument in trainer, but in theory I'd like the implementation to be agnostic (conv can be in bf16, f32, or f16). Basically, I'm just trying to find the clean way to cast the input to be the same type as the rest of the model.

dlwh:
For now, can you do: `spec = spec.astype(conv1.weight.dtype)`? Seems like probably that should be the default behavior in Conv? WDYT?

Helw150:
That fixes the error and code-wise seems clean to me! Just didn't know if there was a more canonical way to do this with JMP. 

One thought: Does this casting incur a significant runtime cost? It seems like casting should be very quick compared to transferring tensors from CPU->GPU/TPU but wanted to confirm.

dlwh:
yeah I'm sure you won't notice it perf wise.

I do think JMP isn't going to ultimately be enough, especially with int8 and fp8. In Haiku, JMP was coupled with a registry system that could have per-module JMP policies.

FWIW, Flax accepts dtypes as the argument to its modules and otherwise uses numpy type promotion, which is probably cleanest. I have a similar idea implemented in a branch in Haliax: "semantic dtypes"

https://github.com/stanford-crfm/haliax/pull/65/files

I'll probably merge it soonish, but I was trying to figure out fp8 and int8 first.

Helw150:
I think the astype is good for now! Kicking off a Whisper-Tiny training run on Librispeech now with all the fixes in. I can work on trying to figure out the streaming bug on that other dataset once I have a simple training job running to make sure the resulting model spits out at least semi-reasonable outputs!

dlwh:
sounds good!

Helw150:
Still must have a bug somewhere™️ (need to add more tests to isolate whether it's in training or exporting to HuggingFace). The model trains successfully and has a reasonable loss curve



However, when I run inference on HuggingFace the model just repeats the same tokens over and over: https://huggingface.co/WillHeld/levanter-whisper-tiny-librispeech. Its learning *something* because it only generates all caps tokens (Librispeech is all caps). I wouldn't expect the tiny model trained on Librispeech to be amazing, but certainly wouldn't expect this repetitive behavior if everything is working correctly.

Helw150:
If there are bugs in export or training, they are subtle. One suspicion I have is that the whisper vocabulary is a bit finicky for Librispeech data size. Trying to resolve this by adding an option to define custom tokenizer (rather than just HF processors) and seeing if using the Character Level Wav2Vec tokenizer works better!

dlwh:
:( I'm around on campus tomorrow if you want to chat. Can also pair remotely. My guess is we need to step through a forward path for both in lock step and see where they differ (not jitted jax)

Helw150:
I'm confident that the model implementation and export to HF are equivalent. Tested the HF Whisper, Levanter Whisper (initialized from pretrained whisper), and HF loaded from Levanter export side by side on a ton of examples and got equivalent results for all.

The issue is somewhere in training itself, but I'm starting to narrow in on it (and I think it's an hparam issue, not an implementation issue). Will update tomorrow to see if a pair session would be useful (though I'm not physically at Stanford right now, I'm back at Georgia Tech finishing the last of my coursework since I started my degree before Diyi's move)

dlwh:
OK sounds good! (Probably friday if not in person then, since I'm on campus rarely)

Helw150:
Working Training run complete - demo of inference on HuggingFace.

https://colab.research.google.com/drive/1vrLjZL_ysaT21B0D5g0716v3X80W-unZ?usp=sharing

dlwh:
nice! that seems pretty good for a tiny model!

dlwh:
do you think this is ready to review/merge?

Helw150:
Yup! Just added a proof of life test, but ready for review.

Helw150:
Oop, actually let me get my tests working first.

Helw150:
Ok - now it's ready for review!

src/levanter/compat/hf_checkpoints.py
```diff
@@ -216,6 +228,9 @@ class HFCheckpointConverter(Generic[LevConfig]):
     tokenizer: PreTrainedTokenizerFast | PreTrainedTokenizer
     "The tokenizer to use. If None, will be inferred from the reference_checkpoint"
 
+    feature_extractor: Optional[FeatureExtractionMixin] = None
```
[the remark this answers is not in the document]
dlwh: we can rule of three this: next time we add something like this (tokenizer, feature_extractor) we figure out the abstraction or split things up

src/levanter/data/audio.py
```diff
@@ -78,20 +77,25 @@ class BatchAudioProcessor(BatchProcessor[Tuple[np.ndarray, int, str]]):
     def __init__(
         self,
         processor: ProcessorMixin,
+        tokenizer: PreTrainedTokenizerBase,
+        enforce_bos=True,
         enforce_eos=True,
         *,
         batch_size=128,
         override_resources=None,
+        max_length=448,
```
dlwh, line 86: that's a strange value!

src/levanter/data/audio.py
```diff
@@ -208,23 +212,29 @@ def fsspec_expand_glob(url):
...
     enforce_eos: bool = True  # whether to append eos even if the tokenizer doesn't
 
-    ignore_token_id: Optional[int] = None
-
     @cached_property
     def the_processor(self) -> PreTrainedTokenizerBase:
         return load_processor(self.processor)
 
+    @cached_property
+    def ignore_token_id(self) -> int:
```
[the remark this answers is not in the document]
dlwh: (but if it's called ignore_token in HF's stuff, fine)

src/levanter/main/train_asr.py
```diff
@@ -0,0 +1,199 @@
...
+    use_hf_model_config: bool = False  # if true, replace the model config with the hf config from the checkpoint
+
+    # TODO: atm we don't support loading from a checkpoint that has a different tokenizer. this is a bit annoying
+    # TODO: atm you have to at least specify a levanter model config with the same type as the hf checkpoint
+
+    hf_save_path: Optional[str] = None
+    hf_upload: Optional[str] = None
+    hf_save_steps: int = 10000
+
+    update_hessian_steps: int = 10
```
dlwh, line 47: copy pasta?

src/levanter/models/whisper.py
```diff
@@ -60,19 +61,30 @@ class WhisperConfig(HFCompatConfig):
     gradient_checkpointing: bool = True
 
     # Attention-related config
-    upcast_attn: bool = False
+    upcast_attn: bool = True
     use_flash_attention: bool = False
     flash_attention_block_size: Optional[int] = None
 
+    # Tokenizer Info
+    # (Annoyingly Required for HF Compat because no separation of concerns)
```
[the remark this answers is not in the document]
Helw150: I could probably work this into hf_compat at export time since we have both the model and the tokenizer there at that stage. That way the Levanter model keeps its separation of concerns, but export still goes smoothly to HF.

The HF Whisper implementation requires this defined in their model config because you can call `model.generate` WITHOUT a tokenizer (so it uses these to force the BOS token and stop generation).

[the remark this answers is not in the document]
Helw150: I'll see if I can make that work cleanly before I merge.

src/levanter/compat/hf_checkpoints.py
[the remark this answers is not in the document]
Helw150: The good news here is that this should also work for VLMs since the FeatureExtractorMixin covers both audio and vision feature extractors.

tests/whisper_test.py
```diff
@@ -112,10 +142,14 @@ def test_hf_roundtrip():
...
-    ds = load_dataset("hf-internal-testing/librispeech_asr_dummy", "clean", split="validation")
+    ds = load_dataset("WillHeld/test_librispeech_parquet", split="validation")
     inputs = processor.feature_extractor(ds[0]["audio"]["array"], return_tensors="pt")
     input_features = inputs.input_features
-    decoder_input_ids = torch.tensor([[1, 1]]) * c.decoder_start_token_id
+    tokenized = processor.tokenizer(
+        ds[0]["text"], max_length=6, padding="max_length", truncation=True, return_tensors="pt"
+    )
+    decoder_input_ids = tokenized["input_ids"]
+    torch.tensor([[1, 1]]) * c.decoder_start_token_id
```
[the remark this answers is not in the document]
Helw150: Yup, stray code from when I was debugging. Will remove.

src/levanter/data/audio.py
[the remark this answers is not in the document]
Helw150: I'll switch to pad_token_id if that's your preference. I used ignore here to keep the parallel with the names in `data/text.py`

src/levanter/data/audio.py
Helw150: It's the og Whisper's default - I also don't quite understand why? Another annoying thing from the og Whisper is that it uses an encoder size that isn't divisible by 2, so flash attention doesn't work 😂

src/levanter/main/train_asr.py
Helw150: Oops, yes. Don't know why mypy didn't yell at me for unused var.

src/levanter/data/audio.py
dlwh: wow

Helw150: Strange values galore

src/levanter/main/train_asr.py
dlwh: it's a memvar so I don't think it cares. it's probably vestigial in train_lm too

tests/whisper_test.py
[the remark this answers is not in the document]
Helw150: The function itself has a skip if no torch though.

src/levanter/models/whisper.py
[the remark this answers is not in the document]
Helw150: Relevant change here: https://github.com/stanford-crfm/levanter/pull/522/files#diff-4ad20b164ff0ec1d287ec49989d2861baf52b206289753541840ba9f448f8fc9R671-R687

src/levanter/compat/hf_checkpoints.py
```diff
@@ -643,6 +668,23 @@ def _save_pretrained_local(
             else:
                 raise
 
+        if self.tokenizer:
+            dict_config = mergedeep.merge(
+                {},
+                dict_config,
+                {
+                    "pad_token_id": self.tokenizer.pad_token_id,
```
dlwh, line 676: not all tokenizers have a pad_token_id (or eos or bos), can you gate these behind whether or not they're there? or maybe -1 will be ok i dunno

Helw150: Gated!

Helw150: Confirmed the None checks won't throw errors, HuggingFace defines all these properties in the Base Tokenizer: https://github.com/huggingface/transformers/blob/v4.38.2/src/transformers/tokenization_utils_base.py#L1165-L1183

dlwh: i meant on the loading side

dlwh: lgtm!

Issue #508, opened 2024-03-05
Implement QLoRA


sr5434:
I am trying to fine-tune Phind 34B using Levanter on a TPU v4-8 VM. How can I implement QLoRA without getting an error telling me I must use full or half precision?

dlwh:
Yeah sorry, we don't have QLoRA. I don't really know the best way to do it in JAX is and I'm unlikely to implement it. Happy to give advice.

I think a v4-8 should be able to fit 34B for unquantized LoRA with mixed fp32/bf16 though?

sr5434:
I think it is trying to run a distributed training, and duplicating it across each core. Is there a way for it to use the TPU-v4 as one big device? Also, could I implement QLoRA by just doing: 

```python
weights.to(jnp.int4)
```

dlwh:
We shard the model parameters so each should only get 1/4 of the model... If it's not fitting it's a bug I'm pretty sure.

Quantization is more complicated than that, unfortunately. There are different strategies and I don't know the best one on TPU/JAX.

sr5434:
I think it might be a bug, as I'm getting an OOM. However, it is not a TPU OOM, but a CPU one for some reason.

sr5434:
I'm using a virtual environment for python 3.10. [This gist](https://gist.github.com/sr5434/b5d75b6b6e527d9551a3d495552aa120) contains my script and config. Also, I can't tell if the error message is about the CPU running out of ram(implied by the tmalloc warnings) or the TPU, implied by the error.
These are my logs:

```
2024-03-05T22:15:58 - 0 - levanter.distributed - distributed.py:252 - INFO :: Successfully started ray head on port 61964.
2024-03-05T22:15:58 - 0 - levanter.distributed - distributed.py:267 - INFO :: ray.init(address='10.130.0.8:61964', namespace='levanter', **{})
2024-03-05 22:15:58,436 INFO worker.py:1540 -- Connecting to existing Ray cluster at address: 10.130.0.8:61964...
2024-03-05 22:15:58,445 INFO worker.py:1715 -- Connected to Ray cluster. View the dashboard at 10.130.0.8:8265 
[...]

PR #523, opened 2024-03-12
Fix HF model loading/lora on low memory accelerators

dlwh:
We now intelligently decide between loading a model on CPU or GPU, and we put out_axis_resources back in a few places that really shouldn't need it, but do


dlwh:
Fixes #508

PR #536, opened 2024-03-21
Single Node and Multi-Node GPU Docs

vadam5:
Adds documentation for both single and multi-node GPU training in levanter.

### DIFF
--- a/docs/Getting-Started-GPU.md
+++ b/docs/Getting-Started-GPU.md
@@ -248,9 +248,47 @@ Then, submit the job with sbatch:
 sbatch my-job.sh
 ```
 
-#### Multinode
+### Multi-Node GPU Training
+For multi-gpu training, you need to additionally have [nvidia-fabricmanager](https://docs.nvidia.com/datacenter/tesla/pdf/fabric-manager-user-guide.pdf) installed on each of your nodes.
 
-Something is wrong and this doesn't work right now on the NLP cluster. In theory this should work:
+```
+sudo apt-get install cuda-drivers-fabricmanager
+sudo systemctl start nvidia-fabricmanager
+```
+
+#### Multi-Node Docker Environment Setup
+If you are using a docker container to train your model, your docker run command should look similar to this
+
+```
+sudo docker run -it --network=host -v ~/src/levanter/cache:/cache -v /home/user/levanter:/levanter --gpus=all --shm-size=16g  ghcr.io/nvidia/jax:levanter
+```
+The main difference between the command here and the one found in the [GPU Docker Development Guide](dev/GPU-Docker-Dev.md) is the `--network=host` argument. This tells the docker container to use the host machine's network instead of the default docker `bridge` network. Using `host` is the easiest way to do multi-node networking with docker and should be sufficient for your training purposes. Please see docker's [host](https://docs.docker.com/network/network-tutorial-host/) and [bridge](https://docs.docker.com/network/network-tutorial-standalone/) network documentation for more information.
+
+#### Multi-Node Training Command
+We use [JAX Distributed](https://jax.readthedocs.io/en/latest/multi_process.html) to help manage multi-node training in Levanter. On each node you can run a command like the following to kick off a training job:
+
+```
+NCCL_DEBUG=INFO python src/levanter/main/train_lm.py \
+  --config_path config/gpt2_7b.yaml \
+  --trainer.ray.auto_start_cluster false \
+  --trainer.per_device_parallelism -1 \
+  --trainer.distributed.num_processes 4 \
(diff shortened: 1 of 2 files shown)


docs/Training-On-Your-Data.md
```diff
@@ -369,9 +369,50 @@ This will spin up a TPU VM instance and install Levanter on it. You can then run
...
+sudo apt-get install cuda-drivers-fabricmanager
+sudo systemctl start nvidia-fabricmanager
+```
+
+If you are using a docker container to train your model, your docker run command should look similar to this
+
+```
+sudo docker run -it --network=host -v ~/src/levanter/cache:/cache -v /home/user/levanter:/levanter --gpus=all --shm-size=16g  ghcr.io/nvidia/jax:levanter
+```
+The main difference between the command here and the one found in the [GPU Docker Development Guide](dev/GPU-Docker-Dev.md) is the `--network=host` argument. This tells the docker container to use the host machine's network instead of the default docker `bridge` network. The `bridge` network is used for communication between docker containers that are on the same host machine, while the `host` network settings is better for container communication across different hosts. Since we are using multiple nodes, we want to use the `host` network setting. Please see the docker's [host](https://docs.docker.com/network/network-tutorial-host/) and [bridge](https://docs.docker.com/network/network-tutorial-standalone/) network documentation for more information.
```
dlwh, line 390: I don't think this is quite right. host is just the easiest way to do multinode, but there's overlay networks and stuff. I'd just say host networks are the easiest and are probably sufficient for this purpose

docs/Training-On-Your-Data.md
```diff
@@ -369,9 +369,50 @@ This will spin up a TPU VM instance and install Levanter on it. You can then run
...
+- `--trainer.distributed.process_id` - The process ID of the current node. If the node is coordinator for the training run (its IP address was the one specified at `--trainer.distributed.coordinator_address`), its process ID needs to be set to zero. All other nodes in the train run should have a unique integer ID between [1, `num_processes` - 1].
+
+When the above command is run on the coordinator node, it will block until all other processes connect to it. All the other nodes will connect to the coordinator node before they can begin training. All other training run arguments have the same meaning as with single node runs. We recommend thinking about increasing your `--trainer.train_batch_size` value when you scale from single node to multi-node training, as this is the global batch size for your training job and you've now increased your compute capacity.
+
+
+### Switching Between GPU and TPU
+In levanter, you can switch between using TPUs and GPUs in the middle of a training run. See our tutorial on [Switching Hardware Mid-Training Run](Hardware-Agnostic-Training.md) to learn more.
```
dlwh, line 415:
```suggestion
In Levanter, you can switch between using TPUs and GPUs in the middle of a training run. See our tutorial on [Switching Hardware Mid-Training Run](Hardware-Agnostic-Training.md) to learn more.
```

docs/Training-On-Your-Data.md
```diff
@@ -369,9 +369,50 @@ This will spin up a TPU VM instance and install Levanter on it. You can then run
 gcloud compute tpus tpu-vm ssh my-tpu   --zone us-east1-d --worker=all --command="WANDB_API_KEY=... levanter/infra/launch.sh python levanter/src/levanter/main/train_lm.py --config_path gs://path/to/config.yaml"
 ```
 
-### GPU
 
-TODO, but you can mostly follow the guide for [TPU](#tpu) above.
+### GPU Training
```
dlwh, line 373: I think this should go in the GPU guide actually. We should also say that if you're SLURM (and using Pyxis), you won't need to do provide the distributed arguments. JAX/Levanter will infer them for you.

dlwh (changes requested):
Couple of thoughts/corrections. Thanks!

docs/Getting-Started-GPU.md
```diff
@@ -265,16 +303,15 @@ Something is wrong and this doesn't work right now on the NLP cluster. In theory
...
-
-srun --nodes=2 python -m levanter.main.train_lm --config config/gpt2_small.yaml --trainer.per_device_parallelism -1
+CONTAINER_PATH="ghcr.io/nvidia/jax:levanter"
+TRAINING_COMMAND="python -m levanter.main.train_lm --config_path config/gpt2_7b.yaml --trainer.ray.auto_start_cluster false --trainer.per_device_parallelism -1"
 
-Submit the job with sbatch:
-
-```bash
-sbatch my-job.sh
+srun docker run --gpus=all --shm-size=16g --rm $CONTAINER_PATH $TRAINING_COMMAND
```
vadam5, line 309: Pretty sure srun + docker run is wrong. Need to test on the NLP cluster.

dlwh: huh i thought i had tested that, but i dunno

dlwh (approved):
lgtm thanks!

PR #558, opened 2024-04-19
Visualization Fixes

Helw150:
Tiny fixes I made to get visuals working appropriately while I was debugging my training runs.

1) Switch to `step.step` since StepInfo's string representation has changed.
2) `np.exp` on the Negative Log Probs rather than the Positive Log Probs to get appropriate probability.


dlwh:
do you just need to merge main?

dlwh:
ah yes

dlwh (approved):

Helw150:
Yeah, I think that's it. Just ran it!

PR #571, opened 2024-05-01
ensure we cast to the right dtype in attention

### LINKED ISSUE #569: Scanning layers seems not to allow the mixing of precisions in mp anymore
I'm training a Mistral model with the default config on a v4-32 with. Before 2516d06be6fec2ff5660f144649a9f5f577b06e9, using `scan_layers: true` allowed for mixing precision in `mp`, such as `p=f32,c=bfloat16`. However, after that commit, I can no longer mix precisions and have to set `p=f32,c=f32` for the model to train. Otherwise, I get this error message:

```python
TypeError: Scanned function carry input and carry output must have equal types (e.g. shapes and dtypes of arrays), but they differ:
  * the input carry component carry[<flat index 0>] has type bfloat16[512,8192,4096] but the corresponding output carry component has type float32[512,8192,4096], so the dtypes do not match
```

dlwh:
I think fixes #569 

@versae: can you give this branch (fix_attn_dtype) a try? I'm traveling and don't have time right now

cc @Helw150 since it tweaks your code.

### DIFF
--- a/src/levanter/models/attention.py
+++ b/src/levanter/models/attention.py
@@ -689,10 +689,6 @@ def _tpu_splash_attention(
     k_ = _reshape_axes_for_bshd_bins(key, k_class, output_order=list("BHSD")).array
     v_ = _reshape_axes_for_bshd_bins(value, v_class, output_order=list("BHSD")).array
 
-    # jax.debug.inspect_array_sharding(q_, callback=lambda sharding: print(f"q_: {sharding}"))
-    # jax.debug.inspect_array_sharding(k_, callback=lambda sharding: print(f"k_: {sharding}"))
-    # jax.debug.inspect_array_sharding(v_, callback=lambda sharding: print(f"v_: {sharding}"))
-
     B, Hq, Sq, D = q_.shape
     Bk, Hk, Sk, Dk = k_.shape
 
@@ -798,7 +794,6 @@ def _tpu_splash_attention(
         precision,
         prng=prng,
     )
-    print(reference_out_shape.dtype, attn_output.dtype)
     attn_output = attn_output.rearrange(reference_out_shape.axes).astype(reference_out_shape.dtype)
 
     attn_output = haliax.shard(attn_output)
--- a/src/levanter/models/attention.py
+++ b/src/levanter/models/attention.py
@@ -678,6 +678,11 @@ def _tpu_splash_attention(
     if bias is not None:
         raise NotImplementedError("Splash attention does not support bias")
 
+    if attention_dtype is not None and attention_dtype != jnp.float32:
+        warnings.warn("Splash attention only supports float32. Switching to float32.")
+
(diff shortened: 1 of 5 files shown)


Helw150 (approved):
Changes to Whisper code look good to me. 

Overall PR also looks good to me, though I haven't been hitting this issue in my training runs so don't have a reproduction case.

dlwh:
It’s a new regression from the TPU splash attention kernel. Still need to
benchmark it at a full scale

versae:
Sure, I'll test it as soon as I can (OOO atm).

Ivan-Zhou:
A few data points with Llama2-7B run on V5lite:
- I hit the exact same dtype error using the commit 2516d06be6fec2ff5660f144649a9f5f577b06e9 (splash attention, April 26): https://wandb.ai/stanford-mercury/markweb/runs/eo41dcbo40o40505;
- I am able to launch a training job successfully using one commit earlier bd2aad66a2c301a52b52a65b455617aa2e452ba6 with batch size 1024 (April 24): https://wandb.ai/stanford-mercury/markweb/runs/eweowbo40o40505;
I tested with this branch, `fix_attn_dtype`, then the dtype error was avoided, but it kept running into OOM error, even if I reduce the batch size to 256:
- OOM at batch size 1024: https://wandb.ai/stanford-mercury/markweb/runs/eo41dcbo40o410240505
- OOM at batch size 512: https://wandb.ai/stanford-mercury/markweb/runs/eo41dcbo40o45120505
- OOM at batch size 256: https://wandb.ai/stanford-mercury/markweb/runs/eo41dcbo40o42560505

I tried adding back the line to not cast to x.dtype when config.upcast_attn = False, but it doesn't help. It still goes OOM: https://wandb.ai/stanford-mercury/markweb/runs/eo41llama40o42560505.

versae:
Training a Mistral 7B (seq len 2048, batch size 2048, parallelism 16), I can confirm `fix_attn_type` seems to fix the dtype error. I haven't encountered any OOM in my limited experiments yet.

dlwh:
@Ivan-Zhou set model.flash_attention_block_size to null/None and see how it goes?

dlwh:
then try 512

Ivan-Zhou:
@dlwh You are right. With reduced FA block size, I am able to resolve this OOM error. 

I am able to train Llama 7B with using 1024 FA block size and up to 2048 batch size and 4k seq length: https://wandb.ai/stanford-mercury/markweb/runs/eo44039204924kd030d20480507

dlwh:
awesome. can you try with it just set to null/unset? this is a new low level kernel built into JAX and the default block size is 512

Ivan-Zhou:
@dlwh I re-launched training jobs with FA block at null (pink) and 512 (blue). They have identical throughput, both less than 1024 (green):
image

PR #575, opened 2024-05-13
Removing unnecessary chunking and merging of long texts

### LINKED ISSUE #515: Remove hack around long sequence lengths for llama tokenizer
As of tokenizers 0.15.2/transformers 4.38.2 (and possibly a bit earlier), llama tokenizer is no longer quadratic.

I put a hack in in f74fc5c589cc4a58dfc4d431a32c01a254ed0d33 (sorry, not a clean commit) and we just need to revert that change.

https://github.com/huggingface/tokenizers/issues/1449

versae:
Fixes #515. Since https://github.com/huggingface/tokenizers/issues/1449 has been fixed for a while, the entire long text workaround seems unnecessary.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
     "haliax>=1.4.dev291",
     "equinox>=0.11.4",
     "jaxtyping>=0.2.20",
+    "tokenizers>=0.15.2",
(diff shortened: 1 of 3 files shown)


dlwh (approved):
Thanks!

PR #583, opened 2024-05-14
fix bos bug fixes #574

### LINKED ISSUE #574: Double BOS token on Mistral
I was checking the bad performance of a fine-tuned model when I came across a strange issue in tokenization. It seems that for some tokenizers, a double BOS token is added. The reason seems the `enforce_bos` set to `True` as default.

https://github.com/stanford-crfm/levanter/blob/2888a35e9697f12d02a71d8f16e7053e10ddf522/src/levanter/data/text.py#L359-L362

Consequently, a regular space is later added that does not tokenize properly:

https://github.com/stanford-crfm/levanter/blob/2888a35e9697f12d02a71d8f16e7053e10ddf522/src/levanter/data/text.py#L375-L376

For example:
```python
# dataset is a tokenized parquet file from the cache as built by levanter
In [65]: dataset["train"][2]["input_ids"][:6]
Out[65]: [1, 1, 28705, 2213, 24222, 28710]

In [66]: tokenizer.encode("Manfredi")
Out[66]: [1, 2213, 24222, 28710]

In [67]: tokenizer.decode(dataset["train"][2]["input_ids"][:6])
Out[67]: '<s><s>  Manfredi'

In [68]: tokenizer.decode(tokenizer.encode("Manfredi"))
Out[68]: '<s> Manfredi'
```

One possible solution is to change the value to `False` by default, or to change the logic slightly:
```python
 if enforce_eos or enforce_bos: 
     input_ids = tokenizer("hi there")["input_ids"] 
     should_append_eos = input_ids[-1] != tokenizer.eos_token_id and enforce_eos 
     should_append_bos = input_ids[0] != tokenizer.bos_token_id and enforce_bos 
```

dlwh:


dlwh:
cc @Helw150

Helw150 (commented):
:+1: - Sorry for the sloppy code from me! Don't know how I missed the flipped equality.

Helw150 (approved):

dlwh:
eh it happens! I should have caught it review/the tests should have caught it

PR #585, opened 2024-05-15
Remove tip about pdsh.

rjpower:
pdsh is convenient but the buffering interferes with progress updates during training.


dlwh:
oops sorry I should have removed that a long time ago!

PR #587, opened 2024-05-16
Add gemma model and test with 2b.

rjpower:
No expert on Gemma and I know it's probably less useful than Llama3, but I saw the issue open and thought I would try adding it to learn more about Levanter. Despite not reading the documentation yet, I found it really hackable!

Some notes:

* Gemma ties the embedding weights and model weights together in the HF files. I don't preserve this explicitly here, so they can diverge during fine-tuning. I'm not sure _how_ to tie weights with Haliax, but if there are some tips please let me know!
* I'm not sure if the logic I use for handling this mapping is "correct" (in the `from_state_dict` code). I found the prefix logic a bit hard to follow there.
* I adjusted the parameter dtype to bfloat16 to get training working, otherwise the model OOMs on v5e. I suspect this is the fault of the giant embedding layer for Gemma. I'm sure we can shard over the embedding layer but I will need to learn more on the 
* For testing, the roundtrip tests past, but if there was an inference tool to test with some prompts I'd feel more confident about the results. I didn't see anything off-hand.

### DIFF
--- a/src/levanter/main/eval_lm.py
+++ b/src/levanter/main/eval_lm.py
@@ -1,5 +1,5 @@
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Optional
 
 import equinox as eqx
@@ -32,9 +32,9 @@ logger = logging.getLogger(__name__)
 class EvalLmConfig:
     checkpoint_path: Optional[str] = None
     hf_checkpoint: Optional[RepoRef] = None
-    trainer: TrainerConfig = TrainerConfig()
-    data: LMDatasetConfig = LMDatasetConfig()
-    model: LmConfig = Gpt2Config()
+    trainer: TrainerConfig = field(default_factory=TrainerConfig)
+    data: LMDatasetConfig = field(default_factory=LMDatasetConfig)
+    model: LmConfig = field(default_factory=Gpt2Config)
 
     compare_torch: bool = False
     eval_on_train: bool = False
--- a/src/levanter/models/gemma.py
+++ b/src/levanter/models/gemma.py
@@ -1,0 +1,369 @@
+import dataclasses
+from dataclasses import dataclass
+from typing import Dict, Optional, Type, Union
+
+import equinox as eqx
+import jax.numpy as jnp
+import jax.random as jrandom
+from jaxtyping import PRNGKeyArray
+
+import haliax as hax
+import haliax.nn as hnn
+from haliax import Axis, AxisSpec, NamedArray
+from haliax.jax_utils import maybe_rng_split, named_call, shaped_rng_split
+from haliax.nn.scan import Stacked
+
+from levanter.compat.hf_checkpoints import HFCheckpointConverter, HFCompatConfig
+from levanter.compat.torch_serialization import (
+    StateDict,
+    StateDictSerializationMixin,
+    apply_prefix,
+    stack_state_dict,
+    unstack_state_dict,
+)
+from levanter.logging import silence_transformer_nag
+from levanter.models.llama import (
+    LlamaAttention,
+    LlamaEmbedding,
+    LlamaMlp,
+)  # Gemma attention and MLP is identical to LLama
+from levanter.models.attention import AttentionMask
+from levanter.models.lm_model import LmConfig, LmHeadModel
+from levanter.types import BlockFoldable
+from levanter.utils.py_utils import cached_classproperty
+
+
+silence_transformer_nag()
+from transformers import GemmaConfig as HfGemmaConfig  # noqa: E402
+from transformers import PretrainedConfig as HfConfig  # noqa: E402
+
+# Gemma is... very similar to Llama, so we use much of the same modeling code.
+#
+# The key differences are:
+# * Activation is changed to approximate gelu
+# * Embedding weights are tied to the LM head
+# * Gemma allows specifying a head dimension independently of the hidden and intermediate dims.
+
+
+@LmConfig.register_subclass("gemma")
+@dataclass(frozen=True)
+class GemmaConfig(HFCompatConfig):
+    """Config for GemmaModel.
+
+    Defaults are set for Gemma-2B.
+
+    Args:
+        seq_len (int, optional): maximum length of the input sequence. Defaults to 8192.
+        hidden_dim (int, optional): dimension of the hidden state. Defaults to 2048.
+        intermediate_dim (int, optional): dimension of the intermediate state. Defaults to 16384.
+        num_layers (int, optional): number of hidden layers in the Transformer encoder. Defaults to 18.
+        num_heads (int, optional): number of attention heads for each attention layer. Defaults to 8.
+        num_kv_heads (int, optional): number of attention heads for keys and values in each attention layer.
+            Setting to 1 means MQA. Setting to num_heads means MHA. Otherwise GQA.
+            Note that num_heads must be divisible by this number. Defaults to 1.
+        activation_function (str, optional): activation function for the hidden layer. Defaults to "gelu".
+        rope_scaling (Dict, ignored): dict containing the scaling configuration for the Rotary Positional Embedding.
+    """
+
+    activation_function: str = "gelu"
+    initializer_range: float = 0.02
+    layer_norm_epsilon: float = 1e-5
+
+    seq_len: int = 8192
+    hidden_dim: int = 2048
+    intermediate_dim: int = 16384
+    vocab_size: int = 256_000
+    num_layers: int = 18
+    num_heads: int = 8
+    head_dim: int = 256
+    num_kv_heads: int = 1
+    attn_dropout = 0.0
+    norm_eps = 1e-6
+
+    rope_base: int = 10_000
+    norm_embeddings: bool = True
+
+    # Attention-related config
+    upcast_attn: bool = False
+    use_flash_attention: bool = True
+    flash_attention_block_size: Optional[int] = None
+
+    gradient_checkpointing: bool = True
+    gradient_checkpointing_block_size: int = 5
+    scan_layers: bool = True
+
+    use_bias: bool = False
+    rope_scaling: Optional[dict] = None
+
+    # Axis
+    Pos = property(lambda self: Axis(name="position", size=self.seq_len))
+    KeyPos = property(lambda self: self.Pos.alias("key_position"))
+    Embed = property(lambda self: Axis(name="embed", size=self.hidden_dim))
+    Heads = property(lambda self: Axis(name="heads", size=self.num_heads))
+    KVHeads = property(lambda self: Axis(name="kv_heads", size=self.num_kv_heads))
+    Layers = property(lambda self: Axis(name="layers", size=self.num_layers))
+    Mlp = property(lambda self: Axis(name="mlp", size=self.intermediate_dim))
+    HeadSize = property(lambda self: Axis(name="head_size", size=self.hidden_dim // self.num_heads))
+
+    def __post_init__(self):
+        assert (
+            self.num_heads % self.num_kv_heads == 0
+        ), f"num_heads={self.num_heads} not divisible by num_kv_heads={self.num_kv_heads}."
+
+    @cached_classproperty
+    def default_hf_checkpoint_converter(cls) -> HFCheckpointConverter["GemmaConfig"]:  # type: ignore
+        return HFCheckpointConverter(
+            cls,  # type: ignore
+            reference_checkpoint="google/gemma-2b",
+            trust_remote_code=True,
+            HfConfigClass=HfGemmaConfig,
+        )
+
+    # The activation function specified in the Gemma HF repo is "gelu", but this is incorrect, it should
+    # be "gelu_pytorch_tanh". For backwards compatibility, HF did not change the value in the repo, but
+    # instead patches around it. We mimic this behavior and use the approximate gelu internally, and
+    # specify the approximate gelu for HF when appropriate.
+    # See https://github.com/huggingface/transformers/pull/29402 for more detail.
+    @classmethod
+    def from_hf_config(cls, hf_config: HfConfig):
+        if hf_config.hidden_activation:
+            activation_function = hf_config.hidden_activation
+        else:
+            activation_function = hf_config.hidden_act
+
+        if activation_function == "gelu_pytorch_tanh":
+            activation_function = "gelu"
+
+        assert activation_function is not None, "No activation function found in HF configuration."
+        return GemmaConfig(
+            seq_len=hf_config.max_position_embeddings,
+            activation_function=activation_function,
+            hidden_dim=hf_config.hidden_size,
+            intermediate_dim=hf_config.intermediate_size,
+            num_layers=hf_config.num_hidden_layers,
+            num_heads=hf_config.num_attention_heads,
+            num_kv_heads=hf_config.num_key_value_heads,
+            initializer_range=hf_config.initializer_range,
+            layer_norm_epsilon=hf_config.rms_norm_eps,
+            rope_base=hf_config.rope_theta,
+        )
+
+    def to_hf_config(self, vocab_size: int, config_overrides: Optional[Dict] = None) -> HfGemmaConfig:
+        """Convert to HuggingFace's GemmaConfig
+
+        Args:
+            vocab_size (int, optional): Vocabulary size of the tokenizer. Defaults to 32000.
+            config_overrides (dict, optional): Overrides for the config. Defaults to None.
+
+        Returns:
+            HfGemmaConfig: HuggingFace's GemmaConfig
+        """
+        if config_overrides is None:
+            config_overrides = {}
+
+        config = HfGemmaConfig(
+            max_position_embeddings=self.seq_len,
+            hidden_size=self.hidden_dim,
+            intermediate_size=self.intermediate_dim,
+            num_hidden_layers=self.num_layers,
+            num_attention_heads=self.num_heads,
+            num_key_value_heads=self.num_kv_heads,
+            head_dim=self.hidden_dim // self.num_heads,
+            hidden_activation=(
+                "gelu_pytorch_tanh" if self.activation_function == "gelu" else self.activation_function
+            ),
+            initializer_range=self.initializer_range,
+            rms_norm_eps=self.layer_norm_epsilon,
+            vocab_size=vocab_size,
+            **config_overrides,
+        )
+        return config
+
+    @property
+    def model_type(cls) -> Type["GemmaLMHeadModel"]:
+        return GemmaLMHeadModel
+
+
+class GemmaRMSNorm(hnn.LayerNorm):
+    """
+    Like Llama, Gemma uses an RMSNorm instead of a layer norm.
+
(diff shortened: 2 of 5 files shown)


src/levanter/models/gemma.py
```diff
@@ -0,0 +1,608 @@
...
+
+    def _state_dict_key_map(self) -> Dict[str, Optional[str]]:
+        """Map from Levanter model names to HF."""
+        return {"transformer": "model", "embeddings": None}
+
+    def from_state_dict(self, state_dict: StateDict, prefix: Optional[str] = None):
+        # unflatten the linear layers of HF state_dict to match the shape of GemmaMlp
+        d = state_dict.copy()
+
+        # Gemma lm_head weights are tied (shared) with the embedding layer.
```
dlwh, line 553: yeah weight tying is something that isn't handled super well in Equinox (and thus Haliax) I typically just hard code the relationship if they're tied (gpt2 is implemented this way). Equinox's solution is [this](https://docs.kidger.site/equinox/api/nn/shared/) but I haven't really used it in anger yet.

dlwh: (Fine to leave it as is, or you can switch to the gpt2-style tying, or you can try the Equinox thing if you want. I'll leave it to you)

src/levanter/models/gemma.py
```diff
@@ -0,0 +1,608 @@
...
+
+        my_dict.update(
+            flatten_linear_layers(apply_prefix(prefix, "lm_head"), self.lm_head, out_dims_first_in_dict=True)
+        )
+
+        state_dict.update(my_dict)
+        return state_dict
+
+
+def _rotate_half(x: NamedArray) -> NamedArray:
```
dlwh, line 573: what are your thoughts on this kind of copy-paste? HF has an explicit "no code reuse" policy, and there are extremes in the other direction (MaxText is very shared-code happy, and xtransformers even more so). 

My inclination is for things that are literally the same like this to be shared, but if there's even a little bit of mismatch, to not share. (I.e. no adding flags just to make things shared). So I'm kinda inclined to say we should move these functions from llama to their own file and call them in both places. WDYT?

src/levanter/models/gemma.py
```diff
@@ -0,0 +1,608 @@
...
+        )
+        my_dict.update(
+            flatten_linear_layers(apply_prefix(prefix, "down_proj"), self.down_proj, out_dims_first_in_dict=True)
+        )
+
+        state_dict.update(my_dict)
+        return state_dict
+
+
+class GemmaAttention(StateDictSerializationMixin, eqx.Module):
```
dlwh, line 255: this is another one that could just be LlamaAttention? WDYT?

dlwh (approved):
Thanks for taking this on! Overall looks great. I wanted to get your thoughts on the code duplication. So far my stance has been to not do literal copy and paste but copy-and-small-modifications when things need to be modified.

e.g. for mistral we reused LlamaAttention etc.

dlwh:
> 
> No expert on Gemma and I know it's probably less useful than Llama3, but I saw the issue open and thought I would try adding it to learn more about Levanter. Despite not reading the documentation yet, I found it really hackable!

We wanted Gemma, and it's a great first project!

> 
> Some notes:
> 
>     Gemma ties the embedding weights and model weights together in the HF files. I don't preserve this explicitly here, so they can diverge during fine-tuning. I'm not sure how to tie weights with Haliax, but if there are some tips please let me know!

There are a couple of ways to do this (see review comments). But I'm happy to leave them tied or untied.

>     I'm not sure if the logic I use for handling this mapping is "correct" (in the from_state_dict code). I found the prefix logic a bit hard to follow there.

Yeah sorry it sucks. If you have round trip tests passing it should pass. I have a branch in Levanter/Haliax that removes the need for almost all of this. 

>     I adjusted the parameter dtype to bfloat16 to get training working, otherwise the model OOMs on v5e. I suspect this is the fault of the giant embedding layer for Gemma. I'm sure we can shard over the embedding layer but I will need to learn more on the

Embedding layer should already be sharded by default though (b/c of the named axis stuff). v5e-16 is not a ton of tpu ram, but it should fit?

>     For testing, the roundtrip tests past, but if there was an inference tool to test with some prompts I'd feel more confident about the results. I didn't see anything off-hand.


Yeah this is coming up a lot. I've punted on inference just because I wanted to do it "right" if I was gonna do it at all, but it's probably close to time to do something, even if it's not the best. Don't let the good be the enemy of the okay, as a former coworker used to say :-)

src/levanter/models/gemma.py
rjpower: Definitely, I'll switch this over to LLama: I started with the copy-paste without realizing how similar the models were.

src/levanter/main/eval_lm.py
```diff
@@ -32,9 +32,9 @@
 class EvalLmConfig:
     checkpoint_path: Optional[str] = None
     hf_checkpoint: Optional[RepoRef] = None
-    trainer: TrainerConfig = TrainerConfig()
-    data: LMDatasetConfig = LMDatasetConfig()
-    model: LmConfig = Gpt2Config()
+    trainer: TrainerConfig = field(default_factory=TrainerConfig)
+    data: LMDatasetConfig = field(default_factory=LMDatasetConfig)
+    model: LmConfig = field(default_factory=Gpt2Config)
```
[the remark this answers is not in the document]
rjpower: No worries, I just wanted the VSCode integration to work :).

src/levanter/models/gemma.py
```diff
@@ -0,0 +1,608 @@
...
+            trust_remote_code=True,
+            HfConfigClass=HfGemmaConfig,
+        )
+
+    # The activation function specified in the Gemma HF repo is "gelu", but this is incorrect, it should
+    # be "gelu_pytorch_tanh". For backwards compatibility, HF did not change the value in the repo, but
+    # instead patches around it. We mimic this behavior and use the approximate gelu internally.
+    # See https://github.com/huggingface/transformers/pull/29402 for more detail.
+    @classmethod
+    def from_hf_config(cls, hf_config: HfConfig):
```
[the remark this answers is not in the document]
rjpower: Yeah... the HF logic is weird. They were trying to achieve some backwards-compatibility, so now the model itself says "gelu" but in the modeling code it ignores this unless you use the new name...

https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma/modeling_gemma.py#L174

src/levanter/models/gemma.py
rjpower: Thanks for the pointeres! I'll take a look at these tomorrow (out today). I think we should do it, otherwise the output model won't be compatible with the HF Gemma anymore IIUC.

rjpower:
> Thanks for taking this on! Overall looks great. I wanted to get your thoughts on the code duplication. So far my stance has been to not do literal copy and paste but copy-and-small-modifications when things need to be modified.
> 
> e.g. for mistral we reused LlamaAttention etc.

Yeah, I'm happy to adopt the exact LlamaAttention stuff here. I forked Llama thinking there would be more changes. 

I agree, where there's an exact match it seems more sensible just to reuse. I found some frameworks try a bit too hard to reuse, and then you end up with this awkward "programming by configuration" where you have some omni-transformer that's now nigh-impossible to understand or edit.

There's probably some philosophical questions around how models are viewed here. e.g. if models were in a `levanter-user-demo` repo, we'd probably be fine with lots of copy-paste, and just expect them to hew to a common interface. Here your suggestion feels appropriate: let's pull out the truly common patterns but having a bit of overlap in the interest of making it easier to develop new things seems like a good tradeoff.

For now I'll pull from LLama but can refactor later.

src/levanter/models/gemma.py
dlwh: yeah I like that. I'm not as certain about the DecoderLayer and in particular I worry that sometimes it will make HF compatibility harder (breaking the ~isomorphism to the HF structure will complicate export serialization)

src/levanter/models/gemma.py
dlwh: totally makes sense! Yeah at this point there's really just the one transformer architecture with a few random tweaks.

dlwh:
> if models were in a levanter-user-demo repo

I should probably have tried hard to support this from the beginning. I also think the dominant class of researcher these days is more comfortable with copying a big file, tweaking it and running it, rather than my more monolith-y/engineering-y sentiments.

rjpower:
> There are a couple of ways to do this (see review comments). But I'm happy to leave them tied or untied.

Thanks for the tips, I used your GPT-2 approach which worked really nicely.

> > I adjusted the parameter dtype to bfloat16 to get training working, otherwise the model OOMs on v5e. I suspect this is the fault of the giant embedding layer for Gemma. I'm sure we can shard over the embedding layer but I will need to learn more on the
> > ```
> 
> Embedding layer should already be sharded by default though (b/c of the named axis stuff). v5e-16 is not a ton of tpu ram, but it should fit?

You're right, I was just surprised because I thought "2B model, I should be able to pump up the batch size". I think this is actually the activations for the final layer. With seq_len=8k and vocab=256k, we end up with 2B output activations. That seems absurd, but I guess it's right? Then unless we're sharding across the input sequence, even with bfloat16 activations, a batch size of 2 is taking up half our memory 🤦 .

> Yeah this is coming up a lot. I've punted on inference just because I wanted to do it "right" if I was gonna do it at all, but it's probably close to time to do something, even if it's not the best. Don't let the good be the enemy of the okay, as a former coworker used to say :-)

No pressure! For this type of thing I definitely would be happy with a simple script that just loads the model and runs a fixed prompt through it.

rjpower:
> > if models were in a levanter-user-demo repo
> 
> I should probably have tried hard to support this from the beginning. I also think the dominant class of researcher these days is more comfortable with copying a big file, tweaking it and running it, rather than my more monolith-y/engineering-y sentiments.

This is something I've been thinking about a lot recently, so apologies for the brain dump. But your point is really insightful so I want to highlight it: hand-waving, but I think you have 2 groups of users:

* People who want to toy around and maybe fine-tune a model on their dataset
* People who want to develop new models

And I think it's hard to cater to both well, and worth thinking about who your audience is. The former are probably well-served by things like HF which have a lot of integrations and emphasis on end-to-end workflows. But like in my case (yes I'm a systems person and... ignore the fact that my last ML research is 7 years ago), I'm really interested in exploring distillation and layer-wise training and factorized models and... I look at a lot of LLM libraries and they just don't fit my needs well. Trying to use HF in a way that doesn't fit their existing workflows is terrible: no offense to them, it's just not a workflow they prioritize.

I kind of go in planning to hack at the modeling code and that a default supervised training loop won't work for what I need. But of course as you mention there's a tradeoff: the provided training script will probably handle things like checkpointing and restore correctly and have nice integrations etc. And those are good and useful but I've got to feel like I can get something working first, and then maybe "grow up" to use some of the more complete integrations.

FWIW, I'm enjoying working with Levanter so far, and I think you've got a good balance! It's certainly the easiest framework I've found for working with TPUs. Thanks for the quick review and guidance!

src/levanter/models/gemma.py
rjpower: Thanks for the tip, I implemented the gpt2-style tying as you mentioned and it worked perfectly.

rjpower:
I updated the code (sorry for the chatter, I'm not super great with the Github review interface...).

As suggested, I went ahead and used the Llama classes. Because Gemma has the specific casting behavior for  RMSNorm, we end up with a copy of the decoder and transformer classes. As you mention, we could refactor these, but honestly I'm comfortable with it as is for now: if it seems like a super stable interface having a more general transformer block could be useful later on, but I don't think the copying hurts for now.

Let me know if you have any other thoughts!

dlwh:
> > Embedding layer should already be sharded by default though (b/c of the named axis stuff). v5e-16 is not a ton of tpu ram, but it should fit?
> 
> You're right, I was just surprised because I thought "2B model, I should be able to pump up the batch size". I think this is actually the activations for the final layer. With seq_len=8k and vocab=256k, we end up with 2B output activations. That seems absurd, but I guess it's right? Then unless we're sharding across the input sequence, even with bfloat16 activations, a batch size of 2 is taking up half our memory 🤦 .

Ah yeah the vocab size on a small model is killer. 256k is crazy big (though 8k tokens on a model that small is probably overkill?) I think we'd need to do TP over the seqlen to make that work, though probably better to accept a small batch size

> 
> > Yeah this is coming up a lot. I've punted on inference just because I wanted to do it "right" if I was gonna do it at all, but it's probably close to time to do something, even if it's not the best. Don't let the good be the enemy of the okay, as a former coworker used to say :-)
> 
> No pressure! For this type of thing I definitely would be happy with a simple script that just loads the model and runs a fixed prompt through it.

Yeah, I'm not saying I'll do it this week but maybe soon? I'd want at least basic kv caching and then have to figure out the modified model interface...

dlwh:
> > > if models were in a levanter-user-demo repo
> > 
> > 
> > I should probably have tried hard to support this from the beginning. I also think the dominant class of researcher these days is more comfortable with copying a big file, tweaking it and running it, rather than my more monolith-y/engineering-y sentiments.
> 
> This is something I've been thinking about a lot recently, so apologies for the brain dump. But your point is really insightful so I want to highlight it: hand-waving, but I think you have 2 groups of users:
> 
>     * People who want to toy around and maybe fine-tune a model on their dataset
> 
>     * People who want to develop new models
> 

Yeah I agree there are those two but there's also a difference with e.g. the optimization research that's been done with Levanter, and there's also the custom-dataset-at-a-nontrivial-scale-set which is also different and really what Levanter was originally targeted at. I did kind of try to court fine-tuners but they're happy with PyTorch on the GPUs under their desks.

> 
> And I think it's hard to cater to both well, and worth thinking about who your audience is. The former are probably well-served by things like HF which have a lot of integrations and emphasis on end-to-end workflows. But like in my case (yes I'm a systems person and... ignore the fact that my last ML research is 7 years ago), I'm really interested in exploring distillation and layer-wise training and factorized models and... I look at a lot of LLM libraries and they just don't fit my needs well. Trying to use HF in a way that doesn't fit their existing workflows is terrible: no offense to them, it's just not a workflow they prioritize.

Interesting, I always see HF as a kind of jack-of-all-trades/master-of-none kind of code base. Their transformers expose a ton of inner workings, but the implementation quality is often not super great.

> 
> I kind of go in planning to hack at the modeling code and that a default supervised training loop won't work for what I need. But of course as you mention there's a tradeoff: the provided training script will probably handle things like checkpointing and restore correctly and have nice integrations etc. And those are good and useful but I've got to feel like I can get something working first, and then maybe "grow up" to use some of the more complete integrations.
> 

Yeah I'm still trying to figure out the right abstractions for the Trainer (or really, I'm trying to turn the Trainer into legos and disassemble it).

> FWIW, I'm enjoying working with Levanter so far, and I think you've got a good balance! It's certainly the easiest framework I've found for working with TPUs. Thanks for the quick review and guidance!

Thanks! I need to put that quote on the front page!

Please keep me in the loop, and would love to chat if you have interest in our larger group efforts (or even if you don't!)

dlwh (approved):
Thanks! looks great!

dlwh:
i cannot for the life of me figure out how to make it run the tests automatically for this PR. I swear it used to do it for outside PRs

rjpower:
>
> This is something I've been thinking about a lot recently, so apologies
> for the brain dump. But your point is really insightful so I want to
> highlight it: hand-waving, but I think you have 2 groups of users:
>
> * People who want to toy around and maybe fine-tune a model on their dataset
>
> * People who want to develop new models
>
> Yeah I agree there are those two but there's also a difference with e.g.
> the optimization research that's been done with Levanter, and there's also
> the custom-dataset-at-a-nontrivial-scale-set which is also different and
> really what Levanter was originally targeted at. I did kind of try to court
> fine-tuners but they're happy with PyTorch on the GPUs under their desks.
>
Yes, very true. If someone can do their task with a GPU and PT, it's hard
to dissuade them...

It's a good point about doing research with Levanter: just using it for
active research ensures it's useful for active research, if that makes
sense? I've seen a bunch of toolkits now that sort of assume all models
look a certain way and end up not being useful for much as a result. Anyway
this is just my idle chatter so please don't take it as a critique or
anything.

Interesting, I always see HF as a kind of jack-of-all-trades/master-of-none
> kind of code base. Their transformers expose a ton of inner workings, but
> the implementation quality is often not super great.
>
>
Yeah, I just found it tough to break into and use for new experiments.
There's multiple attention implementations, the input generation is mixed
into the model, the KV-cache logic is weaved through everything... there
was just a lot you'd have to sort of understand or tear-out if you want to
try out models that don't fit the mold well.

Yeah I'm still trying to figure out the right abstractions for the Trainer
> (or really, I'm trying to turn the Trainer into legos and disassemble it).
>
No criticism here! I'm sure as we end up trying new things the appropriate
legos will start to appear.

> FWIW, I'm enjoying working with Levanter so far, and I think you've got a
> good balance! It's certainly the easiest framework I've found for working
> with TPUs. Thanks for the quick review and guidance!
>
> Thanks! I need to put that quote on the front page!
>
> Please keep me in the loop, and would love to chat if you have interest in
> our larger group efforts (or even if you don't!)
>
Sure, I'd love to collaborate more! I actually started poking around
because an old friend from the TPU team mentioned working with Levanter on
the off-chance I could be helpful for some of the larger projects. (I
worked with TPUs a lot at Google and led some of the Cloud TPU work in the
past...)  It's been fun so far :). I'll ping you on email and we can follow
up from there!

dlwh:
> Yes, very true. If someone can do their task with a GPU and PT, it's hard to dissuade them... It's a good point about doing research with Levanter: just using it for active research ensures it's useful for active research, if that makes sense? I've seen a bunch of toolkits now that sort of assume all models look a certain way and end up not being useful for much as a result. Anyway this is just my idle chatter so please don't take it as a critique or anything. 

Yep! 

>> Interesting, I always see HF as a kind of jack-of-all-trades/master-of-none kind of code base. Their transformers expose a ton of inner workings, but the implementation quality is often not super great.
> Yeah, I just found it tough to break into and use for new experiments. There's multiple attention implementations, the input generation is mixed into the model, the KV-cache logic is weaved through everything... there was just a lot you'd have to sort of understand or tear-out if you want to try out models that don't fit the mold well.

Hrm. KV-cache is just kind of that way, yeah? You have to implement two very similar code paths if you don't just merge them... It is part of why I haven't implemented inference. At some point I was wondering if you could implement kv-cache as a JAX transform, but decided that was probably even worse.

>> Yeah I'm still trying to figure out the right abstractions for the Trainer  (or really, I'm trying to turn the Trainer into legos and disassemble it).
> No criticism here! I'm sure as we end up trying new things the appropriate legos will start to appear.

Yeah, for sure. it takes time and experience to find the right abstractions, as always..

> Sure, I'd love to collaborate more! I actually started poking around because an old friend from the TPU team mentioned working with Levanter on the off-chance I could be helpful for some of the larger projects. (I worked with TPUs a lot at Google and led some of the Cloud TPU work in the past...) It's been fun so far :). I'll ping you on email and we can follow up from there!

Yes, I had gotten word that you might be poking around, and I'm very glad you've done so! I'll look for your email!

PR #588, opened 2024-05-16
Partially-sharded-data-parallel

blahBlahhhJ:
An explanation:

Here I will ignore the difference between DP and FSDP and assume there're only 2 axis (DP, TP) because I handled all DP, FSDP axis together `DP_FSDP_index = DP_index * FSDP_size + FSDP_index`

Suppose we have this mesh of DP=n, TP=4: (d -> device. p-> process)
```
         TP0    TP1     TP2      TP3
DP0    (d0 p0) (d1 p0) (d4 p1) (d5 p1)
DP1    (d8 p2) (d9 p2) (d12 p3) (d13 p3)
  ... a bunch of other DP groups in between so that devices are non-contiguous
DPk     (d2 p0) (d3 p0) (d6 p1) (d7 p1)
```

The 4 devices (e.g. d0d1d4d5) in every rows gets the same data and performs TP (shards the model). Each column receives different data.

We first take care of each process individually and ensure 
* devices in the same DP group (with different TP index) receives the same data
* devices in different DP groups receive different data

This is handled by `local_device_mapping()`. 

From each device, we can extract its DP index by its position in the mesh and map it to a uid. In this example, we will have a mapping of
```
p0: {0: 0, k: 1}
p1: {0: 0, k: 1}
p2: {1: 0, ...}
p3: {1: 0, ...}
```
When we call `make_array_from_callback()` with the mesh, each device gets a slice of size `per_device_batch_size` and thus the slice's start will be `global_begin = DP_index * per_device_batch_size`. Thus we can extract `DP_index` from the slice
```
DP_index = global_begin // per_device_batch_size
```

See `batch_callback()` in `loader.py`, devices will get
```
local_batch[uid*per_device_batch_size : (uid+1)*per_device_batch_size]
```
For `p0`, this means devices with DP index 0 (d0&d1) will get the first half of `local_batch`, and devices with DP index k (d2&d3) will get the second half. This satisfied the two bullet points above.

Next, we take care of the process-level stuff and ensure
* different processes' devices that are in the same DP group (different TP index) receive the same data.
* different processes' devices that are in different DP groups receive different data.

This is handled by `process_mesh_mapping()`.

Each process:
1. maps to its upper left device in the mesh
2. maps to the location of the device
3. maps to the location of the device with TP0 in the same DP group
4. maps to the device in that TP0 location
5. maps to the process of that device
6. maps to a uid
```
p0 -> d0 -> (DP0 TP0) -> (DP0 TP0) -> d0 -> p0 -> 0
p1 -> d4 -> (DP0 TP2) -> (DP0 TP0) -> d0 -> p0 -> 0
p2 -> d8 -> (DP1 TP0) -> (DP1 TP0) -> d8 -> p2 -> 1
p3 -> d12-> (DP1 TP2) -> (DP1 TP0) -> d8 -> p2 -> 1
```
Thus we get this mapping
```
{p0: 0, p1: 0, p2: 1, p3: 1}
```
The uid becomes the `shard_idx` of the dataloader. Thus, p0 and p1 as processes will get the same shard_idx because and will get the same `local_batch`. 

When looking at device-level, we already ensure that d0&d1 gets the first half, d2&d3 gets the second half. This is also true for p1: d4&d5 gets the first half and d6&d7 gets the second half. Now because p0 and p1 have the same shard_idx, we further ensure that d0&d1&d4&d5 gets the same data, d2&d3&d6&d7 gets the same data.

Therefore, these 2 mapping function works for any DP/TP configuration :)


docs/Configuration-Guide.md
```diff
@@ -100,7 +100,7 @@ The following table lists some of the parameters that you might want to change.
 | `seed`                         | The random seed                                                     | 0                                                         |
 | `num_train_steps`              | The number of training steps to run                                 | 400,000                                                   |
 | `train_batch_size`             | The batch size                                                      | 32                                                        |
-| `per_device_train_parallelism` | Number of examples to process on each device during training        | `train_batch_size / (num_accelerators * model_axis_size)` |
+| `per_device_train_parallelism` | Number of examples to process on each device during training        | `train_batch_size / (num_accelerators * model_ici_axis_size)` |
```
dlwh, line 103: should this be model_ics_axis_size * model_dcn_axis_size?

src/levanter/trainer.py
```diff
@@ -550,11 +550,17 @@ class TrainerConfig:
     tensor_parallel_axes: Optional[List[str]] = None  # Axes, if any, to use for tensor parallelism
 
     # TODO: in theory we can support tuples of physical axis names, but I don't think anyone actually uses that.
```
dlwh, line 552:
```suggestion
```

src/levanter/trainer.py
```diff
@@ -550,11 +550,17 @@ class TrainerConfig:
...
-    axis_resources: Mapping[str, str] = field(default_factory=dict)
+    axis_resources: Mapping[str, Any] = field(default_factory=dict)
```
dlwh, line 553:
```suggestion
    axis_resources: Mapping[str, Union[List[str], str]] = field(default_factory=dict)
```

src/levanter/trainer.py
```diff
@@ -550,11 +550,17 @@ class TrainerConfig:
...
     """mapping from logical axis to physical axis. batch_axis, fsdp_axis, and tensor_parallel_axes are preferred"""
-    parameter_axis_resources: Mapping[str, str] = field(default_factory=dict)  # overrides axis_mapping for parameter
+    parameter_axis_resources: Mapping[str, Any] = field(default_factory=dict)  # overrides axis_mapping for parameter
```
dlwh, line 555:
```suggestion
    parameter_axis_resources: Mapping[str, Union[List[str], str]] = field(default_factory=dict)  # overrides axis_mapping for parameter
```

src/levanter/trainer.py
```diff
@@ -550,11 +550,17 @@ class TrainerConfig:
...
     """logical->physical mapping for parameter/optimizer sharding. fsdp_axis and tensor_parallel_axes are preferred"""
-    model_axis_size: int = 1  # how many devices to shard each model over. Data axis is the other axis
+
+    """Interchip Interconnect (ICI) & Data Center Networking (DCN) shardings"""
```
dlwh, line 558: put a link to the page on multislice

src/levanter/trainer.py
```diff
@@ -636,19 +642,58 @@ def initialize(self):
...
+        return Mesh(devices, (ResourceAxis.REPLICA, ResourceAxis.DATA, ResourceAxis.MODEL))
 
     @property
     def eval_batch_size(self):
         return self.per_device_eval_parallelism * self.data_axis_size
 
+    @property
+    def num_slices(self):
+        """number of nodes"""
+        return getattr(jax.devices()[-1], "slice_index", 0) + 1
```
dlwh, line 667: nit: I don't think it's wise to assume that these are sorted. could we max and use cached_property

src/levanter/trainer.py
```diff
@@ -636,19 +642,58 @@ def initialize(self):
...
+
+    @property
+    def num_local_devices(self):
```
dlwh, line 670: local in JAX means "this process" what about `num_devices_in_slice`

src/levanter/data/loader.py
```diff
@@ -172,8 +173,11 @@ def __init__(
         self.mesh = mesh
         self.Batch = Batch
 
-        process_data_pos = override_process_data_pos or levanter.mesh.process_mesh_position(mesh)[0]
-        num_data_process_groups = override_process_data_groups or levanter.mesh.process_mesh_size(mesh)[0]
+        process_mesh_map = process_mesh_mapping(self.mesh)
+        local_devices_map = local_devices_mapping(self.mesh)
+        logger.info(process_mesh_map, local_devices_map)
```
dlwh, line 178: change to debug

src/levanter/data/loader.py
```diff
@@ -182,20 +186,33 @@ def __init__(
...
         self.item_dataset = local_dataset.shard(process_data_pos, num_data_process_groups)
         super().__init__(max_capacity, axis_resources)
 
     def _produce_batches(self) -> Iterator[PyTree]:
         one_item_generator = non_caching_cycle(self.item_dataset)
         batched = _batched(one_item_generator, self.local_batch_size)
 
+        def batch_callback(global_begin, _):
+            # global_begin is uid for DP/FSDP
+            # DP_id * per_device_bs = global_begin
```
dlwh, line 202: could you add a note explaining why this is correct (also so I can be sure I understand)

src/levanter/mesh.py
```diff
@@ -8,29 +8,51 @@
...
+    my_device_pos = np.vectorize(lambda dev: dev.process_index == process_index)(mesh.devices)
     return my_device_pos.nonzero()
 
 
-def process_mesh_position(mesh, process_index: Optional[int] = None) -> tuple[int, ...]:
+def local_devices_mapping(mesh: Mesh, process_index: Optional[int] = None) -> dict[int, int]:
     """
-    If we envision each process as a subgrid of the mesh for its devices, this is the position of the process
-    in the coarsened process-level mesh
+    Handles the case when different devices in same process share the same data in TP.
```
dlwh, line 20: does it handle different devices in different processes?

src/levanter/mesh.py
```diff
@@ -8,29 +8,51 @@
...
-    local_mesh_size = mesh.local_mesh.devices.shape
-    pos = upper_left_position // local_mesh_size
-    return pos
+    local_device_pos = local_device_grid_positions(mesh, process_index)[:2]  # first 2 axes are DP axes.
+    result = {}
+    uid = 0
+    for local_device_index in range(len(local_device_pos[0])):
+        key = local_device_pos[0][local_device_index] * mesh.devices.shape[1] + local_device_pos[1][local_device_index]
+        if key not in result:
+            # when two devices maps to the same key (different TP index), they will get the same data
```
dlwh, line 29: wait, if the TP index is different they get the same data?

src/levanter/mesh.py
```diff
@@ -8,29 +8,51 @@
...
+    return result
 
 
-def process_mesh_size(mesh: Mesh) -> tuple[int, ...]:
+def process_mesh_mapping(mesh) -> dict[int, int]:
     """
-    If we envision each process as a subgrid of the mesh for its devices, then there is a process grid that
-    is a coarsened version of the mesh. This is the size of the process grid.
+    Handles the case when different processes share the same data in TP.
+    If we envision each process as a subgrid of the mesh for its devices, this is the position of the process
```
dlwh, line 38: i don't actually understand how process_mesh is still a valid abstraction/idea in the world with "non-contiguous" devices meshes

dlwh (commented):
I'm confused. Can you explain (as a comment) why the data uid stuff is correct?

dlwh:
also can you merge main so that the TPU tests run

src/levanter/data/loader.py
```diff
@@ -16,9 +16,10 @@
 from haliax.partitioning import ResourceMapping
 from haliax.util import is_named_array
 
-import levanter.mesh
+# import levanter.mesh
```
dlwh, line 19: rm

src/levanter/data/loader.py
blahBlahhhJ: do you mean `process_mesh_map[jax.process_index()] = override_process_data_pos`?
I'm not sure how to use `override_process_data_groups` for these mappings.

src/levanter/mesh.py
blahBlahhhJ: what do you mean?

src/levanter/mesh.py
blahBlahhhJ: Yes. If they get the same DP/FSDP id, they should be getting the same data.

src/levanter/data/loader.py
blahBlahhhJ: updated the PR description.

src/levanter/mesh.py
blahBlahhhJ: updated the PR description.

src/levanter/mesh.py
dlwh: i misunderstood what this comment means

src/levanter/mesh.py
dlwh: oh i see

dlwh (approved):
ok i think i'm convinced!

PR #592, opened 2024-05-20
Adjust test precision and miscellaneous cleanups.

rjpower:
While testing a factorized model I noticed the threshold for the roundtrip tests was too large: with random initialization and 1000 outputs, _all_ outputs of the final softmax will tend to be within 1e-2 of each other!  I tightened the tolerance to 1e-4 for Llama and Mistral accordingly.

Gemma requires more wiggle room, which I'm not happy about: I assume there's a precision difference between our implementation and PyTorch, but if it's this close after several layers at least the model architecture is right.

I also cleaned up some warning noise: adjusted all of the `hax.dot` calls to use named axis arguments, fixed a few places where we generated 1 random split, and updated calls to treemap.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,6 +133,8 @@ dmypy.json
 # JetBrains
 .idea/
 
+# vscode
+.vscode
 
 # Wandb stuff
 /wandb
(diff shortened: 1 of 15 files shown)


dlwh (approved):
thanks!

PR #593, opened 2024-05-20
While waiting for the sharding of long documents, the state becomes invalid

versae:
This PR fixes a rare race condition when the cache for very long documents is still being built but the workers are asking for the shards. The state becomes invalid so capturing the specific exception makes the output cleaner. I also added a `sleep()` so the message is not written every other millisecond.

### DIFF
--- a/src/levanter/data/shard_cache.py
+++ b/src/levanter/data/shard_cache.py
@@ -1640,6 +1640,13 @@ class ShardCache(Iterable[pa.RecordBatch]):
                     current_timeout *= 2
                     current_timeout = min(current_timeout, 100)
                     continue
+                except asyncio.exceptions.InvalidStateError:
+                    self.logger.warning(f"Invalid state waiting for chunk {mapped_index} for {int(next_time - time_in)} seconds")
+                    next_time = time.time()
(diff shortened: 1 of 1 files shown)


src/levanter/data/shard_cache.py
```diff
@@ -1640,6 +1640,13 @@ def _get_chunk_unmapped(self, mapped_index: int, *, timeout: Optional[float] = N
                     current_timeout *= 2
                     current_timeout = min(current_timeout, 100)
                     continue
+                except asyncio.exceptions.InvalidStateError:
+                    self.logger.warning(f"Invalid state waiting for chunk {mapped_index} for {int(next_time - time_in)} seconds")
```
dlwh, line 1644: weird. can you please log the exception (and send me a log?)

dlwh (commented):
would you mind logging the exception stack trace please (and send me a log when it shows up)

otherwise, lgtm!

src/levanter/data/shard_cache.py
versae: Sure, but it'll take me some time as I now have all my TPU quota in use.

PR #596, opened 2024-05-20
Try launching unit tests on TPUs from CI

dlwh:


dlwh:
cc @rjpower since I fixed up a few tests on TPU (mostly just scaled down some initializers?)

dlwh:
one of the fixed tests is actually my "pure jax" flash attention vs vanilla dot product, so it's the same algorithm on both CPU and TPU, but the matmuls are different enough on TPU that it matters, even with HIGHEST.

I mostly agree with your hypothesis, though I framed it in terms of "floating point numbers are more precise near 0"

Floating point is so annoying

PR #599, opened 2024-05-22
Seek for Audio

Helw150:
Implementing seek in the much simpler case of audio. Need to do more testing to make sure this behaves appropriately, but putting the PR up now under WIP status since this has the main idea implemented.

Uses the chunk metadata to create a dummy iterator of the same shape as the true iterator and keeps track of the shard id and the row index reached during the resume process. Then, initializes the actual iterator from the location reached during the dummy iteration.

### DIFF
--- a/src/levanter/data/audio.py
+++ b/src/levanter/data/audio.py
@@ -264,7 +264,7 @@ class ProcessedAudioCache(ShardableDataset[AudioTextStorageBatch]):
 
     def __iter__(self):
         shard_index, row_index = self.chunk_cache.get_resume_information()
-        logger.info(f"Starting Real Iterator at Shard {shard_index}, Row {row_index}.")
+        logger.debug(f"Starting Real Iterator at Shard {shard_index}, Row {row_index}.")
         for batch in self._chunks():
             unarrow = dict_from_record_batch(batch)
             # Flatten Singleton Batch Dimension
--- a/src/levanter/data/audio.py
+++ b/src/levanter/data/audio.py
@@ -263,6 +263,8 @@ class ProcessedAudioCache(ShardableDataset[AudioTextStorageBatch]):
         self.chunk_cache = chunk_cache.with_batch_size(1)
 
     def __iter__(self):
+        shard_index, row_index = self.chunk_cache.get_cache_location()
+        logger.info(f"Starting Real Iterator at Shard {shard_index}, Row {row_index}.")
         for batch in self._chunks():
             unarrow = dict_from_record_batch(batch)
             # Flatten Singleton Batch Dimension
@@ -270,6 +272,9 @@ class ProcessedAudioCache(ShardableDataset[AudioTextStorageBatch]):
             singleton_dict["input_features"] = singleton_dict["input_features"].reshape(singleton_dict["audio_shape"])
             del singleton_dict["audio_shape"]
             yield singleton_dict
+
+    def seek(self):
+        yield from self.chunk_cache.scan_batches_from_chunks()
 
     def _chunks(self):
         return self.chunk_cache.iter_batches_from_chunks()
@@ -473,3 +478,6 @@ class AudioTextDataset(ShardableDataset[AudioTextExample]):
             for example in self.dataset:
                 converted_example = _convert_example(example)
                 yield converted_example
+
+    def seek(self):
+        yield from self.dataset.seek()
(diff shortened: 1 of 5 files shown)


src/levanter/data/audio.py
```diff
@@ -271,6 +273,9 @@ def __iter__(self):
             del singleton_dict["audio_shape"]
             yield singleton_dict
 
+    def seek(self):
```
dlwh, line 276: wdyt about seek just taking an int that tells you where to seek?

src/levanter/data/shard_cache.py
```diff
@@ -1744,10 +1767,30 @@ def with_batch_size(self, batch_size):
             self.cache_dir, batch_size, self._ledger, self._broker, self._reader_offset, self._num_readers
         )
 
+    def _scan_chunk(self, chunk):
+        num_batches = (chunk.num_rows + self._batch_size - 1) // self._batch_size
```
dlwh, line 1771: so the chunk metadata knows exactly how many examples/rows are in each chunk

Also, I'm not sure that the batches from parquet are always the same size (e.g. for remainder at end). I'd be fine with remove the batch_size argument to this class

dlwh (commented):
We can use metadata to make this go faster, especially if we pass in the index we're seeking to. wdyt?

src/levanter/data/shard_cache.py
Helw150: I'm using the ChunkMetadata right now I believe! This skips the actual reading of the parquet file. Or are you saying there is a higher level metadata file I can read from to make this even faster?

src/levanter/data/audio.py
Helw150: That's where I started, but the hard part there is what that int should represent and how that changes at different levels of the code. 

At the trainer level, we want to seek to the batch corresponding to a training step, but at lower levels we need to seek to different levels of granularity (e.g. some datasets need to seek to the right start token, some need to seek to the right document, and ultimately all of them need to seek to the right chunk in storage).

I think that if we use `seek(index: int)` as the arg format, it might get a bit confusing what that integer represents at different parts of the code base. I think that also risks `seek` logic diverging from the actual data loading logic which could lead to subtle resume bugs.

Since the actual data loading logic uses generators, using generators in the seek code made it feels easier to write seek in such a way that I could largely keep it joined to the existing data loading logic while making the perf improvements really close to the expensive file reads.

Helw150: It could be that I just couldn't see the path to write a clean version of the `seek(index: int)`, but it felt like I'd end up having to do arithmetic at each level to convert from the index that made sense at one level to the index that was needed at the lower level, rather than just piggy pack on the data loading logic directly.

src/levanter/data/shard_cache.py
Helw150: Empirically, this seek implementation drops my resume time for my big audio training jobs from hours to under a minute since the main bottleneck happens in the `reader = _ChunkReader.from_metadata(self.cache_dir, chunk, self._batch_size)` step.

Helw150: One thing that may be the source of confusion here -  the variable `chunk` is actually just the `ChunkMetadata` - I've updated the type signature to make that clearer.

I think it might be better to change the variable name entirely, but opted to be consistent with the original `_read_chunk` in which `chunk` also refers to a `ChunkMetadata` variable.

https://github.com/stanford-crfm/levanter/blob/main/src/levanter/data/shard_cache.py#L1754-L1757

Helw150: re: batches from parquet not always being the same size. I actually was a bit puzzled at this because you are right, it's not always divisible. 

Right now in the true data loading logic though, it looks as if the remainder at the end gets returns as it's own (smaller) batch, so I'm mirroring this in the seek logic:  https://github.com/stanford-crfm/levanter/blob/main/src/levanter/data/shard_cache.py#L1815-L1817

src/levanter/data/audio.py
dlwh: i think you do have to do that arithmetic, but maybe it's not worth. i have an idea that is going to be better overall than everything we have now. I need to whiteboard it tomorrow

dlwh:
this is obviated by the big cache changes i made

PR #600, opened 2024-05-24
Check in Dolma data config and OLMo 1B model config

Ivan-Zhou:
1. Add Dolma dataset config. Note that to use OLMo's tokenizer, it requires `pip install ai2-olmo`. 
2. Add Llama model with OLMo 1B config


blahBlahhhJ (approved):
Thanks. will start using this config after debugging psdp

Issue #328, opened 2023-09-27
Adopt Queued Resource on TPU


dlwh:
Going forward, TPUs are supposed to be controlled with [Queued Resource](https://cloud.google.com/tpu/docs/queued-resources) which use a different API for creating TPUs.

So far, the new API is basically worse for us in most ways (but so close to being useful!), but we need to adopt it. It looks like it works with V3 TPUs now (and not just v4s), so at least we can use just it.

While we're doing this, we should probably migrate to using setup scripts

With the new API, the flow for babysitting should look like this:

```
state = NO_MACHINE

while state != DONE:
   state = infer_state()
    match state:
        case NO_MACHINE:
            request_tpu_with_setup_and_metadata()

        case REQUESTED:
           # wait

        case ACTIVE:
           launch_command()  # currently this blocks, but we should probably move to a world where we check on a tmux or something

        case IN_USE:
           # wait

        case PREEMPTED:
            delete_tpu_and_resource()  # Since Queued Resource will not re-enqueue preempted TPUs, delete them.

        case ERROR:
            delete_tpu()  # Delete the TPU in case of an error during job execution.
            exit_with_error()  # Exit the script with an error message or code.

        case DONE:
            delete_tpu_and_resource()  # Make sure to delete the TPU resource.
```

PR #601, opened 2024-05-27
Setup Docker for TPU execution and update infra scripts.

rjpower:
(Still needs some cleanup & doc updates but sending for thoughts.

A few notes:

* We could use draccus configs for these infra parameters, but need to adjust it to support remainder args. I think it's okay to do that as a separate cleanup pass.
* I moved a few scripts into Python: trying to figure out how to preserve commands through 3 layers of bash is challenging... I think it's worth moving more of these over and cleaning it up a bit over time.
* It's still possible to use the `git clone` with this via a volume mount, but I think this should work as-is for most use cases.
* With the current setup each project needs to host the base Levanter image themselves.  We can split the Docker build so that we have a base image and just add the current source on top of it. If this looks reasonable I'll do that.

```
FROM stanford/levanter # contains all of the setup and package installation
...
```

)

 
I tried to optimize the Docker image size a bit using a staged build, as Ray currently requires a source build of Meson, which requires a Clang installation... even with this jax & libtpu are each themselves >250MB installs, so there's no avoiding a large image size at the moment.

Still, with this configuration, a v5-32 (the most I could get given GCPs stingy IP address allocation) takes about 50 seconds to run setup-vm.sh and pull the initial image.  After the initial pull, new deployments take a few seconds to package up the current source directory.

### DIFF
--- a/docs/Getting-Started-TPU-VM.md
+++ b/docs/Getting-Started-TPU-VM.md
@@ -164,6 +164,10 @@ Afterward, you can use the config directly from the TPU VM instance, e.g.:
 With this configuration (unless `trainer.load_checkpoint` is false), Levanter will automatically
 try to load the latest checkpoint if it exists.
 
+Tokenizers and configuration files are loaded via `fsspec` which supports remote
+filesystems , so you can also copy your tokenizer or config file to GCS and use
+a `gs://` path to access it.
+
 ## Common Issues
 ### (CRFM) Permission denied on `/files`
 
--- a/docs/Getting-Started-TPU-VM.md
+++ b/docs/Getting-Started-TPU-VM.md
@@ -95,24 +95,31 @@ First create a configuration file for future launches in your Levanter directory
 ```
 cat > .config <<EOF
 env:
-    WANDB_API_KEY:  ...
-    WANDB_ENTITY: ...
-    WANDB_PROJECT: levanter
-    HF_TOKEN: ...
+    WANDB_API_KEY: 
+    WANDB_ENTITY: 
+    WANDB_PROJECT: 
+    HF_TOKEN: 
+    TPU_STDERR_LOG_LEVEL: 0
+    TPU_MIN_LOG_LEVEL: 0
+    LIBTPU_INIT_ARGS: <extra args to libtpu>
 
 docker_repository: levanter
 zone: us-west4-a
-tpu: test-tpu
+tpu_name: test-spin-up-32
+tpu_type: "v5litepod-16"
+vm_image: "tpu-ubuntu2204-base"
+preemptible: true
+autodelete: false
+subnetwork: "default"
+
 EOF
 ```
 
-Everything after the `--` is run on each worker.
+Now run `launch.py`. This will package your current directory into a Docker image and run it on your workers. Everything after the `--` is run on each worker.
 
 ```bash
 python infra/launch.py -- python levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml --trainer.checkpointer.base_path gs://<somewhere>'
 ```
-
-`launch.py` will package your directory and create and deploy a Docker image  on each worker.
 
 ### Launch a GPT-2 Small in interactive mode
 
@@ -124,18 +131,11 @@ python infra/launch.py -- python levanter/src/levanter/main/train_lm.py --config
 ### Babysitting Script
 
 If you are using a preemptible TPU VM, you probably want to use the "babysitting" script that automatically re-creates
-the VM. This is because preemptible instances can be preempted and will always be killed every 24 hours. The babysitting
-script handles both the creation of the node and the running of a job, and also relaunches the TPU VM if it gets preempted.
-It keeps running the command (and relaunching) until the command exits successfully.
-
-Note that the babysitting-script will automatically set the `RUN_ID` environment variable if not set, and pass it to the
-training command. This ensures that restarted jobs have the same run id, which is important for resumes to work.
-
-You can run it like this:
+the VM. This is because preemptible instances can be preempted and will always be killed every 24 hours. You can run `launch.py` with the `--retries` and `--foreground` parameter to accomplish this. If `--retries` is greater than 1, `launch.py` will automatically attempt to re-create the VM and re-run the command if it fails. (`--foreground` is necessary to keep the script from returning immediately.)
 
 ```bash
-infra/babysit-tpu-vm <name> -z <zone> -t <type> [--preemptible]  -- \
-    python infra/launch.py -- levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml
+    python infra/launch.py --retries=100 --foreground --tpu_name=my_tpu -- python src/levanter/main/train_lm.py --config_path config/my_config.yaml \
+    --trainer.checkpointer.base_path gs://path/to/checkpoints/
 ```
 
 That `--` is important! It separates the spin up args from the running args.
@@ -144,27 +144,25 @@ background mode will always return immediately.
 
 ### Running your own config
 
-If you want to run your own config, we suggest you start from one of the existing configs. Then, if you're not using
-an NFS server or similar, you should upload your config to GCS:
+If you want to run your own config, we suggest you start from one of the existing configs. Just copy it to
+a new file:
 
-```bash
-gsutil cp my_config.yaml gs://my_bucket//my_config.yaml
+`cp config/gpt2_small.yaml config/my_config.yaml`
+
+If you're using `launch.py`, the config will be automatically uploaded as part of your Docker image, so you
+can just reference the local config path in your command line:
+
 ```
 
 Afterward, you can use the config directly from the TPU VM instance, e.g.:
 
 ```bash
-infra/babysit-tpu-vm <name> -z <zone> -t <type> [--preemptible] -- \
-    python infra/launch.py -- python levanter/src/levanter/main/train_lm.py --config_path gs://my_bucket/my_config.yaml \
+    python infra/launch.py -- python src/levanter/main/train_lm.py --config_path config/my_config.yaml \
     --trainer.checkpointer.base_path gs://path/to/checkpoints/
 ```
 
-The `--config_path` argument can be a local path, a GCS path, or any URL loadable by fsspec.
 With this configuration (unless `trainer.load_checkpoint` is false), Levanter will automatically
 try to load the latest checkpoint if it exists.
-
-Tokenizers are also loaded via fsspec, so you can use the same trick to load them from GCS if you have a custom
-tokenizer, or you can use an HF tokenizer.
 
 ## Common Issues
 ### (CRFM) Permission denied on `/files`
--- a/docs/Getting-Started-TPU-VM.md
+++ b/docs/Getting-Started-TPU-VM.md
@@ -85,18 +85,40 @@ the VM. That's explained down below in the [Running Levanter GPT-2](#running-lev
 ## Running Levanter GPT-2
 Now that you have a TPU VM instance, you can follow the [Getting Started](Getting-Started-Training.md) steps, but here are a few shortcuts:
 
-### Launch a GPT-2 Small in unattended mode (using nohup)
-```bash
-gcloud compute tpus tpu-vm ssh $NAME --zone $ZONE --worker=all --command 'WANDB_API_KEY=... levanter/infra/launch.sh python levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml --trainer.checkpointer.base_path gs://<somewhere>'
+### Launch a GPT-2 Small in unattended mode
+
+You will need a [Docker installation](https://docs.docker.com/engine/install/)
+on your development machine to build and run images on TPUs.
+
+First create a configuration file for future launches in your Levanter directory:
+
+```
+cat > .config <<EOF
+env:
+    WANDB_API_KEY:  ...
+    WANDB_ENTITY: ...
+    WANDB_PROJECT: levanter
+    HF_TOKEN: ...
+
+docker_repository: levanter
+zone: us-west4-a
+tpu: test-tpu
+EOF
 ```
 
-`launch.sh` will run the command in the background and redirect stdout and stderr to a log file in the home directory
-on each worker.
+Everything after the `--` is run on each worker.
+
+```bash
+python infra/launch.py -- python levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml --trainer.checkpointer.base_path gs://<somewhere>'
+```
+
+`launch.py` will package your directory and create and deploy a Docker image  on each worker.
 
 ### Launch a GPT-2 Small in interactive mode
-This version writes to the terminal, you should use tmux or something for long running jobs for this version. It's mostly for debugging.
+
+To run in the foreground, use `--foreground` with the `launch.py` script. You should use tmux or something for long running jobs for this version. It's mostly for debugging.
 ```bash
-gcloud compute tpus tpu-vm ssh $NAME --zone $ZONE --worker=all --command 'WANDB_API_KEY=... levanter/infra/run.sh python levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml --trainer.checkpointer.base_path gs://<somewhere>'
+python infra/launch.py -- python levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml --trainer.checkpointer.base_path gs://<somewhere>'
 ```
 
 ### Babysitting Script
@@ -113,11 +135,12 @@ You can run it like this:
 
 ```bash
 infra/babysit-tpu-vm <name> -z <zone> -t <type> [--preemptible]  -- \
-    WANDB_API_KEY=... levanter/infra/run.sh python levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml
+    python infra/launch.py -- levanter/src/levanter/main/train_lm.py --config_path levanter/config/gpt2_small.yaml
 ```
 
-That `--` is important! It separates the spin up args from the running args. Also, you should never use `launch.sh`
-with `babysit`, because nohup exits immediately with exit code 0.
+That `--` is important! It separates the spin up args from the running args.
+Also you should always use `--foregrouund` with `babysit-tpu-vm`, as the
+background mode will always return immediately.
 
 ### Running your own config
 
@@ -132,7 +155,7 @@ Afterward, you can use the config directly from the TPU VM instance, e.g.:
 
 ```bash
 infra/babysit-tpu-vm <name> -z <zone> -t <type> [--preemptible] -- \
-    WANDB_API_KEY=... levanter/infra/run.sh python levanter/src/levanter/main/train_lm.py --config_path gs://my_bucket/my_config.yaml \
+    python infra/launch.py -- python levanter/src/levanter/main/train_lm.py --config_path gs://my_bucket/my_config.yaml \
     --trainer.checkpointer.base_path gs://path/to/checkpoints/
 ```
 
--- a/infra/launch.py
+++ b/infra/launch.py
@@ -10,6 +10,7 @@ from infra.helpers import cli
 
 
 def setup_vm_docker(tpu_name, zone, docker_base_image):
+    """Change docker permissions on `tpu_name` and setup the cache volume."""
     cli.tpu_ssh(
         tpu_name,
         zone,
@@ -52,9 +53,7 @@ def list_tpus(zone):
     return tpus
 
 
-def start_tpu_vm(
-    tpu_name, *, tpu_type, preemptible, version, zone, autodelete, project, docker_repository, docker_base_image
-):
+def start_tpu_vm(tpu_name, *, tpu_type, preemptible, version, zone, autodelete):
     tpu_exists = any([tpu["NAME"] == tpu_name for tpu in list_tpus(zone)])
     if tpu_exists:
         if not autodelete:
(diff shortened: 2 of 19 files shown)


infra/launch.py
```diff
@@ -0,0 +1,106 @@
...
+
+import argparse
+import getpass
+import subprocess
+
+from infra import deploy
+from google.cloud import storage
+
+
+# Oddly enough, there's no API to simply fetch the current gcloud configuration...
```
dlwh, line 11: 🤦

infra/launch.py
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--tpu", type=str, required=True)
+    parser.add_argument("--project", type=str, default=gcloud_config()["project"])
+    parser.add_argument("--zone", type=str, required=True)
+    parser.add_argument("--docker_repository", type=str, required=True)
+    parser.add_argument("--wandb_api_key", type=str, required=False)
```
dlwh, line 25: i aspire to use google's secrets and/or some simple lightweight thing (read a .env file or something?) but haven't gotten there. a minimum, might be nice to have a -e (I sometimes pass in libtpu flags, e.g.)

infra/launch.py
```diff
@@ -0,0 +1,106 @@
...
+        "project": client.project,
+    }
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--tpu", type=str, required=True)
+    parser.add_argument("--project", type=str, default=gcloud_config()["project"])
+    parser.add_argument("--zone", type=str, required=True)
+    parser.add_argument("--docker_repository", type=str, required=True)
```
dlwh, line 24: would be nice if we could put this in a .config in the levanter base dir or something

infra/launch.py
```diff
@@ -0,0 +1,106 @@
...
+        "-t",
+        "--privileged",
+        "--shm-size=32gb",
+        "--net=host",
+        "-v",
+        "/home:/home",
+        "-v",
+        "/tmp:/tmp",
+        "-e",
+        "PYTHONPATH=/opt/levanter:/opt/levanter/src:/opt/levanter/examples",
```
dlwh, line 67:
```suggestion
        "PYTHONPATH=/opt/levanter:/opt/levanter/src:/opt/levanter/examples:/opt/levanter/tests",
```

(I've started running tests on TPU in CI)

infra/launch.py
```diff
@@ -0,0 +1,106 @@
...
+        region=region,
+        repository=docker_repository,
+        image_name=image_id,
+    )
+
+    full_image_id = f"{region}-docker.pkg.dev/{project}/{docker_repository}/{image_id}"
+
+    docker_command = [
+        "docker",
+        "run",
```
dlwh, line 57: the current semantics of `launch.sh` are that it runs in a nohup/background, whereas "run.sh" blocks until the job is done. I think this new script is blocking?

dlwh (commented):
This is amazing and wonderful. Thank you

Mostly just making sure I understand: the new workflow is:

* I hack on my own fork of levanter. 
* Then, when I'm ready, I just run the new launch command
* It will automatically spin up a tpu
* build a docker image from my current checkout
* push the image
* deploy and run the image on the tpu machine

One thing that I think might be lost (unless I'm misreading) is that ATM we log the git commit to wandb. Is that right?

infra/launch.py
rjpower: > would be nice if we could put this in a .config in the levanter base dir or something 

Good call. I added `.config_tpu` YAML and pulled default values for all the args from there. It looks like:

```
env:
    WANDB_API_KEY: ...
    WANDB_ENTITY: ...
    WANDB_PROJECT: levanter
    HF_TOKEN: ...

docker_repository: levanter
zone: us-west4-a
tpu: test-spin-up-32
```

infra/launch.py
rjpower: Ack, see above.

infra/launch.py
rjpower: Ack, applied.

infra/launch.py
rjpower: Good catch. I added a `--foreground` flag to toggle the behavior. By default it will run in the background.

(Docker will automatically capture stdout and stderr for all containers and makes it accessible via the `docker logs` command).

rjpower:
> This is amazing and wonderful. Thank you
> 
> Mostly just making sure I understand: the new workflow is:
> 
> * I hack on my own fork of levanter.
> * Then, when I'm ready, I just run the new launch command
> * It will automatically spin up a tpu
> * build a docker image from my current checkout
> * push the image
> * deploy and run the image on the tpu machine
> 
> One thing that I think might be lost (unless I'm misreading) is that ATM we log the git commit to wandb. Is that right?

Yes mostly: I don't spin up the new TPU yet: you still use `spin-up-vm.sh`, and that pulls a "default" image, just to warm up the VM. From that point on, when you run launch, it's exactly as you say. It takes negligible time to setup and push any changes, as it's just the last "layer" of the docker image and <1MB of data.

I think having the combined workflow is a really good idea though and I can add that in a follow-up PR.

rjpower (commented):
I cannot for the life of me figure out the review process on Github: I can only find the "Submit Review" button from inside vscode so here goes...

rjpower:
I'll update the docs and validate testing works and then give you a ping.

rjpower:
Okay, I think I caught all the docs etc. I updated the .github test integration to use `launch.py` but it won't run directly for me: I don't think Github will inject credentials for external PRs, so it's just failing in spin up. 

I've verified it runs correctly (there are a few test failures, but they aren't my fault 😁) with the following command manually:

```
python infra/launch.py --foreground --tpu=test-spin-up-1 --zone=us-west4-a -- /opt/levanter/.venv/bin/pytest tests
```

We can of course tweak `launch.py` to use env/secrets as needed for that.

PTAL.

infra/launch.py
```diff
@@ -0,0 +1,129 @@
...
+    project = args.project
+    tpu_name = args.tpu
+    image_id = args.image_name
+    docker_repository = args.docker_repository
+    command = args.command
+    foreground = args.foreground
+    env = {k: v for k, v in args.env}
+
+    if not "WANDB_PROJECT" in env:
+        env.append["WANDB_PROJECT"] = "levanter"
```
dlwh, line 71:
```suggestion
        env["WANDB_PROJECT"] = "levanter"
```

docs/Getting-Started-TPU-VM.md
```diff
@@ -85,18 +85,37 @@ the VM. That's explained down below in the [Running Levanter GPT-2](#running-lev
 ## Running Levanter GPT-2
 Now that you have a TPU VM instance, you can follow the [Getting Started](Getting-Started-Training.md) steps, but here are a few shortcuts:
 
-### Launch a GPT-2 Small in unattended mode (using nohup)
+### Launch a GPT-2 Small in unattended mode
```
dlwh, line 88: we should tell people to install docker and have it running on their development machine

infra/launch.sh
```diff
@@ -1,23 +0,0 @@
-#!/bin/bash
```
dlwh, line 1: can we leave this here for a bit just to give people a chance to migrate

infra/launch.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/python
```
dlwh, line 1: ok, actually having tried this out now, it's kinda super slow for first time getting started. Is this just the "base image" slowness?

Do you think we can have a "base base image" that we host on a public docker registry (ghcr or something) that speeds this up a bit more for first time (with a possibly stale install of most deps or something)?

dlwh (commented):
still trying it out, but mostly looks good to me! just thinking through the migration path and the first time launch experience.

dlwh:
Running directly off your branch I get 

```
Running  docker tag levanter-dlwh us-west4-docker.pkg.dev/hai-gcp-models/levanter/levanter-dlwh:1716873733
Error response from daemon: No such image: levanter-dlwh:latest
  File "/Users/dlwh/src/levanter/infra/launch.py", line 79, in <module>
    full_image_id = deploy.push_to_gcp(
  File "/Users/dlwh/src/levanter/infra/deploy.py", line 133, in push_to_gcp
    _run(["docker", "tag", image_name, full_image_name])
  File "/Users/dlwh/src/levanter/infra/deploy.py", line 35, in _run
    return subprocess.check_output(*args, **kw)
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['docker', 'tag', 'levanter-dlwh', 'us-west4-docker.pkg.dev/hai-gcp-models/levanter/levanter-dlwh:1716873733']' returned non-zero exit status 1.
```

rjpower:
> Running directly off your branch I get
> 
> ```
> Running  docker tag levanter-dlwh us-west4-docker.pkg.dev/hai-gcp-models/levanter/levanter-dlwh:1716873733
> Error response from daemon: No such image: levanter-dlwh:latest
> Traceback (most recent call last):
>   File "/Users/dlwh/src/levanter/infra/launch.py", line 79, in <module>
>     full_image_id = deploy.push_to_gcp(
>   File "/Users/dlwh/src/levanter/infra/deploy.py", line 133, in push_to_gcp
>     _run(["docker", "tag", image_name, full_image_name])
>   File "/Users/dlwh/src/levanter/infra/deploy.py", line 35, in _run
>     return subprocess.check_output(*args, **kw)
>   File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/subprocess.py", line 421, in check_output
>     return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
>   File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/subprocess.py", line 526, in run
>     raise CalledProcessError(retcode, process.args,
> subprocess.CalledProcessError: Command '['docker', 'tag', 'levanter-dlwh', 'us-west4-docker.pkg.dev/hai-gcp-models/levanter/levanter-dlwh:1716873733']' returned non-zero exit status 1.
> ```

Doh, that's probably something I just screwed up: I added a timestamp to uniquely identify the tagged image, and neglected to update that command. So you can understand my pain:

Docker has a "latest" tag which works mostly as you expect, _except_ when you're using remote images.
When you're using a remote image, and you use

`docker pull xyz.pkg.dev/levanter/levanter:latest`

it works correctly the first time, but if you update the package on `pkg.dev` and call pull again, there's no check for staleness: Docker just assumes if it has that tag locally everything must be okay. So basically we can't use the "latest" tag except for things like a base image that we don't mind going stale.

infra/launch.py
rjpower: Oops good catch, thanks.

docs/Getting-Started-TPU-VM.md
rjpower: Done.

infra/launch.sh
rjpower: Good call, reverted.

infra/launch.py
rjpower: Yeah, I was thinking the same thing. It is a one-time cost for your development environment, but yeah, it has to install everything from scratch and then push it to GCP for the first build/launch. I went ahead and split the Dockerfiles into base and incremental: you'll still need to download the base layers to your local machine, but this should be as fast as your internet connection.

For the base image: I'd like to use Dockerhub but I can't see a way to create a "levanter" repo without creating a paid organization. Using GCP is better performance wise: you won't have to "re-push" the base layers you download when building, but there's some egress cost to pay.

I have it pointing at my image for testing, but if you can push to the Stanford repo (or point me at a project to push to) we can use that instead.

rjpower:
Okay, it's faster with the incremental build (and a little more predictable).

```
[+] Building 44.5s (10/10) FINISHED
```

vs building the base image:

```
[+] Building 169.1s (16/16) FINISHED                                                                                                                                                                                     docker:desktop-linux
```

In both cases, the build is cached, so you should only pay this once.  Further deployments are almost instantaneous, e.g. to touch a single file I get:

```
[+] Building 0.1s (10/10) FINISHED 
```

infra/launch.py
dlwh: Yeah Docker is stingy about free orgs now. We can use github's repo for the base image? I'll try that out. (I'll also see about granting you repo access)

dlwh:
> it works correctly the first time, but if you update the package on `pkg.dev` and call pull again, there's no check for staleness: Docker just assumes if it has that tag locally everything must be okay. So basically we can't use the "latest" tag except for things like a base image that we don't mind going stale.

Ah yes, Docker is very annoying about things like that. :(

dlwh:
I had to run 

`gcloud auth configure-docker us-central2-docker.pkg.dev`

But now I get 

```
name unknown: Repository "levanter" not found
  File "/Users/dlwh/src/levanter-dev/infra/launch.py", line 83, in <module>
    full_image_id = push_docker.push_to_gcp(
  File "/Users/dlwh/src/levanter-dev/infra/push_docker.py", line 153, in push_to_gcp
    _run(["docker", "push", full_image_name])
  File "/Users/dlwh/src/levanter-dev/infra/push_docker.py", line 35, in _run
    return subprocess.check_output(*args, **kw)
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['docker', 'push', 'us-central2-docker.pkg.dev/hai-gcp-models/levanter/levanter-dlwh:1716924862']' returned non-zero exit status 1.
```

rjpower:
> I had to run
> 
> `gcloud auth configure-docker us-central2-docker.pkg.dev`
> 
> But now I get
> 
> ```
> name unknown: Repository "levanter" not found

Crap, that was me: I was trying to factor out the push setup to test out pushing to Docker hub and removed the GCP setup from the `push_to_gcp` flow.  Can you try again now?

I'll also try testing with a different repository name to see if that breaks anything.

rjpower:
I went ahead and added a few cleanups:

* The launch script now has retries, TPU starting and setup, so you can use it in place of babysit+spinup+setup+run
* I added pushing to ghcr.io: https://ghcr.io/rjpower/levanter. The first push to GCP will take a while with this setup, as you have to push the ~500MB base image, but the build is still quick and after that everything is speedy.
* Refactored so the scripts share some CLI helpers etc

I found this setup nice for dependencies, for instance, I had a few changes to Haliax so I just copied `haliax` into my levanter directory and added `cd haliax && pip install -e .` to the Dockerfile.

Let me know if you see any other things you'd like addressed!

dlwh:
sorry getting back to this, some thoughts:

* it'd be nice if `launch` would be willing to reuse an existing tpu-vm if it already exists
* small bytes-vs-str thing in configure_gcp_docker I'll add a suggestion for (mostly note to self)
* after fixing that, i'm getting this:

```
Running: gcloud alpha compute tpus tpu-vm ssh dlwh-test --worker=all --zone=us-east1-d --command=docker volume create --driver=local levanter
SSH key found in project metadata; not updating instance.
Using ssh batch size of 4. Attempting to SSH into 1 nodes with a total of 4 workers.
SSH: Attempting to connect to worker 0...
SSH: Attempting to connect to worker 1...
SSH: Attempting to connect to worker 2...
SSH: Attempting to connect to worker 3...
levanter
levanter
levanter
levanter
Running  gcloud artifacts repositories describe --location=us-east1 levanter
ERROR: (gcloud.artifacts.repositories.describe) NOT_FOUND: Requested entity was not found.
Error running command.
```

infra/push_docker.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def configure_gcp_docker(project_id, region, repository):
+    """Setup Artifact registry repository and configure permissions to enable TPU access."""
+    # check if the repository already exists
+    try:
+        _run(["gcloud", "artifacts", "repositories", "describe", f"--location={region}", repository])
+        return
+    except subprocess.CalledProcessError as e:
+        if "NOT_FOUND" not in e.output:
```
dlwh, line 49:
```suggestion
        if b"NOT_FOUND" not in e.output:
```

(so annoying)

infra/launch.py
```diff
@@ -0,0 +1,211 @@
...
+    config = cli.load_config()
+
+    cli.add_arg(parser, config, ["--autodelete"], default=False, action="store_true")
+    cli.add_arg(parser, config, ["--docker_base_image"], default="ghcr.io/rjpower/levanter:latest")
+    cli.add_arg(parser, config, ["--docker_repository"], default="levanter")
+    cli.add_arg(parser, config, ["--foreground"], default=False, action="store_true")
+    cli.add_arg(parser, config, ["--image_name"], default=f"levanter-{getpass.getuser()}")
+    cli.add_arg(parser, config, ["--preemptible"], default=False, action="store_true")
+    cli.add_arg(parser, config, ["--project"], default=cli.gcloud_config()["project"])
+    cli.add_arg(parser, config, ["--tpu"], required=True)
```
dlwh, line 104: sorry to bikeshed, but can we call this 'name' maybe? I know `tpu_type` is right there, but it kind of looks like the kind of tpu you want

infra/launch.py
```diff
@@ -0,0 +1,211 @@
...
+    cli.add_arg(parser, config, ["--tpu_type"])
```
dlwh, line 105: this should be required no?

dlwh:
lol now i'm getting this completely useless error

```
~/src/levanter-dev (faster-setup)$ PYTHONPATH=. python infra/launch.py --foreground --tpu=test-spin-up-1 --zone=us-central2-b --tpu_type v4-8 -- /opt/levanter/.venv/bin/pytest tests
Listed 0 items.
Creating new TPU test-spin-up-1 in us-central2-b of type v4-8...
Running: gcloud alpha compute tpus tpu-vm create test-spin-up-1 --accelerator-type=v4-8 --version=tpu-ubuntu2204-base --zone=us-central2-b  --quiet
ERROR: (gcloud.alpha.compute.tpus.tpu-vm.create) unrecognized arguments:

To search the help text of gcloud commands, run:
  gcloud help -- SEARCH_TERMS
Error running command.
```

dlwh:
repr helped me figure it out, lol:

Running: 'gcloud' 'alpha' 'compute' 'tpus' 'tpu-vm' 'create' 'test-spin-up-1' '--accelerator-type=v4-8' '--version=tpu-ubuntu2204-base' '--zone=us-central2-b' '' '--quiet'

infra/launch.py
```diff
@@ -0,0 +1,211 @@
...
+        "tpus",
+        "tpu-vm",
+        "create",
+        tpu_name,
+        f"--accelerator-type={tpu_type}",
+        f"--version={version}",
+        "--zone=" + zone,
+        "--preemptible" if preemptible else "",
+        "--quiet",
+    )
```
dlwh, line 90:
```suggestion
    command = [
        "gcloud",
        "alpha",
        "compute",
        "tpus",
        "tpu-vm",
        "create",
        tpu_name,
        f"--accelerator-type={tpu_type}",
        f"--version={version}",
        "--zone=" + zone,
        "--quiet",
    ]
    if preemptible:
        command.append("--preemptible")
    cli.run_command(*command)
```

docs/Getting-Started-TPU-VM.md
```diff
@@ -124,18 +131,11 @@ python infra/launch.py -- python levanter/src/levanter/main/train_lm.py --config
...
 If you are using a preemptible TPU VM, you probably want to use the "babysitting" script that automatically re-creates
-the VM. This is because preemptible instances can be preempted and will always be killed every 24 hours. The babysitting
-script handles both the creation of the node and the running of a job, and also relaunches the TPU VM if it gets preempted.
-It keeps running the command (and relaunching) until the command exits successfully.
-
-Note that the babysitting-script will automatically set the `RUN_ID` environment variable if not set, and pass it to the
-training command. This ensures that restarted jobs have the same run id, which is important for resumes to work.
-
-You can run it like this:
+the VM. This is because preemptible instances can be preempted and will always be killed every 24 hours. You can run `launch.py` with the `--retries` and `--foreground` parameter to accomplish this. If `--retries` is greater than 1, `launch.py` will automatically attempt to re-create the VM and re-run the command if it fails. (`--foreground` is necessary to keep the script from returning immediately.)
```
dlwh, line 134: FWIW, a desideratum I'll have after this is for us to use [queued-resources](https://cloud.google.com/tpu/docs/queued-resources) which is a kinda annoying API in the current regime but might be better with this fix in. In this world, you want to pass our setup script as a vm setup script rather than scping it. See #597 / #328 

The advantages of queued-resources are:

* there are now "best effort" vms which are exactly like preemptible AFAICT except they don't get auto-killed every 24 hours.
*  multislice works

The downside is that you maybe have to manually poll if you want to launch a command separately, and you have to poll separately for the VM getting preempted anyway...

docs/Getting-Started-TPU-VM.md
```diff
@@ -144,28 +144,26 @@ background mode will always return immediately.
...
     --trainer.checkpointer.base_path gs://path/to/checkpoints/
 ```
 
-The `--config_path` argument can be a local path, a GCS path, or any URL loadable by fsspec.
 With this configuration (unless `trainer.load_checkpoint` is false), Levanter will automatically
 try to load the latest checkpoint if it exists.
 
-Tokenizers are also loaded via fsspec, so you can use the same trick to load them from GCS if you have a custom
-tokenizer, or you can use an HF tokenizer.
-
```
dlwh, line 145: I should find a place for this to be documented. We have used it before

dlwh (approved):
lgtm, thanks!

rjpower:
Okay, cleaned everything up. I think I can't submit because the TPU tests are failing for an unrelated reason so you may need to merge for me:

```
gcloud compute tpus tpu-vm delete ci-run-9473410622 --zone us-central2-b --quiet
ERROR: (gcloud.compute.tpus.tpu-vm.delete) Error parsing [tpu].
The [tpu] resource is not properly specified.
Failed to find attribute [project]. The attribute can be set in the following ways: 
- provide the argument `tpu` on the command line with a fully specified name
- provide the argument `--project` on the command line
- set the property `core/project`
```

It appears github secrets aren't available in forked repos: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow (even if approved). I couldn't find a good workaround unfortunately; I'm sure there's some rube-goldberg machine setup where you can have GCP cloud run pick up approved requests and call web services, but I expect it would work 4% of the time.

dlwh:
> I couldn't find a good workaround unfortunately; I'm sure there's some rube-goldberg machine setup where you can have GCP cloud run pick up approved requests and call web services, but I expect it would work 4% of the time.

a very simple solution is for me to add you to the repo :-)

dlwh:
(I don't care enough about pre-commit. we can fix it later)

dlwh:
Thanks so much!

PR #606, opened 2024-05-31
let's try upcasting rmsnorm

dlwh:


src/levanter/models/llama.py
```diff
@@ -305,17 +305,20 @@ def update_state_dict(self, state_dict: StateDict, prefix: Optional[str] = None)
...
+    axis: AxisSpec = eqx.static_field()
+    weight: Optional[NamedArray]
+    bias: Optional[NamedArray]
+
+    eps: float = eqx.static_field(default=1e-5)
+    dtype: Optional[jnp.dtype] = eqx.static_field(default=jnp.float32)
 
     @staticmethod
-    def init(axis: AxisSpec, eps: float = 1e-6, use_weight: bool = True, use_bias: bool = False):
+    def init(axis: AxisSpec, eps: float = 1e-5, use_weight: bool = True, use_bias: bool = True, dtype=jnp.float32):
```
Ivan-Zhou, line 321: What's your rationale behind these changes to the default value?

Llama's implementation doesn't have bias, right?
https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L88

The default rms eps is 1e-6: 
https://huggingface.co/docs/transformers/en/model_doc/llama2#transformers.LlamaConfig

Ivan-Zhou (commented):
Do you want me to try this change on our Llama 1B run?

src/levanter/models/llama.py
dlwh: copypasta

dlwh, line 321:
```suggestion
    def init(axis: AxisSpec, eps: float = 1e-6, use_weight: bool = True, use_bias: bool = True, dtype=jnp.float32):
```

dlwh:
@Ivan-Zhou yes please!

PR #607, opened 2024-05-31
Create HF repo if it doesn't exist

versae:
The `HFCheckpointConverter.save_pretrained()` function needs an existing repo for the push to HF to be successful. This PR checks whether the repo already exists, and if not, it creates a private one so the training doesn't fail because of the push to HF failed.

Feel free to ignore if the functionality is not desired.

### DIFF
--- a/src/levanter/compat/hf_checkpoints.py
+++ b/src/levanter/compat/hf_checkpoints.py
@@ -21,7 +21,8 @@ import jax.numpy as jnp
 import mergedeep
 import safetensors
 import safetensors.numpy
-from huggingface_hub import hf_hub_download, snapshot_download
+
+from huggingface_hub import HfApi, hf_hub_download, repo_exists, snapshot_download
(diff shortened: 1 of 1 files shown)


src/levanter/compat/hf_checkpoints.py
```diff
@@ -757,6 +759,9 @@ def save_pretrained(
                 upload_to_hf = self.reference_checkpoint
             if not isinstance(upload_to_hf, bool):
                 assert isinstance(upload_to_hf, (str, RepoRef))
+                if isinstance(upload_to_hf, str) and not repo_exists(upload_to_hf, repo_type="model"):
```
versae, line 762: I assume a `RepoRef` can only exist when an actual repo also exists in HF.

dlwh: makes sense to me!

dlwh (approved):
Thanks!

PR #617, opened 2024-06-07
some changes in prep for changing default linear output order in Haliax

dlwh:


dlwh:
cc @Helw150 changing Linear to be out_first=True by default, which is not how gpt2 does thing (and presumably whisper?)

PR #618, opened 2024-06-08
fix some tests in prep for new initialization, 

dlwh:
which caught some problems

tagging @rjpower . Turns out we need to use approx=True for gelu in JAX to really match Gemma

### DIFF
--- a/src/levanter/models/llama.py
+++ b/src/levanter/models/llama.py
@@ -257,6 +257,13 @@ class LlamaAttention(StateDictSerializationMixin, eqx.Module):
         )
         return LlamaAttention(config, q_proj, k_proj, v_proj, o_proj)
 
+    def _rope_scale_factor(self) -> float:
+        # hasattr for gemma and I'm feeling lazy
+        if hasattr(self.config, "rope_scaling") and self.config.rope_scaling is not None:
+            assert self.config.rope_scaling["type"] == "linear"
+            return self.config.rope_scaling["factor"]
+        return 1.0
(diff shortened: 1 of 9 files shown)


dlwh:
(also i know the hasattr is gross, I'll clean it up)

src/levanter/models/llama.py
```diff
@@ -257,6 +257,13 @@ def init(config: LlamaConfig, *, key) -> "LlamaAttention":
         )
         return LlamaAttention(config, q_proj, k_proj, v_proj, o_proj)
 
+    def _rope_scale_factor(self) -> float:
```
rjpower, line 260: Feel free to file an issue and I can look at normalizing how Gemma handles this. I can't remember if HFGemma doesn't use it at all or if it was doing something different and weird.

rjpower (approved):
Nice catch! Here I went through all those hoops to deal with the weird HF activation handling and I managed to replicate the same issue...

src/levanter/models/llama.py
dlwh: HfGemma doesn't use it at all, which seems like an oversight more than anything. (Llama 2 also doesn't use it by default, but you can turn it on and do some tuning to get the length extrapolation, I think)

dlwh:
yeah, these things are kind of finicky. It's so easy to have something that seems right but it's wrong in some tiny detail like this. It doesn't help that gelu and gelu_new are completely confusing

PR #623, opened 2024-06-11
Faster setup (so i can run tpu tests)

dlwh:


dlwh:
@rjpower the tpu tests aren't your fault. i think they're fixed in main

rjpower (approved):

PR #628, opened 2024-06-12
Combine some commands to make docker launches faster.

rjpower:
Adds a PTY to show Docker push etc status and merges some commands to speed launches.

Before:

touch config/0; python infra/launch.py --foreground --tpu_name=test-spin-up-8  --  --version  7.16s user 1.22s system 31% cpu 26.867 total

After:

touch config/1; python infra/launch.py --foreground --tpu_name=test-spin-up-8  --  --version  3.65s user 0.67s system 29% cpu 14.765 total

### DIFF
--- a/infra/launch.py
+++ b/infra/launch.py
@@ -10,7 +10,7 @@ from infra.helpers import cli
 
 
 def setup_vm_docker(tpu_name, zone, docker_base_image):
-    """Change docker permissions on `tpu_name` and setup the cache volume."""
+    """Change docker permissions on `tpu_name`, remove any old runs, and setup the cache volume."""
     cli.tpu_ssh(
         tpu_name,
         zone,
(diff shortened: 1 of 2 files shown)


infra/launch.py
```diff
@@ -19,11 +19,18 @@ def setup_vm_docker(tpu_name, zone, docker_base_image):
         "-aG",
         "docker",
         getpass.getuser(),
+        "&&" "docker",
+        "volume",
+        "create",
+        "--driver=local",
+        "levanter",
+        "&&",
```
dlwh, line 27: I don’t have an opinion here just noting: previously this would happen regardless and now it skips if previous step fails

dlwh (approved):

dlwh:
Ugh I broke all the tests. I will fix tonight

infra/launch.py
rjpower: Yeah, I don't love it, but in theory the behavior should be the same (because `tpu_ssh` by default aborts if the command fails). 

(All of the commands should always succeed. `docker volume create` and `docker rm -f levanter` both always return 0 for most normal situations. It's a no-op if the volume already exists, and `rm -f` returns 0 if the container doesn't exist.)

dlwh (approved):

PR #631, opened 2024-06-13
Use wandb compatible run ids.

rjpower:
When presented with a runid in base36 encoding, wandb will generate a pretty name for the run. It otherwise uses the runid as the run name.


dlwh (approved):
thanks! such a weird policy

PR #633, opened 2024-06-14
 Try out adding some cache metadata to check for bad tokenizers. 

rjpower:
There's probably a bit cleaner way to do this, but just tried adding some simple metadata to check if the tokenizer or vocab size is the same as the cached version.

(Dumb question: Is the tokenizer slow enough that we need to cache it? Or could you just reshard characters and change the tokenizer at will? I suppose the way documents are sliced depends on the tokenizer...)


dlwh (approved):
You can tokenize faster than you can train for even moderately sized models, but I wasn't totally sure how to implement seek/instant resume (which I admit we haven't implemented, but I swear I'm about to!) because you slice docs and the mapping from strings to tokens isn't straightforward without tokenizing. Probably one could be sufficiently clever and remember byte offsets to avoid it.

And we want to get to a proper shuffle anyhow... I have a plan, but I don't know if it's a good plan.

dlwh (approved):

rjpower:
Ack, I figured it was the seeking/sharding that would be more challenging.

If you have ideas on even adding non-proper shuffle, I'd be interested in using it! I admit I didn't _really_ try to figure out how the dataset processing is working yet 😄. I was trying some experiments with multiple LoRA models recently and found that varying the initial weights resulted in the same end behavior, but the input shuffling seems to impact the behavior...

Well, at least in my crappy torchtune implementation that has bunch of other randomness, but that's why I'd like to confirm with Levanter!

dlwh:
we recently added shuffle buffers actually

```yaml
data_seed: 42  # if you want same init but different data seeds
data:
  cache_dir: ...
  stop_strategy: restart
  shuffle_buffer_size: 100000
```

PR #634, opened 2024-06-15
Fix #418: add flop calculations for models

### LINKED ISSUE #418: Implement more accurate MFU computation
https://github.com/karpathy/nanoGPT/issues/322

We've been underselling ourselves badly for small models.

model_flops should probably be a property/computation on LMHeadModel that needs to be implemented by subclasses.
Something like

```python
def model_flops(self, example: LMExample) -> float:
   raise NotImplementedError
```

and in GPT2 it would look something like

```python
def model_flops(self, example: LMExample) -> float:
   num_tokens = example.input_ids.size
   num_layers = self.config.num_layers
   num_params = parameter_count(self)
   return 6 * num_tokens * num_params + the attention aware calculation
```

We can then improve/refactor the log_performance_stats metric to log Model-Flops/second

As a follow-up step we can add some device detection to get theoretical flops and do the division.

Aphoh:
Implements flop calculations for models. This is then piped into the trainer of `lm_trainer` for throughput logging.

### DIFF
--- a/src/levanter/main/train_lm.py
+++ b/src/levanter/main/train_lm.py
@@ -159,7 +159,11 @@ def main(config: TrainLmConfig):
             )
             trainer.add_hook(cb, every=config.trainer.steps_per_eval)
 
-        trainer.add_hook(callbacks.log_performance_stats(Pos.size, trainer.config.train_batch_size), every=1)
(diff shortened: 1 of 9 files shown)


dlwh (approved):
Thanks for this, I appreciate it!

PR #638, opened 2024-06-17
A few "cleanups" for model instantiation.

rjpower:
* Makes the HF converter an instance method instead of a class property. This
  makes it easier to instantiate a model family like LLama with a different
  reference checkpoint.

* Change the HF checkpoint converter to always accept an LmConfig instead of a
  bare model or config object. There were only a few places where the model was
  still being passed in.

### DIFF
--- a/examples/alpaca-lora/alpaca_lora.py
+++ b/examples/alpaca-lora/alpaca_lora.py
@@ -90,7 +90,7 @@ def train(config: TrainArgs):
         logger.info(f"Loading pretrained model from {converter.reference_checkpoint}")
         # load untrainable params in compute precision to save memory
         model: LmHeadModel = converter.load_pretrained(  # type: ignore
-            model_config, axis_mapping=parameter_axis_mapping, dtype=trainer.mp.compute_dtype
+            model_config.model_type, model_config, axis_mapping=parameter_axis_mapping, dtype=trainer.mp.compute_dtype
         )
 
         # Major difference from Alpaca: we loraize the model.
@@ -119,7 +119,7 @@ def train(config: TrainArgs):
 
         logger.info(f"Total parameter count: {all_param_count}")
         logger.info(f"Trainable parameter count: {just_lora_params}")
-        logger.info(f"Fraction of parameters that are trainable: {just_lora_params * 1.0 / all_param_count%.3}")
+        logger.info(f"Fraction of parameters that are trainable: {just_lora_params * 1.0 / all_param_count:.3f}")
 
         # Levanter has two kinds of data loaders: sharded and replicated. Replicated is simpler and allows for
         # single pass training. Sharded only loads a subset of the data on each device, and is more efficient for large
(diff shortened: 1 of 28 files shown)


rjpower (commented):
The only reason we need the model class as a parameter is because of the ASRModel mixin. I'm not sure if the way ASR/Whisper is structured is required, or if Whisper model could just inherit from ASRModel itself; if so we could remove the extra param from HF loading.

Alternatively we could create the ASR model from the whisper model tree after the fact but I didn't investigate that.

src/levanter/compat/torch_serialization.py
```diff
@@ -421,29 +420,15 @@ def get_to_cpu(arr):
...
-                out = jax.jit(_identity_fn, out_shardings=sharding)(arr)
-                return np.array(out)
+                shardings = [None for i in range(len(arr.shape))]
+                sharding = NamedSharding(arr.sharding.mesh, PartitionSpec(*shardings))
+
+                def _copy(in_array):
+                    return jax.lax.with_sharding_constraint(in_array, sharding)
+
+                # arr = jax.jit(_copy, donate_argnums=0)(arr)
+                arr = jax.device_put(arr, sharding)
```
dlwh, line 430: As per the deleted comment, I don't think this will work to serialize a 7B model (in fp32)?

dlwh (commented):
So I think the simplified sharding thing won't work in general? Otherwise changes lgtm

dlwh:
@Helw150 what do you think about ^^

Helw150:
Looking at the code, I don't see a good reason it's structured the way I have it now 😅 . I think it's fine to just have Whisper inherit from ASRModel directly.

examples/alpaca-lora/alpaca_lora.py
```diff
@@ -119,7 +119,7 @@ def loraize_hf_model(model):
 
         logger.info(f"Total parameter count: {all_param_count}")
         logger.info(f"Trainable parameter count: {just_lora_params}")
-        logger.info(f"Fraction of parameters that are trainable: {just_lora_params * 1.0 / all_param_count%.3}")
```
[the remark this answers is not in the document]
rjpower: Ha yeah either flake8 or mypy just refused to let me leave this be...

src/levanter/compat/torch_serialization.py
rjpower: Ack, didn't mean to include this here, though I have an alternative solution which avoids some of the weird sharding issues. I'll put that in a separate PR to look at.

rjpower:
Thanks for the feedback, I'll keep ASR as is for now and clean-up in the next pass. I think a little "de-factoring" of the converter could make it a bit easier to follow but I'll think about it some more first.

(for now this gets me running with Llama 3 so back to experiments...)

dlwh (approved):

rjpower:
The only test failures are in llama flops: it looks like we don't have an HF token set in the secrets so we can't load that at the moment.

PR #639, opened 2024-06-18
Add chunked device->host copy for torch serialization.

rjpower:
Previously we applied a naive sharding to arrays when copying to the host to
avoid risking a device OOM. This instead flattens and copies arrays in chunks
to the CPU to achieve the same effect but avoids some issues with how JAX
handles device meshes.

### DIFF
--- a/src/levanter/compat/torch_serialization.py
+++ b/src/levanter/compat/torch_serialization.py
@@ -442,7 +442,7 @@ def to_numpy_state_dict(model, prefix: Optional[str] = None) -> StateDict:
 
                 shardings = [None if i != axis_to_shard else "device" for i in range(len(arr.shape))]
                 sharding = NamedSharding(process_mesh, PartitionSpec(*shardings))
-                out = jax.jit(_identity_fn, out_shardings=sharding)(arr)
+                out = jax.device_put(arr, sharding)
                 return np.array(out)
 
         # need to make sure the model is on *this machine* and *this machine's CPU* before saving


src/levanter/compat/torch_serialization.py
```diff
@@ -421,29 +421,38 @@ def get_to_cpu(arr):
...
-                return np.array(out)
+                # Replicating a large variable when tight on memory can push us over the TPU memory limit.
+                # We instead flatten our source array, slice out copy_chunk_size bytes at a time, concatenate
+                # the resulting arrays on the CPU, and then reshape back to our original shape.
+                shardings = [None]
+
+                sharding = NamedSharding(arr.sharding.mesh, PartitionSpec(*shardings))
+
+                def _slice_and_replicate(in_array, offset, size):
+                    in_array = jax.numpy.ravel(in_array)
```
dlwh, line 432: I'm just a tad worried this can trigger a deshard which kind of defeats the purpose (my assumption is flattening unshards the array until proven otherwise). Any chance I could impose upon you to check?

Though maybe I should trust XLA more

rjpower: It's a fair concern... I'm not sure how best to test it. I can't seem to allocate smaller than a v5-16; I have some v4 quota so I'll see if I can get a v4-8 working.

What configuration was failing in the past? IIUC for this to trigger, you need a model where weights + optimizer state is almost equal to accelerator memory, and then one giant tensor (like the embeddings) which pushes this over the edge? It needs to be a single tensor because we convert them one by one. I'm surprised though because then you'd have almost no space for activations either.

dlwh: Ok yeah so I'm pretty sure this de-shards, even in jax 0.4.30. 

(Welcome to the first layer of inner mysteries of JAX-on-XLA-on-TPU.)

Here's a test script: https://gist.github.com/dlwh/54be96f93f395071ad4688d2ee356c90#file-slice-py

and here's the output from running with `XLA_FLAGS="--xla_dump_to=/tmp/output_folder/xla_dumps --xla_dump_hlo_pass_re=.*" python slice.py` on a v4-8:

https://gist.github.com/dlwh/54be96f93f395071ad4688d2ee356c90#file-module_0014-jit__slice_and_replicate-after_optimizations_after_buffer_assignment-txt

This line is what I worried about:

```
all-gather = f32[134217728]{0:T(1024)} all-gather(bitcast.2), channel_id=1, replica_groups={{0,1,2,3}}, dimensions={0}, use_global_device_ids=true, metadata={op_name="jit(_slice_and_replicate)/jit(main)/dynamic_slice[slice_sizes=(524288,)]" source_file="/home/dlwh/slice.py" source_line=24}, backend_config={"flag_configs":[],"barrier_config":{"barrier_type":"CUSTOM","id":"0"},"scoped_memory_configs":[],"used_scoped_memory_configs":[]}
```

(13421778 is the full size of the array)

even changing the line to this:

```
 in_array = jax.lax.with_sharding_constraint(jnp.ravel(in_array), in_sharding)
```

doesn't do what we we need (the output is identical up to variable names)

rjpower: Argh, so frustrating. I think it's possible XLA will fuse things later on and avoid allocations, but it's impossible to tell without setting things up "just so" to trigger the bad condition. (Do we have a configuration that reliably does this)?

I tried changing this to slice over the largest dimension instead of flattening the arrays, but now I get errors in my test where somehow a "GSPMDSharding" has been introduced, which conveniently doesn't have a mesh attached. If I try to something like:

`mesh = np.ravel(list(arr.sharding.device_set), "r")` 

then I get mismatched

yashk2810:
Have you tried transferring to `pinned_host` memory space? `jax.device_put(x, NamedSharding(mesh, pspec, memory_kind='pinned_host')`

you won't have to chunk things up then.

rjpower:
Per advice from @yashk2810, I just changed this to a `device_put(arr, sharding)`. I'm not sure if the `pinned_host` spec is also necessary here?

dlwh (approved):
awesome!

PR #640, opened 2024-06-21
Fix torch tests

dlwh:


dlwh:
cc @Aphoh

Aphoh:
@dlwh I have some fixes for the gemma torch tests in my histogram branch I think

PR #641, opened 2024-06-21
Add optional gate activation histogram logging during eval

Aphoh:
The main goal of this PR is to add the ability to log activation statistics of the MLPs of models.

In it's current state, this involves one big, slightly inconvenient change: every model's `compute_loss` function now returns a tuple of `(loss: Array, extras: dict)` where `extras` can contain any auxiliary data to log. Thus all the upstream code had to be modified to accomidate this change.

Currently there's code to measure the activation statistics of llama models during eval only, as computing the histograms is incredibly inefficient on TPUs. For LLaMa 7b, computing the histograms takes roughly 4x as long as the rest of the forward pass. AFAIK there's no faster way to do this, but it's just during eval so 🤷.

The code's a little messy, so some review would be appreciated.


src/levanter/models/gemma.py
```diff
@@ -129,7 +133,9 @@ def from_hf_config(cls, hf_config: HfConfig):
         if hf_config.hidden_activation:
             activation_function = hf_config.hidden_activation
         else:
-            activation_function = hf_config.hidden_act
+            # This is the implementation in huggingface
+            # https://github.com/huggingface/transformers/blob/12b1620e615592fbf099d4ec44af7b9f2d1b48aa/src/transformers/models/gemma/modeling_gemma.py#L200
+            activation_function = "gelu_pytorch_tanh"
```
dlwh, line 138: i swore we already did this

src/levanter/eval.py
```diff
@@ -123,6 +125,17 @@ def eval_callback(step: StepInfo):
             _join_prefix(prefix, "loading_time"): result.total_eval_loading_time,
             _join_prefix(prefix, "total_time"): time_fn(),
         }
+        if (gate_hist := result.extras.get("gate_hist", None)) is not None:
```
dlwh, line 128: so i think i'm gonna have a strong preference for 

1) extracting this block (and the part in the loop) into a class (sort of like runningmean)
2) not actually checking the usage of it in taggedevaluator (or in the models) into main, but instead
3) making a little guide on how to add it in, since it's something that people want to play with sometimes but kinda adds a bunch of noise

src/levanter/models/llama.py
```diff
@@ -193,16 +203,20 @@ def init(
         if isinstance(activation_fn, str):
             activation_fn = ACT2FN[activation_fn]
         act = activation_fn  # type: ignore
-        return LlamaMlp(gate_proj, up_proj, down_proj, act)
+        get_bins()  # initialize bins
```
dlwh, line 206: rm?

src/levanter/eval.py
```diff
@@ -203,23 +216,32 @@ def accum_for_batch(
                 safe_mean = hax.where(this_tokens_per_tag, this_loss_per_tag / this_tokens_per_tag, 0.0)
                 mean_per_tag = mean_per_tag.add(safe_mean, this_tokens_per_tag)
 
-            return mean, mean_per_tag
+                if extras:
+                    for key in extras:
+                        curr = total_extras.get(key, jnp.zeros_like(extras[key]))
+                        total_extras[key] = extras[key] + curr
```
dlwh, line 222: is summing always going to be the right reduction here?

src/levanter/tracker/histograms.py
```diff
@@ -0,0 +1,68 @@
...
+    else:
+        res = histogram(a, bins)
+    return res
+
+
+NSIDE = 254
+NBINS = 2 * NSIDE + 3
+
+
+@jax.jit
```
dlwh, line 59: generally speaking it's not worth putting `jit` around helpers, though sometimes it is

src/levanter/tracker/histograms.py
```diff
@@ -0,0 +1,68 @@
...
+def get_bins() -> Array:
+    bins = jnp.logspace(-16, 6, 254, base=2.0)
+    inf = jnp.array([jnp.inf])
+    zero = jnp.array([0.0])
+    _BINS = jnp.concatenate([-inf, -bins[::-1], zero, bins, inf])
+    return _BINS
+
+
+BIN_AX = Axis("bins", NBINS - 1)
```
dlwh, line 68: rm?

src/levanter/tracker/histograms.py
```diff
@@ -0,0 +1,68 @@
...
+from jax import Array
+from haliax import Axis, NamedArray
+from haliax.partitioning import ResourceAxis
+import haliax
+from jax.sharding import PartitionSpec
+from jax.experimental.shard_map import shard_map
+
+
+@jax.jit
+def histogram(a: Array, bins: Array) -> Array:
```
dlwh, line 12: add a reference to that git issue about why we need this?

src/levanter/tracker/histograms.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    a = a.flatten()
+    prefix_sum = jnp.sum((a < bins[:, None]).astype(jnp.int32), axis=1)
+    last_count = jnp.sum(a <= bins[-1])
+    prefix_sum = prefix_sum.at[-1].set(last_count)
+    return jnp.expand_dims(jnp.diff(prefix_sum), 0)
+
+
+@jax.jit
+def sharded_histogram(a: Array, bins: Array) -> Array:
```
dlwh, line 31: let's maybe just call this histogram and the other thing _histogram?

src/levanter/tracker/histograms.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Args:
+        a (Array): The input array to compute the histogram of
+        bins (Array): The bins for the histogram
+
+    Returns:
+        Array: The resulting counts. Length is len(bins) - 1
+    """
+    P = PartitionSpec
+    in_specs = (P(ResourceAxis.DATA, None), P(None))
```
dlwh, line 42: let's use `with_sharding_constraint` to force a into the right sharding. I guess this is a good reason to stick with namedarrays huh

src/levanter/tracker/histograms.py
```diff
@@ -0,0 +1,68 @@
...
+        res = histogram(a, bins)
+    return res
+
+
+NSIDE = 254
+NBINS = 2 * NSIDE + 3
+
+
+@jax.jit
+def get_bins() -> Array:
```
dlwh, line 60: make this take a number of bins (with the current default?)

dlwh (commented):
OK so I kinda want to not make a whole bunch of changes to the model API just yet, and would rather have a guide on how to hack this in, since these things tend to be special snowflakes.

I also wonder if we  just should consider using a debug callback (see e.g. jit_log_metrics) which is a bit gross from a functional purity perspective, but for logging I think it's fine?

dlwh:
I'm gonna close this. I added a modification of your histogram code (thanks!) to #381 . the activation logging in the eval feels a bit too specific. Happy to discuss if you disagree though

PR #642, opened 2024-06-25
Add fineweb html/markdown data configs and training configs

Ivan-Zhou:


config/data/fineweb_llama_html.yaml
```diff
@@ -0,0 +1,5369 @@
+cache_dir: "gs://marin-data/tokenized/fineweb/llama2_tokenizer/html"
+tokenizer: "meta-llama/Llama-2-7b-hf"
+stop_strategy: restart
+configs:
+  "fineweb":
+    train_urls:
+      - gs://marin-data/processed/fineweb/fw-v1.0/CC-MAIN-2020-10/000_00000/{0..257}_processed_html.jsonl.gz
```
dlwh, line 7: so many urls. I'm a bit worried the tokenizer pipeline might not like this. I guess we'll find out.

scripts/get_files_on_gcs.py
```diff
@@ -0,0 +1,97 @@
+import os
```
dlwh, line 1: this script is pretty bespoke. can we move to marin?

dlwh (commented):
can we move the script to marin

config/data/fineweb_llama_html.yaml
abhinavg4: As all the files are jsonl. I can potentially just concatenate them. 

Btw there will be around 8584 lines. More info [here](https://docs.google.com/spreadsheets/d/1fS9NlGPat4r-rxjoH0qTvpoLknBQ6FVk8VWpmE_3450/edit?usp=sharing)

I can potentially merge files in each `000_0000` to a single file so it will be `gs://marin-data/processed/fineweb/fw-v1.0/CC-MAIN-2020-10/*_md.json.gz`. WIll that be better?

dlwh: yes * does work. We've been avoiding globs since it's a little less precise, but maybe it's the right thing here.

scripts/get_files_on_gcs.py
Ivan-Zhou: Moved to https://github.com/stanford-crfm/marin/pull/40

config/data/fineweb_llama_html.yaml
Ivan-Zhou: > * does work

Oh really? I didn't know that is supported.

Ivan-Zhou: > there will be around 8584 lines

Yes, I added 8584 lines to the md config

PR #644, opened 2024-06-25
Add cast to compute datatype in tagged evaluator

Aphoh:
Ran into an issue with the evaluator on TPUs when some parameters are in float32 and some are in bfloat16. I was adding LoRAs to the attention projections and there must've been some odd casting going on, but it would work on CPU and not on TPU 🤷. Adding a cast to the compute datatype in the jitted accumulator of the evaluator fixed the issue.

### DIFF
--- a/src/levanter/eval.py
+++ b/src/levanter/eval.py
@@ -172,7 +172,7 @@ class TaggedEvaluator:
         device_mesh=None,
         axis_mapping=None,
         max_examples_per_dataset=None,
-        mp: jmp.Policy = None,
+        mp: Optional[jmp.Policy] = None,
     ):
         self.EvalBatch = EvalBatch
         self.dataset = DomainTaggedDataset(tagged_eval_sets, max_examples_per_dataset)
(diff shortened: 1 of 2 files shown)


src/levanter/eval.py
```diff
@@ -162,13 +166,20 @@ class TaggedEvaluator:
...
 
     def __init__(
-        self, EvalBatch: hax.Axis, tagged_eval_sets, device_mesh=None, axis_mapping=None, max_examples_per_dataset=None
+        self,
+        EvalBatch: hax.Axis,
+        tagged_eval_sets,
+        device_mesh=None,
+        axis_mapping=None,
+        max_examples_per_dataset=None,
+        mp: jmp.Policy = None,
```
dlwh, line 175:
```suggestion
        mp: Optional[jmp.Policy] = None,
```

src/levanter/eval.py
```diff
@@ -188,6 +199,7 @@ def accum_for_batch(
             m: LmHeadModel, state: tuple[RunningMean, RunningMean], batch: LmExample, tags: hax.NamedArray
         ):
             m = inference_mode(m, True)
+            m = self.mp.cast_to_compute(m)
```
dlwh, line 202:
```suggestion

            if self.mp is not None:
                m = self.mp.cast_to_compute(m)
```

dlwh (commented):
this is weird to me but thank you

PR #645, opened 2024-06-25
Fix eval-lm when hf_checkpoint is passed

Aphoh:
The codepath of loading a HF checkpoint was being skipped in the tests for `eval_lm`. Added a test case and fixed the bug that was causing loading to fail.

### DIFF
--- a/src/levanter/main/eval_lm.py
+++ b/src/levanter/main/eval_lm.py
@@ -99,7 +99,7 @@ def main(config: EvalLmConfig):
             model_config = config.model
             if not hasattr(model_config, "hf_checkpoint_converter"):
                 raise ValueError("Model config does not have an HF checkpoint converter. Can't load HF checkpoint.")
-            converter: HFCheckpointConverter = model_config.hf_checkpoint_converter
+            converter: HFCheckpointConverter = model_config.hf_checkpoint_converter()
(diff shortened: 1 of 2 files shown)


src/levanter/main/eval_lm.py
```diff
@@ -99,7 +99,7 @@ def compute_loss(model: LmHeadModel, example: LmExample):
             model_config = config.model
             if not hasattr(model_config, "hf_checkpoint_converter"):
                 raise ValueError("Model config does not have an HF checkpoint converter. Can't load HF checkpoint.")
-            converter: HFCheckpointConverter = model_config.hf_checkpoint_converter
+            converter: HFCheckpointConverter = model_config.hf_checkpoint_converter()
```
dlwh, line 102: my kingdom for a decent static type checker

tests/test_eval_lm.py
```diff
@@ -56,3 +57,48 @@ def test_eval_lm():
                 os.unlink("wandb")
             except Exception:
                 pass
+
+
+@pytest.mark.entry
+@skip_if_no_torch
+def test_eval_lm_from_hf():
```
dlwh, line 64: thanks for adding this

dlwh (approved):

PR #647, opened 2024-06-28
add support for different capacity types to launch script

dlwh:
we have some reserved capacity right now. i don't love the argparser shenanigans, but it's close to how the official scripts do it


rjpower (approved):
LGTM, yeah a little odd but seems manageable...

PR #649, opened 2024-07-02
Add rope theta to llama config, test with llama 3

Aphoh:
Llama3 currently gives bad results in levanter, due to it using a different value for the RoPE base/theta. Added a config and a test here. Finetuning Llama3 I _believe_ is still bad, but I'm fairly sure this is a data formatting issue.


dlwh (approved):
thanks! we still need to get to the bottom of llama3.

PR #653, opened 2024-07-05
Add option to use ghcr in launch.py

Aphoh:
Adds the option for pushing containers to GHCR instead of GCP container registry into launch.py and adds a note in the TPU docs for it.

### DIFF
--- a/docs/Getting-Started-TPU-VM.md
+++ b/docs/Getting-Started-TPU-VM.md
@@ -117,6 +117,14 @@ EOF
 
 If you want to customize the docker image that is created and uploaded to GCP Artifact Registry, you can add config `image_name: "YOUR-DOCKER-NAME"`.
 
+Note that you can also configure docker to push to GHCR by setting
+```
+docker_registry: ghcr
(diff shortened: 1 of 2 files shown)


dlwh (approved):
lgtm but @rjpower could you weigh in?

rjpower:
Seems fine, but what's the motivation? Does GCP not work for some use cases?

(The reason for using GCP was the concern that GHCR/Docker would be slow to load for larger pods or throttle/block the connection if e.g. they get 1000 requests at once.)

dlwh:
(I'm guessing the reason is price sensitivity. A lot of TRC users are independent researchers.)

@Aphoh if you could fix merge conflicts (sorry about that!) i'll merge

Aphoh:
@dlwh Nailed it, I'm as cheap as it gets when it comes to cloud stuff 😅. Merge conflicts fixed.

dlwh:
i mean, i get it. I freaked out when some stray google process was costing us ~$500 month due to logging, and I have (some) grant money.

dlwh:
thanks!

Issue #652, opened 2024-07-04
Change launch.py to not rebuild docker container on TPU node failure (with retries)


dlwh:
it should stick with an image once it's built it.

(cc @rjpower )

PR #655, opened 2024-07-06
Don't rebuild Docker images when retrying.

rjpower:
Fixes issue #652 (I think).

I just moved the image build out of the retry loop; let me know if you were looking for something smarter!


dlwh (approved):
Thanks! cc @Ivan-Zhou

PR #660, opened 2024-07-11
Fix typo in GLU MLP calculation

Aphoh:
This was a big dumb on my part. 
* Our `calculated_flops / (2*n_param)` should be _greater_ than 1.0 (about 1.13) since we take into account more than just the projections.
* I had a typo that multiplied the MLP flops by 2 instead of 3 if the model is a GLU. My bad!


dlwh (approved):
nice catch thanks!

PR #663, opened 2024-07-18
Adding configs related to DCLM

abhinavg4:
DCLM 7B related configs


Ivan-Zhou:
LGTM. Can you run pre-commit checks to fix the pre-commit issue? I think it is just some formatting issue. 

```
pre-commit run --all-files
```

src/levanter/models/llama.py
```diff
@@ -64,6 +65,7 @@ class LlamaConfig(HFCompatConfig):
     activation_function: str = "silu"
     initializer_range: float = 0.02
     layer_norm_epsilon: float = 1e-5
+    z_loss_weight: float = 0.0
```
dlwh, line 68: i would rather this not be a property of the model's config but an option on TrainLmConfig, and define a loss function in train_lm and pass it into trainer

src/levanter/models/lm_model.py
```diff
@@ -131,9 +131,12 @@ def compute_loss(
         logits = logits.astype(jnp.float32)
         targets = hax.roll(example.tokens, -1, axis=self.Pos.name)
         target_y = hax.nn.one_hot(targets, self.Vocab, dtype=logits.dtype)
-        loss = cross_entropy_loss(
-            logits, self.Vocab, target_y, reduction, reduction_axis=reduction_axis, where=example.loss_mask
-        )
+        if hasattr(self.config, "z_loss_weight") and self.config.z_loss_weight > 0:
```
dlwh, line 134: really don't like using this here. much cleaner to just pull out the loss function

src/levanter/main/train_lm.py
```diff
@@ -19,12 +19,29 @@
...
 from levanter.optim import AdamConfig, OptimizerConfig
 from levanter.trainer import Trainer, TrainerConfig
+from levanter.types import ComputeLossFunction, M, X
 from levanter.utils.jax_utils import parameter_count
 
 
 logger = logging.getLogger(__name__)
 
 
+class ModuleComputeZLoss(ComputeLossFunction[M, X]):
```
dlwh, line 29: i still don't like this but I think I can't really articulate what I want. i'm gonna push a change to my fork

Issue #629, opened 2024-06-12
Launch script multislice


blahBlahhhJ:
Things I'm going to add to `launch.py`

1. Migrate some commands API from `tpu-vm` to `queued-resources` to work better with multislice training
2. Delete the instance if command fail (regardless of autodelete) so that it won't eat up all the quota after many retries

blahBlahhhJ:
@dlwh @rjpower  FYI

rjpower:
For (1) SGTM, let me know if I can help. Something simple that just runs `queued-resources create` and then polls until the instance is started seems like a good first step.
For (2) don't we normally want to reuse the instance? That's much faster for me in practice than delete/restart. (Or is this for batch job, in which case, sure, makes sense as an option).

blahBlahhhJ:
In my experience when an instance is preempted (suspended), then you cannot create another one with the same name. The suspended instance is going to stay there and take up the quota.

dlwh:
I think you all are taking about slightly different things. If the node is
still up we want to reuse. Otherwise we have to delete the node and queued
resource and recreate

(Russell this is one of the main reasons i do not like queued resources…)

blahBlahhhJ:
@rjpower is there a way to specify which branch to run?

rjpower:
Not specifically. The script packages whatever is in your current
directory, so you can just checkout the branch you want to run.

PR #666, opened 2024-07-21
Enable multislice in launch script

blahBlahhhJ:
Adds `node_count` in config to launch jobs with multislice training.

Refactored the launch script to use `queued-resources` API.

mention #629

### DIFF
--- a/infra/helpers/cli.py
+++ b/infra/helpers/cli.py
@@ -12,6 +12,11 @@ def run_command(*args, **kwargs):
     return subprocess.check_call(args, **kwargs)
 
 
+def run_command_parallel(*args, **kwargs):
+    print("Running:", " ".join(list(args)))
+    return subprocess.Popen(args, **kwargs)
+
+
 def add_ssh_key(ssh_key_filename):
     # format 3072 SHA256:... key-name (RSA)
     key_hash = subprocess.check_output(["ssh-keygen", "-lf", ssh_key_filename]).decode("utf-8").split()[1]
@@ -23,9 +28,12 @@ def add_ssh_key(ssh_key_filename):
     subprocess.check_call(["ssh-add", ssh_key_filename])
 
 
-def tpu_ssh(tpu_name, zone, *args, ignore_failure=False):
+def tpu_ssh(tpu_name, zone, node_count, *args, ignore_failure=False):
     add_ssh_key(os.path.expanduser("~/.ssh/google_compute_engine"))
     try:
+        if node_count > 1:
+            return _tpu_ssh_multislice(tpu_name, zone, node_count, *args, ignore_failure)
+
         return run_command(
             "gcloud",
             "alpha",
@@ -43,6 +51,28 @@ def tpu_ssh(tpu_name, zone, *args, ignore_failure=False):
             print("Ignoring failure:", e)
         else:
             raise
+
+
+def _tpu_ssh_multislice(tpu_name, zone, node_count, *args, ignore_failure=False):
+    processes = []
+    for i in range(node_count):
+        p = run_command_parallel(
+            "gcloud",
+            "alpha",
+            "compute",
+            "tpus",
+            "queued-resources",
+            "ssh",
+            tpu_name,
+            "--worker=all",
+            f"--node={i}",
+            f"--zone={zone}",
+            "--command=%s" % " ".join(args),
+        )
+        processes.append(p)
+    for p in processes:
+        if p.wait() != 0:
+            raise subprocess.CalledProcessError(p.returncode)
 
 
 # Oddly enough, there's no API to simply fetch the current gcloud configuration...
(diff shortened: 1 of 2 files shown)


blahBlahhhJ:
Some UX design discussions: what behavior do we want our launch script to have?
1. * Before: with retry>0, the launch script will retry even if the job finishes without error
    *  After: the launch script will exit (no retry) if the job finishes without error
2. * Before: (not using `queued-resources` API, not sure about the actual behavior)
    * After: If `autodelete` is True, **and the job finishes with no error,** the launch script will tear down the TPU automatically (so that it doesn't eat up the TPU quota). Do we want this to happen when the launch script finishes all the retries as well? e.g. tear down the TPU after the job errors and retries 100 times.
 3. If the user manually kills the script (e.g. by ctrl-c), the TPU will still stay in the `queued-resources` list, taking up some quota even after suspended. Is there something we can do here?

infra/launch.py
```diff
@@ -84,16 +107,16 @@ def start_tpu_vm(tpu_name, *, tpu_type, capacity_type, version, zone, autodelete
...
         f"--accelerator-type={tpu_type}",
-        f"--version={version}",
-        "--zone=" + zone,
+        f"--runtime-version={version}",
+        f"--zone={zone}",
         "--quiet",
     ]
     if capacity_type == "preemptible":
-        command.append("--preemptible")
+        command.append("--best-effort")
```
dlwh, line 119: let's add best-effort as a synonym for preemptible in the argparse

infra/launch.py
```diff
@@ -102,8 +125,33 @@ def start_tpu_vm(tpu_name, *, tpu_type, capacity_type, version, zone, autodelete
...
+        command.append(f"--node-id={tpu_name}")
+    else:
+        command.append(f"--node-count={node_count}")
+
     cli.run_command(*command)
 
+    # wait for queued resource to complete
+    waited = 0
+    while True:
+        time.sleep(60)
```
dlwh, line 139: before sleeping this long we should log that we're gonna wait.

infra/launch.py
```diff
@@ -102,8 +125,33 @@ def start_tpu_vm(tpu_name, *, tpu_type, capacity_type, version, zone, autodelete
         pass
     else:
         raise ValueError(f"Unknown capacity type: {capacity_type}")
+
+    if node_count == 1:
+        command.append(f"--node-id={tpu_name}")
+    else:
+        command.append(f"--node-count={node_count}")
```
dlwh, line 132: set `node-prefix`  to tpu_name?

infra/helpers/cli.py
```diff
@@ -12,6 +12,11 @@ def run_command(*args, **kwargs):
     return subprocess.check_call(args, **kwargs)
 
 
+def run_command_parallel(*args, **kwargs):
```
dlwh, line 15: not really parallel but background yeah?

dlwh (commented):
let's wait on @rjpower (maybe timing out after a couple of days) but other than comments lgtm

dlwh:
To answer your questions:

1. new behavior is better
2. `Use the existing argument autodelete` is better. Existing behavior is kinda weird imho
3. I think the current behavior is the right behavior, especially with TRC TPUs. Maybe we should add a cleanup script or something but I think it's fine.

infra/launch.py
blahBlahhhJ: The tpu name will automatically become the node prefix.

infra/launch.py
```diff
@@ -38,44 +40,65 @@ def setup_vm_docker(tpu_name, zone, docker_base_image):
...
-        [
-            "gcloud",
-            "alpha",
-            "compute",
-            "tpus",
-            "tpu-vm",
-            "list",
-            "--zone=" + zone,
-        ]
+    return json.loads(
```
rjpower, line 43: Does this also show non-queued-resources TPUs? (Or should we always use queued-resources?)

rjpower (approved):
Seems good to me, thanks for doing this!

It seems like this will only work with queued resources now, but @dlwh would know if there were any real downsides of that. Adjusting things to merge both types of TPUs seems tedious and probably not worth it, assuming queued-resources works fine for smaller slices.

infra/launch.py
dlwh: but you're not appending the tpu name?

dlwh: oh i see because it's the queued resource name.

dlwh (approved):

infra/launch.py
blahBlahhhJ: after the migration everything will be in queued resources so we can probably just clean up all the non queued resources tpus and move on.

blahBlahhhJ:
(waiting for a preemption event to see if my script retries correctly and push a final commit addressing @dlwh 's comments before I merge)

dlwh:
will just deleting the TPU (not the resource) simulate preemption well enough?

dlwh:
@blahBlahhhJ is this ready to merge?

PR #672, opened 2024-07-25
refactor trainer to always need a loss function, add z_loss

dlwh:


dlwh:
Tagging @rjpower @Helw150 for visibilit

rjpower (approved):
Nice, thanks!

PR #675, opened 2024-07-25
Support Eleuther LM-Eval-Harness in Levanter

### LINKED ISSUE #564: wire up llm-eval-harness
CF https://github.com/kingoflolz/mesh-transformer-jax/blob/master/eval_harness.py

dlwh:
Adds Eleuther's LM Eval Harness as a callback in Levanter. It's much slower than it needs to be because I'm not doing any sequence packing, but it gets the job done. Scores on Llama 3 seem reasonable, so I think this is right.

Closes #564


config/olmo/olmo_7b_repro.yaml
```diff
@@ -0,0 +1,175 @@
+#data: !include data/dolma_olmo_paloma.yaml
+data:
```
percyliang, line 2: What does this have to do with lm evaluation harness?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
+# Code for running https://github.com/EleutherAI/lm-evaluation-harness inside Levanter runs
```
percyliang, line 1: Turn this into a triple-quoted string with more top-level context?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+        return True
+
+    async def current_len(self) -> Optional[int]:
+        return len(self.examples)
+
+    async def get_batch(self, indices: Sequence[int]) -> List[LmExample]:
+        out = []
+        pad_token_id = self.tokenizer.pad_token_id
+
+        reqs = [(self.examples[i].args[0], self.examples[i].args[1]) for i in indices]
```
percyliang, line 77: Comment what assumptions are made about the structure of the examples (args[0], args[1])?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+            context_enc_len = len(context_enc["input_ids"])
+
+            tokens, length = self._truncate_or_pad(whole_enc, context_enc_len)
+            example = _jit_create_example(self.Pos, tokens, length, pad_token_id)
+
+            out.append(example)
+
+        return out
+
+    def _truncate_or_pad(self, encoded, prompt_length):
```
percyliang, line 92: add type hints

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+
+    async def get_batch(self, indices: Sequence[int]) -> List[LmExample]:
+        out = []
+        pad_token_id = self.tokenizer.pad_token_id
+
+        reqs = [(self.examples[i].args[0], self.examples[i].args[1]) for i in indices]
+
+        for context, completion in reqs:
+            whole_enc = self.tokenizer(context + completion)
+            context_enc = self.tokenizer(context)
```
percyliang, line 81: Must we run tokenizer twice?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+class LevanterHarnessLM(LM):
+    def __init__(self, EvalBatch: hax.Axis, EvalPos: hax.Axis, model: LmHeadModel, axis_resources, tokenizer):
+        super().__init__()
+        self.EvalBatch = EvalBatch
+        self.EvalPos = EvalPos
+        self.model = model
+        self.axis_resources = axis_resources
+        self.tokenizer = tokenizer
+
+        def _eval_loglikelihood(model: LmHeadModel, example: LmExample):
```
percyliang, line 123: type hint for return type

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+    nb that LM Eval Harness has its own TaskConfig, but its defaults are not the same as just passing in
+    a dict, and we want the behavior of passing in a dict.
+
+    See Also:
+        [LM Eval Harness TaskConfig](https://github.com/EleutherAI/lm-evaluation-harness/blob/0ef7548d7c3f01108e7c12900a5e5eb4b4a668f7/lm_eval/api/task.py#L55)
+    """
+
+    task: str
+    task_alias: str | None = None
+    num_fewshot: int | None = None
```
percyliang, line 210: What does None represent that's not representable by an integer?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+    Harness expects.
+
+    nb that LM Eval Harness has its own TaskConfig, but its defaults are not the same as just passing in
+    a dict, and we want the behavior of passing in a dict.
+
+    See Also:
+        [LM Eval Harness TaskConfig](https://github.com/EleutherAI/lm-evaluation-harness/blob/0ef7548d7c3f01108e7c12900a5e5eb4b4a668f7/lm_eval/api/task.py#L55)
+    """
+
+    task: str
```
percyliang, line 208: Can we add docs for each field?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+@dataclass(frozen=True)
+class LmEvalHarnessConfig:
+    task_spec: list[TaskConfig | str] | None = None
+    max_examples: int | None = None
+    max_eval_length: int | None = None
+    log_samples: bool = False
+
+    def task_spec_or_default(self) -> list[str | dict]:
+        if self.task_spec is None:
+            return ["hellaswag", "piqa"]
```
percyliang, line 231: Huh, why are these special?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+        return [task.to_dict() if isinstance(task, TaskConfig) else task for task in self.task_spec]
+
+    def to_task_dict(self) -> dict:
+        import lm_eval.tasks as tasks
```
percyliang, line 235: Add docstring on what this function is doing / what it's for?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+                logger.exception(f"Failed to load task {task}")
+                raise ValueError(f"Failed to load task {task}")
+        return this_tasks
+
+
+@dataclass(frozen=True)
+class EvalHarnessMainConfig:
+    tokenizer: str
+    checkpoint_path: str
+    checkpoint_is_hf: bool = False
```
percyliang, line 262: document what the format is if not hf?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+    return outputs
+
+
+def _actually_run_eval_harness(config: LmEvalHarnessConfig, model, tasks_to_run, tokenizer, EvalBatch, axis_resources):
+    max_examples = config.max_examples
+    max_eval_length = config.max_eval_length
+
+    EvalPos = model.Pos if max_eval_length is None else model.Pos.resize(max_eval_length)
+    harness = LevanterHarnessLM(EvalBatch, EvalPos, model, axis_resources, tokenizer)
+    # we always log_samples here and filter out the samples later if we don't want them
```
percyliang, line 298: log_samples is a verb?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+        metric_values, this_examples_per_task = zip(*valid_tasks)
+
+        # Compute macro and micro averages
+        averages["macro_avg_" + metric] = np.mean(metric_values)
+        averages["micro_avg_" + metric] = np.average(metric_values, weights=this_examples_per_task)
+
+    return averages
+
+
+NAT_TO_BIT = 1 / np.log(2)
```
percyliang, line 353: BITS_PER_NAT?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+
+# eval_harness isn't consistent enough for this to actually be workable
```
percyliang, line 355: What does this mean?  lm evaluation harness is just a framework, so whether it's meaningful or not depends on the actual eval and the model size?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+    model: LmConfig = dataclasses.field(default_factory=Gpt2Config)
+
+    eval_harness: LmEvalHarnessConfig = dataclasses.field(default_factory=LmEvalHarnessConfig)
+
+    @property
+    def EvalBatch(self):
+        return self.trainer.EvalBatch
+
+    @cached_property
+    def the_tokenizer(self):
```
percyliang, line 273: not obvious what this function is doing from the name...maybe `tokenizer_object` or something?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+                to_log[f"{prefix}/averages/{metric_name}"] = metric_value
+
+    tracker.log(to_log, step=None)
+
+
+def lm_eval_harness(config: LmEvalHarnessConfig, tokenizer, EvalBatch, axis_resources):
+    tasks_to_run = config.to_task_dict()
+
+    def lm_eval_harness(step: StepInfo, force=False):
+        # if step.step == 0 and not force:
```
percyliang, line 533: Remove comment?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+            config.eval_harness,
+            model,
+            tokenizer,
+            config.EvalBatch,
+            axis_resources=compute_axis_mapping,
+        )
+
+        logger.info("Finished running LM eval harness")
+        # log the results as json
+        with open("lm_eval_results.json", "w") as f:
```
percyliang, line 491:
```suggestion
        with open("lm_eval_harness_results.json", "w") as f:
```

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+            json.dump(outputs, f, indent=2)
+
+        # also write to stdout
+        if jax.process_index() == 0:
+            print(json.dumps(outputs, indent=2), flush=True)
+
+        # also log the results
+        levanter.tracker.current_tracker().log_artifact("lm_eval_results.json")
```
percyliang, line 499: update the filename (I'd like lm_eval_harness to be a collocation that we use consistently)

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+
+
+def log_report_to_tracker(prefix: str, report: dict, tracker: Optional[levanter.tracker.Tracker] = None):
+    if tracker is None:
+        tracker = levanter.tracker.current_tracker()
+
+    to_log = {}
+    for task_name, task_results in report["results"].items():
+        for metric_name, metric_value in task_results.items():
+            if metric_name.endswith(",none"):
```
percyliang, line 512: What is this hackery?  put assumptions in comment

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+            config,
+            model,
+            tasks_to_run,
+            tokenizer,
+            EvalBatch,
+            axis_resources,
+        )
+
+        if jax.process_index() == 0:
+            with tempfile.NamedTemporaryFile("w", delete=False, suffix=".json") as f:
```
percyliang, line 549: tmp file that doesn't get deleted doesn't sound good - I'd delete or put the file in the evaluation directory (which I guess would have to be passed in as an explicit location)

src/levanter/models/lm_model.py
```diff
@@ -47,6 +48,33 @@ def causal(
...
+    @staticmethod
+    def from_prompt_and_completion(
+        Pos,
+        tokens: hax.NamedArray,
+        prompt_length: NamedOrNumeric,
+        *,
+        ignore_id: Optional[int] = None,
+        all_causal: bool = True,
+    ) -> "LmExample":
+        # mask out the prompt tokens
```
percyliang, line 60: docstring

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+            try:
+                if isinstance(task, str):
+                    this_tasks.update(tasks.get_task_dict(task, manager))
+                else:
+                    our_name = task.get("task_alias", task["task"]) if isinstance(task, dict) else task
+                    our_name = our_name.replace(" ", "_")
+                    task_dict = tasks.get_task_dict([task], manager)
+                    this_task = task_dict.popitem()[1]
+                    # hacky, but this allows us to run multiple instances of the same task with different fewshot settings
+                    this_task.config.task = our_name
```
nikil-ravi, line 250: I think this works well since the result file is saved as `task.jsonl` so using the alias helps us distinguish the number of shots. It might be worth it to move this logic into a helper function `create_task_with_unique_name()` for readability...

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+
+    EvalPos = model.Pos if max_eval_length is None else model.Pos.resize(max_eval_length)
+    harness = LevanterHarnessLM(EvalBatch, EvalPos, model, axis_resources, tokenizer)
+    # we always log_samples here and filter out the samples later if we don't want them
+    outputs = evaluator.evaluate(harness, tasks_to_run, limit=max_examples, log_samples=True)
+
+    averages = _compute_averages(outputs)
+    outputs["averages"] = averages
+
+    if not config.log_samples:
```
nikil-ravi, line 304: Why not just pass in `log_samples=config.log_samples` above?

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+) -> dict:
+    # tasks_to_run = tasks.get_task_dict(config.task_spec_or_default(), tasks.TaskManager())
+    tasks_to_run = config.to_task_dict()
+
+    outputs = _actually_run_eval_harness(config, model, tasks_to_run, tokenizer, EvalBatch, axis_resources)
+
+    return outputs
+
+
+def _actually_run_eval_harness(config: LmEvalHarnessConfig, model, tasks_to_run, tokenizer, EvalBatch, axis_resources):
```
nikil-ravi, line 292: docstring for outputs would be useful here

nikil-ravi (approved):

src/levanter/eval_harness.py
```diff
@@ -0,0 +1,564 @@
...
+    averages = _compute_averages(outputs)
+    outputs["averages"] = averages
+
+    if not config.log_samples:
+        del outputs["samples"]
+
+    return outputs
+
+
+def _compute_averages(outputs):
```
nikil-ravi, line 310: Unrelated to this PR specifically but I think we could add more ways to aggregate results- the DCLM paper for example reports centered accuracy. Maybe we can have this aggregation function be something we pass intoLmEvalHarnessConfig. Or something like subtract_random_baseline: true in the YAML config...

src/levanter/eval_harness.py
dlwh: it's how it's done in lm harness and in side alpaca. it's the easiest thing and not a bottleneck

src/levanter/eval_harness.py
dlwh: i just made a default, i'll remove

src/levanter/eval_harness.py
dlwh: none means the default for the task, whatever it is

src/levanter/eval_harness.py
dlwh: i mean, they don't even have docs for a lot of these.

src/levanter/eval_harness.py
dlwh: good idea!

src/levanter/eval_harness.py
dlwh: i use this convention throughout levanter...

src/levanter/eval_harness.py
dlwh: also, we don't need this behavior actually because i can't do the metrics I want to do with the samples anyhow

src/levanter/eval_harness.py
dlwh: imho we should modify lm harness to do that, but it's a great idea

src/levanter/eval_harness.py
dlwh: oh i lied, we do

src/levanter/eval_harness.py
dlwh: well like, the samples for multiple choice tasks don't all have the answer in a standard format even though they easily could (and must internall at some point)

src/levanter/eval_harness.py
dlwh: they just add `,none` to all the default metrics for some reason. e.g. it's `acc,none`, `acc_norm,none` etc etc

src/levanter/eval_harness.py
dlwh: it's delete=False so that the wandb process still can upload it

percyliang (approved):

PR #679, opened 2024-07-30
[WIP] Implements Roberta Model

JulienDarve:


src/levanter/models/testing.ipynb
dlwh, line 1: rm before merge

src/levanter/models/roberta.py
```diff
@@ -0,0 +1,852 @@
...
+    cos: NamedArray,  # [position, head_size]
+    sin: NamedArray,  # [position, head_size]
+) -> Tuple[NamedArray, NamedArray]:
+    """Applies rotary position embedding to q and k."""
+    q_embed = q * cos + _rotate_half(q) * sin
+    k_embed = k * cos + _rotate_half(k) * sin
+    return q_embed, k_embed
+
+
+def llama_rotary_pos_emb(
```
dlwh, line 836: you'll want to delete this i think? (if you end up needing it, just import from llama.py)

src/levanter/models/roberta.py
```diff
@@ -0,0 +1,852 @@
...
+
+        return RobertaForMaskedLM(roberta, lm_head, Vocab)
+
+    def get_output_embeddings(self):
+        return self.lm_head.decoder
+
+    def set_output_embeddings(self, new_embeddings):
+        self.lm_head.decoder = new_embeddings
+
+    def forward(
```
dlwh, line 778: in general we use __call__ and not forward in Levanter

src/levanter/models/roberta.py
```diff
@@ -0,0 +1,852 @@
...
+            position_ids=position_ids,
+            inputs_embeds=inputs_embeds
+        )
+
+        sequence_output = outputs[0]
+        prediction_scores = self.lm_head(sequence_output)
+
+        masked_lm_loss = None
+        if labels is not None:
+            masked_lm_loss = hnn.cross_entropy_loss(logits=prediction_scores, Label=self.Vocab, targets=labels)
```
dlwh, line 809: my preference is to put this part in a method called compute_loss that calls `__call__` (which computes the logits), and then computes the loss


also you'll want to use the loss mask when that gets merged here

dlwh (commented):
didn't check correctness. just some style/organization thoughts

dlwh:
I should add: looking good! I know this ended up being a big lift and I appreciate you all taking it on!

src/levanter/models/roberta.py
```diff
@@ -0,0 +1,852 @@
...
+
+        context_layer = hax.dot(attention_probs, value_layer, axis=self.KeyPos)
+        
+        outputs = hax.rearrange(context_layer, ("... heads position head_size -> ... position (embed_att: heads head_size)"), heads=self.Heads, head_size=self.HeadSize)
+
+        return outputs
+
+class RobertaSelfOutput(eqx.Module, StateDictSerializationMixin):
+    dense: hnn.Linear
+    LayerNorm: hnn.LayerNorm
```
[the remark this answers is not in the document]
JulienDarve: It is named this in the transformers library for most of the functions, so I made it that way here. I originally had it in snake case.

src/levanter/models/roberta.py
```diff
@@ -0,0 +1,852 @@
...
+        return {"self_attn": "self"}
+
+    @named_call
+    def __call__(
+        self,
+        hidden_states: NamedArray,
+        attention_mask: Optional[NamedArray] = None,
+        *,
+        key
+    ) -> Tuple[NamedArray]:
```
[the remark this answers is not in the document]
JulienDarve: Good point

src/levanter/models/testing.ipynb
JulienDarve: Will do. Do you want me to add it somewhere else in levanter or just keep it to myself?

src/levanter/models/roberta.py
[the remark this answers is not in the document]
dlwh: totally fine then! probably best

src/levanter/models/testing.ipynb
dlwh: you can just move it into examples/roberta/ or something. Ideally it was be proper unit tests

src/levanter/models/roberta.py
```diff
@@ -0,0 +1,852 @@
...
+    @named_call
+    def __call__(
+        self,
+        hidden_states: NamedArray,
+        attention_mask: Optional[NamedArray] = None,
+        *,
+        key = None
+    ) -> Tuple[NamedArray]:
+
+        query_layer = self.transpose_for_scores(self.q_proj(hidden_states))
```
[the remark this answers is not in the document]
dlwh: parameters are best declared in the same order as they are in HF (though that can be worked around) but intermediate values like query_layer can be in any order really. Haliax will automatically transpose things as needed.

versae:
With the release of ModernBERT, I was wondering if there are any plans to merge this RoBERTa into levanter and possibly improve it add some nice things from ModernBERT like RoPE.

dlwh:
@versae Yeah I think that's basically the direction they're/we're heading. Basically just use Llama with a different loss function/acausal attention

PR #680, opened 2024-07-30
[WIP] MLM Training Objective

prady-saligram:
Introduces train_mlm.py, a new file adapted from train_lm.py, to support masked language modeling with dynamic masking as utilized in RoBERTa. A new class, MaskedLMDataset, has been implemented in text.py to handle dynamic masking. This class is instantiated and utilized within train_mlm.py, preserving all structural and sharding-related comments from the original train_lm.py to maintain clarity and continuity. The integration of MaskedLMDataset with the training script has been verified with appropriate parameters to ensure consistency with existing training workflows.

### DIFF
--- a/src/levanter/data/text.py
+++ b/src/levanter/data/text.py
@@ -68,10 +68,11 @@ class MaskedLmDataset(ShardableDataset[MaskedLmExample]):
         dataset: ShardableDataset[np.ndarray],
         QPos: Axis,
         KPos: Axis,
+        mask_token_id: int,
         mask_prob: float = 0.15,
         noise_prob: float = 0.1,
         key: Optional[PRNGKeyArray] = None,
-        ignore_index: Optional[int] = DEFAULT_IGNORE_INDEX,
+        # ignore_index: Optional[int] = DEFAULT_IGNORE_INDEX,
     ):
         self.dataset = dataset
         self.QPos = QPos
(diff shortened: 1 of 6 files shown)


src/levanter/data/text.py
```diff
@@ -64,6 +65,65 @@
...
+    def __iter__(self) -> Iterator[LmExample]:
+        key = self.key
+        sharding = jax.sharding.SingleDeviceSharding(jax.local_devices(backend="cpu")[0])
+
+        with use_cpu_device():
+            @functools.partial(eqx.filter_jit, out_shardings=sharding)
+            def _create_mlm_example(tokens, key):
+                tokens_array = tokens.array
+                
+                example = LmExample.causal(tokens=tokens, ignore_id=self.ignore_id)
```
dlwh, line 102: you need a non-causal attention mask for Roberta, and you need to set a loss_mask to be only the masked tokens

src/levanter/data/text.py
```diff
@@ -64,6 +65,65 @@
...
+        QPos: Axis,
+        KPos: Axis,
+        mask_prob: float = 0.15,
+        key: Optional[PRNGKeyArray] = None,
+        ignore_index: Optional[int] = DEFAULT_IGNORE_INDEX,
+    ):
+        self.dataset = dataset
+        self.QPos = QPos
+        self.KPos = KPos
+        self.mask_prob = mask_prob
```
dlwh, line 81: should there be a mask_prob and a noise_prob?

src/levanter/data/text.py
```diff
@@ -120,6 +180,8 @@ def _create_lm_example(tokens, key):
                 yield example
 
 
+
+
```
dlwh, line 184:
```suggestion
```

src/levanter/data/text.py
dlwh: you also can't use the current LmExample actually because you need a separate targets field (with the non-masked tokens). With more work you could avoid the need for targets (with just masked tokens), but probably better to add an `targets: Optional[NamedArray]` to the class (or make your own class)

PR #682, opened 2024-08-04
Specify node_count as int in launch.py

Ivan-Zhou:
Otherwise, it will throw an error at `if node_count > 1:` in `tpu_ssh()` in cli.py.


dlwh (approved):

blahBlahhhJ (approved):
LGTM

PR #685, opened 2024-08-07
Adding multiple configs

abhinavg4:
Adding the ability to take multiple configs

### DIFF
--- a/src/levanter/config.py
+++ b/src/levanter/config.py
@@ -104,7 +104,7 @@ def _maybe_get_config_path_and_cmdline_args(args: List[str]):
             raise ValueError(f"Multiple config args found in {args}")
         config_path_index = found_indices[0] + 1
 
-        config_paths = []
+        config_paths: List[str] = []
         args = args.copy()
         del args[config_path_index - 1]
         config_path_index -= 1
--- a/src/levanter/config.py
+++ b/src/levanter/config.py
@@ -93,26 +93,51 @@ def _maybe_get_config_path_and_cmdline_args(args: List[str]):
     If URL, we need to download it and save it to a temp file. We then want to remove --config_path
     from the cmdline args so that draccus doesn't try to load it as a config path and return it separately here
     along with the modified cmdline args.
+    We also accept ... --configs <config1> <config2> ... and concatenate them into a single config.
     """
-    if "--config_path" not in args and "--config" not in args:
+    if "--config_path" not in args and "--config" not in args and "--configs" not in args:
         return None, args
     else:
-        try:
-            config_path_index = args.index("--config_path")
-        except ValueError:
-            config_path_index = args.index("--config")
+        config_args = ["--config_path", "--config", "--configs"]
+        found_indices = [args.index(arg) for arg in config_args if arg in args]
+        if len(found_indices) > 1:
+            raise ValueError(f"Multiple config args found in {args}")
+        config_path_index = found_indices[0] + 1
 
-        config_path = args[config_path_index + 1]
-
-        if urllib.parse.urlparse(config_path).scheme:
-            fs: AbstractFileSystem
-            fs, fs_path = fsspec.core.url_to_fs(config_path)
-            temp_file = tempfile.NamedTemporaryFile(prefix="config", suffix=".yaml", delete=False)
-            atexit.register(lambda: os.unlink(temp_file.name))
-            fs.get(fs_path, temp_file.name)
-            config_path = temp_file.name
-
+        config_paths = []
         args = args.copy()
-        del args[config_path_index]
-        del args[config_path_index]
-        return config_path, args
+        del args[config_path_index - 1]
+        config_path_index -= 1
+
+        while config_path_index < len(args) and not args[config_path_index].startswith("-"):
+
+            config_path = args[config_path_index]
+
+            if urllib.parse.urlparse(config_path).scheme:
+                fs: AbstractFileSystem
+                fs, fs_path = fsspec.core.url_to_fs(config_path)
+                temp_file = tempfile.NamedTemporaryFile(prefix="config", suffix=".yaml", delete=False)
+                atexit.register(lambda: os.unlink(temp_file.name))
+                fs.get(fs_path, temp_file.name)
+                config_path = temp_file.name
+
+            config_paths.append(config_path)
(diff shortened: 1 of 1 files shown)


src/levanter/config.py
```diff
@@ -93,26 +93,51 @@ def _maybe_get_config_path_and_cmdline_args(args: List[str]):
     If URL, we need to download it and save it to a temp file. We then want to remove --config_path
     from the cmdline args so that draccus doesn't try to load it as a config path and return it separately here
     along with the modified cmdline args.
+    We also accept ... --configs <config1> <config2> ... and concatenate them into a single config.
```
percyliang, line 96: Why are we not using something like argparse rather than doing command-line parsing ourselves?

src/levanter/config.py
```diff
@@ -93,26 +93,51 @@ def _maybe_get_config_path_and_cmdline_args(args: List[str]):
...
-            atexit.register(lambda: os.unlink(temp_file.name))
-            fs.get(fs_path, temp_file.name)
-            config_path = temp_file.name
+        config_args = ["--config_path", "--config", "--configs"]
+        found_indices = [args.index(arg) for arg in config_args if arg in args]
+        if len(found_indices) > 1:
+            raise ValueError(f"Multiple config args found in {args}")
+        config_path_index = found_indices[0] + 1
 
+        config_paths = []
```
percyliang, line 107: add type hint

src/levanter/config.py
```diff
@@ -93,26 +93,51 @@ def _maybe_get_config_path_and_cmdline_args(args: List[str]):
...
         args = args.copy()
-        del args[config_path_index]
-        del args[config_path_index]
-        return config_path, args
+        del args[config_path_index - 1]
+        config_path_index -= 1
+
+        while config_path_index < len(args) and not args[config_path_index].startswith("-"):
```
percyliang, line 112: I think the code would be cleaner if we first (1) parsed the args to get a list of config_paths, some of which could be remote, and then (2) downloaded them to temp files, and then (3) did the merge...

src/levanter/config.py
```diff
@@ -93,26 +93,51 @@ def _maybe_get_config_path_and_cmdline_args(args: List[str]):
...
+
+            config_path = args[config_path_index]
+
+            if urllib.parse.urlparse(config_path).scheme:
+                fs: AbstractFileSystem
+                fs, fs_path = fsspec.core.url_to_fs(config_path)
+                temp_file = tempfile.NamedTemporaryFile(prefix="config", suffix=".yaml", delete=False)
+                atexit.register(lambda: os.unlink(temp_file.name))
+                fs.get(fs_path, temp_file.name)
```
percyliang, line 121: I know this was there before, but can we just read the config instead of copying it into a temporary file, especially since we're going to be merging?  And if a config file has `foo: 3\nfoo:4`, does override work properly?

percyliang (commented):
Overall, functionality looks good.  I think logic could be cleaner (but some of my comments apply to previous code too).

Most importantly, just wanted to check that duplicate keys are not a problem with the merge (override works as expected).

src/levanter/config.py
```diff
@@ -93,26 +93,51 @@ def _maybe_get_config_path_and_cmdline_args(args: List[str]):
...
+        if len(config_paths) == 1:
+            merged_config_path = config_paths[0]
+        elif len(config_paths) > 1:
+            # merge the configs by concatenating them
+            temp_merged_config_path = tempfile.NamedTemporaryFile(prefix="config_merged", suffix=".yaml", delete=False)
+            atexit.register(lambda: os.unlink(temp_merged_config_path.name))
+            with open(temp_merged_config_path.name, "w") as f:
+                for config_path in config_paths:
+                    with open(config_path) as config_file:
+                        f.write(config_file.read())
```
blahBlahhhJ, line 138: I'm assuming you are literally stitching multiple config files together?

e.g. we must have
```yaml
#data.yaml
data:
    cache_dir: ...
    ...
```

```yaml
#model.yaml
model:
    hidden_dim: 2048
    ...
```

Is the purpose of this PR for stitching top level configs (e.g. `model`, `data`, `trainer`) or do we also want to support stitching sub-configs (e.g. `data.train_weights.yaml`, `data.configs.yaml` ...)? I think it's better to utilize the yaml `!include` syntax, or else users need to pay extra attention on the order of the config they put in...

src/levanter/config.py
dlwh: we are using argparse at the end of the day, but this is a hacky layer to support reading configs from gcs (and i guess now supporting multiple configs)

it'd be better to do this in draccus.

src/levanter/config.py
dlwh: agree with jason

src/levanter/config.py
dlwh: we should just move all this logic into draccus

src/levanter/config.py
abhinavg4: @dlwh Let me know if we are doing this in draccus or levanter

src/levanter/config.py
abhinavg4: Added

src/levanter/config.py
abhinavg4: Incase we want to read the config directly, we should probably do it in Draccus. Also override works properly right now.

abhinavg4:
> Overall, functionality looks good. I think logic could be cleaner (but some of my comments apply to previous code too).
> 
> Most importantly, just wanted to check that duplicate keys are not a problem with the merge (override works as expected).

Yes the override works as expected. If you give --configs <config1> <config2> then <config2> overrides <config1>

src/levanter/config.py
abhinavg4: I think the problem with using `!include` is that when we want to do scaling laws, we would have to create a large number of configs. 

I think the vision I had was:

1. You can run with config just as normal run (Use `!include` and stuff)
2. We give multiple configs and stitch them together to make one config file.

dlwh: it's a good point

src/levanter/config.py
dlwh: draccus is probably better

dlwh (approved):

PR #687, opened 2024-08-11
Allow mounting extrenal directory to docker image

blahBlahhhJ:
This PR supports the case when users want to add an external directory (e.g. configs) to the docker image and send to TPU.

Specify `--mount_path` to point to the directory/file to mount, and the docker will copy the content of this directory/file to the built docker image (under the same path). Users can then reference to the content in this path in subsequent commands (e.g. reference an external config in `train_lm.py --config_path`)

Note that `--mount_path` can be using relative path (e.g. `--mount_path ../my_config/`) but any reference of external path in subsequent commands must be using absolute path (e.g. `python train_lm.py --config_path /home/shared/my_config/`). This is because the levanter repo is always mapped to `/opt/levanter` so the relative path may not be valid.

### DIFF
--- a/docs/Getting-Started-TPU-VM.md
+++ b/docs/Getting-Started-TPU-VM.md
@@ -138,6 +138,13 @@ To run in the foreground, use `--foreground` with the `launch.py` script. You sh
 python infra/launch.py -- python src/levanter/main/train_lm.py --config_path config/gpt2_small.yaml --trainer.checkpointer.base_path gs://<somewhere>'
 ```
 
+### Using external directory/file
+
+In case that you want to reference some external directory/file outside of the levanter repo, you can do it by mounting the external directory/file to the docker image so that it becomes accessible in TPU instances. You can specify the path you want to mount by `--mount_path` with the `launch.py` script. Then, you should be able to use the mounted files in arguments in `train_lm.py` etc.
+```bash
+python infra/launch.py --mount_path <external path> -- python src/levanter/main/train_lm.py --config_path <external path> --trainer.checkpointer.base_path gs://<somewhere>'
+```
+
 ### Babysitting Script
 
 If you are using a preemptible TPU VM, you probably want to use the "babysitting" script that automatically re-creates
--- a/docs/Getting-Started-TPU-VM.md
+++ b/docs/Getting-Started-TPU-VM.md
@@ -140,7 +140,7 @@ python infra/launch.py -- python src/levanter/main/train_lm.py --config_path con
 
 ### Using external directory/file
 
-In case that you want to reference some external directory/file outside of the levanter repo, you can do it by mounting the external directory/file to the docker image so that it becomes accessible in TPU instances. You can specify the path you want to mount by `--extra_context` with the `launch.py` script. Then, you should be able to use the mounted files in arguments in `train_lm.py` etc.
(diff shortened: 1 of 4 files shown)


blahBlahhhJ:
If we want to solve the relative/absolute path issue I mentioned in the description, we can mount levanter to the same path as in the local machine as well. This should be an easy change and I don't see a particular reason why levanter must be mounted to `/opt/levanter`.

dlwh:
Could you add a note to Getting-Started-TPU-VM.md?

dlwh:
@rjpower could I get you to take a look?

infra/push_docker.py
```diff
@@ -160,17 +194,17 @@ def push_to_github(local_image, tag, github_user=None, github_token=None, docker
         print(login_process.communicate(input=github_token.encode(), timeout=10))
 
     remote_name = f"ghcr.io/{github_user}/{local_image}:{tag}"
-    local_name = build_docker(docker_file=docker_file, image_name=local_image, tag=tag)
+    local_name = build_docker(docker_file=docker_file, image_name=local_image, tag=tag, mont_src=mount_path)
```
rjpower, line 197: typo: `mount_src`

rjpower (commented):
I'll defer to David on what makes sense for the project, but s this only for things like config files? If so, does it make sense to consider something having a convention for user resources, and relying on the default Docker behavior? e.g.

> .gitignore

```
# User config & local data directories
config/user
.user
```

That would simplify our logic a bit.

If that doesn't work, then I'd at least modify the naming for this approach.  Mounting & external directories in Docker in general refers to the execution phase (e.g. files on the TPU worker). e.g. something like `extra_context` would be less likely to be confusing.

dlwh:
Thanks @rjpower : for Marin we want to have our config files checked into that repo, which is what led to this particular solution. I guess one could do a checkout of marin or something into the levanter repo, but that felt even more untidy? I dunno

Levanter works with gs:// paths too but, again, we want them in the repo.

rjpower:
Ah, I see. Yeah I don't know of any way to have Docker follow symlinks directly, so that leaves you:

* This approach
* Use symlinks + tar to explicitly make a build context  `tar --exclude-from=.dockerignore -ch . | docker build -`
* Copy your files ahead of time outside of the script

Certainly the alternatives don't seem any better than this approach and probably worse. (You could run the Docker build from the other project if you wanted, and embed Levanter, but that probably has it's own issues).

docs/Getting-Started-TPU-VM.md
```diff
@@ -138,6 +138,13 @@ To run in the foreground, use `--foreground` with the `launch.py` script. You sh
 python infra/launch.py -- python src/levanter/main/train_lm.py --config_path config/gpt2_small.yaml --trainer.checkpointer.base_path gs://<somewhere>'
 ```
 
+### Using external directory/file
+
+In case that you want to reference some external directory/file outside of the levanter repo, you can do it by mounting the external directory/file to the docker image so that it becomes accessible in TPU instances. You can specify the path you want to mount by `--extra_context` with the `launch.py` script. Then, you should be able to use the mounted files in arguments in `train_lm.py` etc.
```
dlwh, line 143: maybe update the language to not use the word mount here?

dlwh (approved):

PR #696, opened 2024-08-21
Expose infra as a package, publish dev builds

dlwh:
We need to be able to launch docker jobs programmatically from Marin, so this refactors the infra docker stuff into a package, and does a bit of refactoring as I go.

Changes:

* Publish levanter dev build automatically
* move stuff from `infra/**/*.py` into `src/levanter/infra/{cli_helpers,tpu,docker}.py`
* make the extra context stuff be optional (I actually have another way of dealing with it in Marin now

other minor stuff


rjpower (approved):
LGTM!

PR #701, opened 2024-08-22
Update beta2=0.95

dlwh:


percyliang:
Could we add @siddk 's links to why 0.95?

dlwh:
done

PR #715, opened 2024-09-03
use hf config from checkpoint by default

### LINKED ISSUE #681: Is there a way to initialize Mistral from a locally downloaded checkpoint?
I have tried using
initialize_from_hf: "/path/to/mistral_checkpoint
use_hf_model_config: true

But it tries to go to HF website to get the model checkpoint?

Is it a bug?

dlwh:
So we should actually respect the config in the HF checkpoint directory since that's usually right. A bunch of tests were broken, but we didn't notice because they all need torch and we don't install torch in CI... (I guess we should)

Fixes #681 ( I think)

### DIFF
--- a/docs/dev/Port-Models.md
+++ b/docs/dev/Port-Models.md
@@ -242,12 +242,7 @@ with tempfile.TemporaryDirectory() as tmpdir:
     ck_path = f"{tmpdir}/hf_model"
     hf_model.save_pretrained(ck_path)
 
-    model = converter.load_pretrained(
-        config.model_type,
-        config,
-        ck_path,
-        resize_vocab_to_match_tokenizer=False
-    )
+    model = converter.load_pretrained(config.model_type, ref=ck_path, resize_vocab_to_match_tokenizer=False)
 
 # compare the output values between Levanter and HF
 # ...
--- a/examples/alpaca-lora/alpaca_lora.py
+++ b/examples/alpaca-lora/alpaca_lora.py
@@ -90,7 +90,7 @@ def train(config: TrainArgs):
         logger.info(f"Loading pretrained model from {converter.reference_checkpoint}")
         # load untrainable params in compute precision to save memory
         model: LmHeadModel = converter.load_pretrained(  # type: ignore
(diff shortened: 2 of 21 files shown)


mhmaqbool:
It would be great if a sample configuration for mistral-7b is made available for loading from a local checkpoint. I still note that HF is referred to for loading the checkpoint, but probably that's something to do with my configuration.
What did I try?
In the configuration (mistral-7b),
1. 
      `initialize_from: /path/to/local/checkpoint`
      that results in the following error,
      `draccus.utils.DecodingError: The fields 'initialize_from' are not valid for TrainLmConfig`
2.
    `initialize_from_hf:/path/to/local/checkpoint'
     tries to find the checkpoint from HF

Any advise please?

dlwh:
does `/path/to/local/checkpoint/model.safetensors.index.json` not exist?

mhmaqbool:
Yeah, I have PyTorch installed and tried again. Following files are available in the checkpoint

- pytorch_model.bin.index.json
- pytorch_model-00001-of-00002.bin
- pytorch_model-00002-of-00002.bin
- a few other files (tokenizer, model etc)

Still tries to refer to HF website for getting the checkpoint?
This is my configuration,

`data:
  train_urls: ["/pth/to/train/json"]
  validation_urls: ["/pth/to/val/json/"]
  tokenizer: "/local/mistral-7b/checkpoint"
model:
  type: mistral
# TODO: uncomment this once we resolve the resource exhaustion issue
  initialize_from_hf: "/local/mistral-7b/checkpoint"
  use_hf_model_config: true
trainer:
  wandb:
    project: "levanter"
    tags: ["openwebtext", "mistral"]

  mp: p=f32,c=bfloat16
  train_batch_size: 256  # set for v4-64 TPU
  num_train_steps: 1000
  steps_per_eval: 50
  tensor_parallel_axes: ["mlp", "heads"]
  fsdp_axis: "embed"
  batch_axis: "batch"
optimizer:
  learning_rate: 1.2E-5  # set low for fine-tuning
  weight_decay: 0.1
  min_lr_ratio: 0.1`

PR #729, opened 2024-09-15
add bits-per-byte calculation to levanter

dlwh:


src/levanter/utils/hf_utils.py
```diff
@@ -20,3 +25,22 @@ def num_cpus_used_by_tokenizer(tokenizer) -> int:
             return min(max(1, logical_cpu_core_count() - 2), 12)
     else:
         return 1
+
+
+def byte_length_of_token(tokenizer, idx: int) -> int:
+    # this is a pain because we want the prefix spaces, but we don't want extra noise for bytes
```
percyliang, line 31: Does this correspond to other implementations that need to compute bpb?  Would be good to reference and comment on whether we're doing the same thing.

tests/test_hf_utils.py
```diff
@@ -22,3 +24,27 @@ def test_load_tokenizer_in_memory_fs():
         )
     tokenizer = load_tokenizer("memory://foo/")
     assert len(tokenizer) == 5027
+
+
+@skip_if_hf_model_not_accessible("meta-llama/Llama-2-7b-hf")
```
percyliang, line 29: Should we add some tests with more funky Unicode characters and tokens that don't align on character boundaries?

src/levanter/eval.py
```diff
@@ -304,19 +342,31 @@ def evaluate(self, m: LmHeadModel):
...
+            macro_avg_bpb = hax.named_jit(lambda x: hax.mean(x).array)(tag_avg_bpb).item()
+        else:
+            micro_bpb = None
+            tag_avg_bpb = None
+            macro_avg_bpb = None
+
         tag_macro_loss = {}
         tag_micro_loss = {}
+        tag_macro_bpb = {}
+        tag_micro_bpb = {}
```
percyliang, line 363: add type hints?

percyliang (approved):
Great.  I'm not used to reading levanter code, so it's sometimes hard to figure out what the type/dimensions of some of the tensors are...  The annotations in the comments (e.g., [Batch, Pos]) are very helpful, but perhaps this could be done more pervasively?

The other thing I was curious about how this bpb implementation (which seems subtle) compares with other ones (e.g., lm-evaluation-harness).

PR #744, opened 2024-09-25
Make new tokenization ~67% faster

dlwh:
Until this PR, the new cache would write every ~8K docs it received, regardless of how many were queued up. (This isn't quite correct, but close enough) Writes are high latency, so this wasn't the best choice.

We now instead write whatever is ready every second or so. In practice this makes us about 67% faster at tokenizing.

image


percyliang:
Nice!

PR #746, opened 2024-09-25
Adding supervised data config

TheQuantumFractal:


src/levanter/data/text.py
```diff
@@ -584,6 +588,103 @@ def tagged_eval_sets(
...
+
+    cache_dir: str = "cache/"
+
+    tags: Optional[List[str]] = None
+    """tags for the dataset. Typically the name of the dataset in the config will be added as a tag as well"""
+    name: Optional[str] = None  # name for hf dataset
+
+    validation_urls: List[str] = ()  # type:ignore
+
+    # def token_seq_dataset(
```
ahmeda14960, line 603: Can you remove this code if it's commented out?

src/levanter/main/train_lm.py
```diff
@@ -16,7 +16,7 @@
 from levanter import callbacks
 from levanter.checkpoint import load_checkpoint
 from levanter.compat.hf_checkpoints import HFCompatConfig, save_hf_checkpoint_callback
-from levanter.data.text import CausalLmDataset, LMDatasetConfig, LMMixtureDatasetConfig
+from levanter.data.text import CausalLmDataset, LMDatasetConfig, LMMixtureDatasetConfig, LMSupervisedDatasetConfig
```
ahmeda14960, line 19: Should you maybe make a different file? I think train_lm is typically expected within levanter to be reserved for pre-training and is specific to the autoregressive loss. I think there's separate files for fine-tuning alpaca for example. 

I'm not sure what the convention is cc: @dlwh

src/levanter/data/text.py
```diff
@@ -584,6 +588,103 @@ def tagged_eval_sets(
...
+    """
+    Prepare an example for training. This function converts the (cached) batch encoding into an LmExample.
+
+    It goes through the following steps:
+
+    1. Pad the batch to the maximum length.
+    2. Mask out the input and prompt if requested.
+    3. Create an LmExample with the input_ids as the input and the next token as the target.
+    """
+    with local_cpu_mesh():
```
ahmeda14960, line 655: Why with local cpu mesh here? how would this work if you wanted to do training?

src/levanter/data/text.py
```diff
@@ -584,6 +588,103 @@ def tagged_eval_sets(
...
+
+def mk_supervised_dataset(config: LMSupervisedDatasetConfig, tokenizer: PreTrainedTokenizerBase):
+    import levanter.data
+
+    validation_urls = [url for url_pat in config.validation_urls for url in fsspec_expand_glob(url_pat)]
+    dataset = levanter.data.datasource_from_jsonl(validation_urls)
+
+    output_exemplar = {"input_ids": np.zeros((0,), dtype=np.int32), "sources_len": np.zeros((), dtype=np.int32)}
+
+    dataset = dataset.map_batches(lambda ex: preprocess_supervised_example(ex, tokenizer), batch_size=128, num_cpus=num_cpus_used_by_tokenizer(tokenizer), output_exemplar=output_exemplar)  # type: ignore
```
ahmeda14960, line 680: How exactly is ```output_exemplar``` used ? I see that we modified the datasets to accept it as an argument but I don't see any changed logic that actually uses it.

src/levanter/data/text.py
TheQuantumFractal: It was throwing an error from the precommit. I need to check with David if we need these functions. If so, I'll need to fix some of these

src/levanter/data/text.py
TheQuantumFractal: This is just for preprocessing the supervised examples on the CPU (tokenizing and masking them). They are put on TPUs after.

src/levanter/data/text.py
TheQuantumFractal: it's used when building the cache. Basically the rank of the input ids and sources len has to be known in order to load from or save to the cache (and the original sharded datasource struct doesn't support this)

src/levanter/main/train_lm.py
TheQuantumFractal: For supervised finetuning, we maybe should make a separate file but the changes in this file are for measuring eval loss on a supervised task.

ahmeda14960: Got it LGTM

src/levanter/data/text.py
```diff
@@ -584,6 +588,103 @@ def tagged_eval_sets(
...
+    #         return {}
+
+    # Add tagged eval set with split for auxiliary and validation dataset
+
+
+def preprocess_supervised_example(batch, tokenizer: PreTrainedTokenizerBase):
+    sources = [example["input"] for example in batch]
+
+    targets = [f"{example['output']}" for example in batch]
+    # TODO: this seems pretty wasteful since you end up tokenizing twice, but it's how the original code does it.
```
dlwh, line 632:
```suggestion
    # TODO: this seems pretty wasteful since you end up tokenizing twice, but it's how alpaca does it.
```

src/levanter/data/text.py
dlwh: rm and don't have LMSupervisedDatasetConfig implement LMDatasetSourceConfig

src/levanter/data/text.py
```diff
@@ -390,6 +390,22 @@ def batch_size(self) -> int:
         return self._batch_size
 
 
+def fsspec_expand_glob(url):
```
dlwh, line 393: can you move to utils/fsspec_utils

src/levanter/data/text.py
dlwh: Yeah it has to be local_cpu_mesh. I'll try to make that part get hidden

dlwh (approved):
Thanks!

PR #747, opened 2024-09-26
Tweaks to Ray TPU stuff

dlwh:
1. num_tpus=1 is actually a bad idea because Ray will mask out the other tpus
2. force non-docker workloads to run in a separate process for stability

### DIFF
--- a/src/levanter/infra/ray_tpu.py
+++ b/src/levanter/infra/ray_tpu.py
@@ -2,14 +2,19 @@ import dataclasses
 import logging
 import os
 import subprocess
+import tempfile
+import time
 from dataclasses import dataclass
-from typing import Sequence
+from pathlib import Path
+from typing import Optional, Sequence
 
 import draccus
 import ray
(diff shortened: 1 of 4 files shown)


src/levanter/infra/ray_tpu.py
```diff
@@ -210,8 +247,92 @@ def _handle_ray_error(tpu_info: _TpuInfo, e: RayError):
         return TpuRunError(tpu_info, e)
 
 
+def _forkify_remote_fn(remote_fn: RemoteFunction | Callable):
+    """
+    This is a bit of a hacky way to force a remote function to run in its own process.
```
percyliang, line 252: What's the high-level mechanism for this?

percyliang (approved):

src/levanter/infra/ray_tpu.py
dlwh: multiprocessing

PR #752, opened 2024-10-01
Simplify tokenization pipeline, make it work with large numbers of shards again, (re)add configuration metadata to cache

dlwh:


ahmeda14960:
I think you need to change the toml for ```deepdiff```

ahmeda14960:
Seems to work for me!

PR #753, opened 2024-10-04
make deepdiff optional

dlwh:


src/levanter/store/cache.py
```diff
@@ -14,7 +14,7 @@
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, TypeVar, Union
 
-import deepdiff
+# import deepdiff
```
percyliang, line 17: Remove comment?

percyliang (commented):
What's the motivation for making this optional?  Can we not include it like a normal package?

dlwh:
so we can avoid a full redeploy of the marin cluster

percyliang:
What's the problem with a full redeploy?  It seems like we want to be able to redeploy easily anyway...

dlwh:
i dunno, i'll close this.

PR #754, opened 2024-10-05
allow mixture components to override cache_dir

dlwh:


src/levanter/data/text.py
```diff
@@ -788,10 +796,23 @@ def build_caches(
             if weight == 0 and split == "train":
                 continue
 
-            source_config_dict = source_config.__dict__
+            source_config_dict = dict(**source_config.__dict__)
```
percyliang, line 799: Add comment on why we need to delete it - to make the LMDatasetMixtureComponentConfig act like a LMDatasetSourceConfig?

src/levanter/data/text.py
```diff
@@ -524,13 +524,18 @@ def fsspec_expand_glob(url):
         return urls
 
 
+@dataclass
+class LMDatasetMixtureComponentConfig(LMDatasetSourceConfig):
```
percyliang, line 528: The only thing different between the base and derived class is cache_dir?  Seems weird since this on the surface doesn't seem to relate to mixtures...

percyliang (approved):

src/levanter/data/text.py
dlwh: more or less. i also cleaned this part up a bit

src/levanter/data/text.py
dlwh: yeah i can just put it in the base i guess

PR #755, opened 2024-10-05
a few final tweaks for marin runs

dlwh:


src/levanter/data/text.py
```diff
@@ -705,6 +707,8 @@ def _convert_id_to_token(self, index: int) -> str:
 class LMMixtureDatasetConfig(LMTaskConfig):
     """This class represents a mixture of datasets with their associated weights."""
 
+    cache_dir: Optional[str] = "cache/"
```
percyliang, line 710: Does it make sense to move this into a base class?

percyliang (approved):

PR #757, opened 2024-10-07
Add an actor pool for batch processing, switch to a thread for writing batches instead of a future

dlwh:
About a 20% speedup. I think there's another 20-40% hiding in there.


abhinavg4:
Adding myself here to track this issue

config/data/pile_mixture.yaml
```diff
@@ -11,11 +14,11 @@ configs:
       - gs://levanter-data/pile-domains/books2/{00..29}.jsonl.zst
     validation_urls:
       - gs://levanter-data/pile-domains/books2/val.jsonl.zst
-  books3:
-    train_urls:
-      - gs://levanter-data/pile-domains/books3/{00..29}.jsonl.zst
-    validation_urls:
-      - gs://levanter-data/pile-domains/books3/val.jsonl.zst
+#  books3:
```
percyliang, line 17: Why is this back, commented?

percyliang (approved):
Nice work!  It would be nice to have a high-level overview of what's happening - maybe just stick some file-level docstrings to provide more context about the motivation?  (Doesn't have to be extensive.)

abhinavg4 (approved):

config/data/pile_mixture.yaml
dlwh: super long docs aren't working super well atm. This is a fight I've not managed to fully win.

PR #758, opened 2024-10-08
Update Audio Data Loader to Support Mixture Dataset

Helw150:
Pulls in the New Mixture Features Into Audio Space!

Tested that this fixes the previous epoching errors in the whisper_tiny config.


Helw150:
Weights & Biases Link for the test training run on the Tutorial Docs!

https://wandb.ai/i18nlp/levanter/runs/rwaft6sv?nw=nwuserheld

dlwh (approved):
Thanks!

dlwh:
also, huge relief that it actually works for audio!

PR #759, opened 2024-10-09
Support multiprocessing for backgrounditerable to mitigate TensorStore crash

dlwh:
Both @Helw150 and I have observed a crash when resuming from a checkpoint. I'm pretty sure it's a race or similar in tensorstore b/c checkpointing and dataloader use TS.

Also, Python is so dumb.


dlwh:
(haven't tested yet)

PR #764, opened 2024-10-13
Support Parquet files in ShardedDataSource

### LINKED ISSUE #763: support loading parquet files in addition to jsonl.gz
This will allow us to more easily train on datasets such as SlimPajama-6B, fineweb-edu, etc.

nikil-ravi:
This PR creates a `ParquetDataSource` class to support loading `.parquet` files.
Closes #763

### DIFF
--- a/src/levanter/data/sharded_datasource.py
+++ b/src/levanter/data/sharded_datasource.py
@@ -21,7 +21,6 @@ import datasets
 import fsspec
 import numpy as np
 import pyarrow.parquet as pq
-import pandas as pd
 
 from levanter.utils import fsspec_utils
 
@@ -248,7 +247,7 @@ class TextUrlDataSource(ShardedDataSource[str]):
                     table = pq.read_table(f)
                     sliced_table = table.slice(row)
                     for record in sliced_table.to_pylist():
-                        yield record[self.text_key] # assumes text_key is in record
+                        yield record[self.text_key]  # assumes text_key is in record
                 case _:
                     raise ValueError(f"Unknown format {format}")
 
@@ -441,7 +440,7 @@ class ParquetDataSource(ShardedDataSource[dict]):
         url = self._shard_name_to_url_mapping[shard_name]
         with fsspec.open(url, "r", compression="infer") as f:
             table = pq.read_table(f)
-            sliced_table = table.slice(row) # zero-copy slicing
+            sliced_table = table.slice(row)  # zero-copy slicing
             for record in sliced_table.to_pylist():
                 yield record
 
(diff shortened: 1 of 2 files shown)


abhinavg4 (approved):
Looks good, Thaks

tests/test_sharded_dataset.py
```diff
@@ -24,6 +25,51 @@ def test_sniff_format_for_json():
...
+    import pyarrow as pa
+    import pyarrow.parquet as pq
+
+    with tempfile.NamedTemporaryFile(suffix=".parquet", delete=False) as f:
+        # Create a simple dataset
+        data = {"column1": ["value1", "value2", "value3"], "column2": [10, 20, 30]}
+        table = pa.Table.from_pydict(data)
+        pq.write_table(table, f.name)
+
+    try:
```
percyliang, line 57: Why not put this within the context manager and set delete to True?

percyliang (approved):

tests/test_sharded_dataset.py
nikil-ravi: Done!

src/levanter/data/sharded_datasource.py
```diff
@@ -238,6 +243,11 @@ def open_shard_at_row(self, shard_name: str, row: int) -> Iterator[str]:
                     data = json.load(f)
                     for doc in data[row:]:
                         yield doc[self.text_key]
+                case ".parquet":
+                    table = pq.read_table(f)
```
dlwh, line 247: read_table is expensive in the general case and it would be better to look at the metadata to figure out which row group to start on and then use `read_row_group` I think

src/levanter/data/sharded_datasource.py
```diff
@@ -417,6 +427,24 @@ def open_shard_at_row(self, shard_name: str, row: int) -> Iterator[dict]:
             return iter(data[row:])
 
 
+class ParquetDataSource(ShardedDataSource[dict]):
```
dlwh, line 430: ideally the TextUrlDataSource would also work with parquet files. That is what we use for training configs typically

dlwh (commented):
thanks for doing this!

src/levanter/data/sharded_datasource.py
nikil-ravi: Makes sense, I added a new test for this in #766

src/levanter/data/sharded_datasource.py
nikil-ravi: Thanks for suggesting this, incorporated this in #766

PR #765, opened 2024-10-14
fix crash in data loader caused by using stale array

dlwh:


src/levanter/data/loader.py
```diff
@@ -98,6 +99,8 @@ def __iter__(self):
         return self.iter_from_step(None)
 
     def iter_from_step(self, start_from_batch: Optional[int] = None):
+        # sometimes we pass in an array for the start_from_batch, so we need to check for that
+        start_from_batch = int(start_from_batch) if start_from_batch is not None else None
```
Helw150, line 103: Is there a way to account for the case the comment describes in the MyPy type hint?

src/levanter/data/loader.py
```diff
@@ -109,115 +112,132 @@ def __init__(self, data_loader: DataLoader, start_from_batch: Optional[int] = No
...
-                        return raw_array
-
-                gda_leaves = [
-                    make_global_array_for_leaf(leaf_index, _batchified_shape(self.dl.Batch, item_leaf))
-                    for leaf_index, item_leaf in enumerate(self.dl._ex_leaves)
-                ]
-
-                gda_tree = jax.tree.unflatten(self.dl._ex_structure, gda_leaves)
-                yield gda_tree
+                    # TODO: this doesn't work with named axes
```
Helw150, line 208: Wouldn't anything with named axes go into the `if isinstance(leaf_data, hax.NamedArray)` above rather than the parent `else`?

Helw150 (approved):
LGTM - just left a few notes on comments I found confusing if I imagined reading them without context.

Helw150:
The error in the TPU test doesn't look like a flaky one to me though!

src/levanter/data/loader.py
dlwh: i could make it be an jnp.ndarray, but i'd rather we didn't pass it in and i'm vaguely annoyed mypy isn't raising an error already.

src/levanter/data/loader.py
dlwh: yeah old comment, sorry

dlwh:
figured it out

PR #771, opened 2024-10-19
add multislice support in ray

blahBlahhhJ:


src/levanter/infra/cli_helpers.py
```diff
@@ -59,6 +59,37 @@ def get_git_commit():
...
+            "-v",
+            "/tmp:/tmp",
+        ]
+
+        self.env_part = []
+        self.add_env(env)
+
+        self.cmd_part = [image_id, *command]
+
+    def add_env(self, env):
```
dlwh, line 84: type hints please

src/levanter/infra/cli_helpers.py
```diff
@@ -59,6 +59,37 @@ def get_git_commit():
     return subprocess.check_output(["git", "rev-parse", "HEAD"]).decode("utf-8").strip()
 
 
+class DockerRunCommand:
+    def __init__(self, image_id, command, *, foreground, env, name="levanter"):
```
dlwh, line 63: type hints please (mostly for env)

src/levanter/infra/ray_tpu.py
```diff
@@ -62,21 +63,28 @@ class TpuRunError(_TpuRunResult):
     error: Exception
 
 
-def run_on_pod(remote_fn: RemoteFunction | Callable, tpu_type: str) -> ray.ObjectRef:
+def run_on_pod(docker_cmd: DockerRunCommand, name: str, tpu_type: str) -> ray.ObjectRef:
```
dlwh, line 66: hrm, we use the non-docker version in Marin so we do actually want the non-docker version to work.

Do we need this change?

src/levanter/infra/ray_tpu.py
```diff
@@ -93,10 +101,87 @@ def do_run(remote_fn) -> _TpuRunResult:
                     logger.exception("Failed to kill job after primary failure")
             return _handle_ray_error(info, e)
 
-    return do_run.remote(remote_fn)
+    return do_run.remote(docker_cmd, name)
+
+
+def run_on_pod_multislice(docker_cmd: DockerRunCommand, name: str, tpu_type: str, num_slices: int) -> ray.ObjectRef:
```
dlwh, line 107: could we maybe just always make the dockerruncommand inject the MEGASCALE environment variables and use ray runtime environments to inject the environment variables that way?

dlwh (commented):
i'd really prefer we didn't require docker since we're not using it in marin.

src/levanter/infra/ray_tpu.py
blahBlahhhJ: single slice doesn't need this change. I only changed it to make it consistent with the multislice. I don't think there's a good way to undo this change for multislice tho.

src/levanter/infra/ray_tpu.py
```diff
@@ -193,11 +274,103 @@ def run_on_pod_resumable(remote_fn, tpu_type, max_retries_preemption=1e6, max_re
...
+            if "preempted" in str(e).lower():
+                num_preemptions += 1
+                logger.warning(f"Preempted {num_preemptions} times, {e}")
+            else:
+                num_failures += 1
+                logger.warning(f"Failed {num_failures} times", exc_info=e)
+            continue
+        except Exception as e:
+            for f in futures:
+                ray.cancel(f)
```
dlwh, line 320: try catch these so we cancel the rest

dlwh (approved):

PR #776, opened 2024-10-23
Infra tweaks

dlwh:
found an issue with the ray stuff where somehow ray seemingly wasn’t raising RayTaskErrors but other exceptions. also fixes an issue with quoting


blahBlahhhJ (approved):
LGTM

PR #779, opened 2024-10-30
Merging DiVA to Levanter Main

Helw150:
Cleaned up version of my code for the Distilled Voice Assistant models that I trained using a fork of Levanter!

@dlwh Main thing I want to check in with you here is what the appropriate design pattern you think would make sense for initializing the model weights from multiple other pretrained models would be! What I've done here is much cleaner than what I did originally for the paper, but still feels a bit messy.

Testing Procedure for the correctness of this training code:
I trained a new DiVA model with this updated code and Llama 3.2 1B using the config in `diva_flash.yaml`. 

Training Log is here: https://wandb.ai/i18nlp/levanter/runs/jnxp463y?nw=nwuserheld
Resulting model is on HF in PyTorch form here: https://huggingface.co/WillHeld/DiVA-llama-3.2-1b
Demo which confirmed the result is ~reasonable here for now: https://b3f161194b514a990f.gradio.live/


src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+        ref: Optional[Union[str, RepoRef]] = None,
+        config: Optional[HFCompatConfig] = None,
+        axis_mapping: Optional[ResourceMapping] = None,
+        resize_vocab_to_match_tokenizer: bool = True,
+        dtype: Optional[jnp.dtype] = None,
+    ) -> ModelWithHfSerializationMixin:
+        lev_model: DivaModel = super().load_pretrained(
+            DivaModel, ref, config, axis_mapping, resize_vocab_to_match_tokenizer, dtype
+        )  # type: ignore[assignment]
+        llm: Union[LlamaLMHeadModel | MistralLMHeadModel | GemmaLMHeadModel] = HFCheckpointConverter(
```
dlwh, line 50: Do you need this Union around your 3.10 union?

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+
+def load_correct_config(reference_decoder):
+    model_id = reference_decoder.lower()
+    hf_config = HfConfig.from_pretrained(reference_decoder)
+    if "llama" in model_id:
+        config = LlamaConfig.from_hf_config(hf_config)
+    elif "gemma" in model_id:
+        config = GemmaConfig.from_hf_config(hf_config)
+    elif "mistral" in model_id:
+        config = MistralConfig.from_hf_config(hf_config)
```
dlwh, line 85: should you raise a better error if it's none of these?

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+    return config
+
+
+def get_prefix(tokenizer_ref):
```
dlwh, line 89: doccomment maybe?

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+    tokenizer = AutoTokenizer.from_pretrained(tokenizer_ref)
+    prefix, suffix = tokenizer.apply_chat_template(
+        [{"role": "user", "content": "PLACEHOLDER"}], tokenize=False, add_generation_prompt=True
+    ).split("PLACEHOLDER")
+    prefix_tok = tokenizer.encode(prefix, add_special_tokens=False)
+    suffix_tok = tokenizer.encode(suffix, add_special_tokens=False)
+    return prefix_tok, suffix_tok
+
+
+@LmConfig.register_subclass("diva")
```
dlwh, line 99: AsrConfig?

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+@dataclass(frozen=True)
+class DivaConfig(HFCompatConfig, ASRConfig):
+    # Reference Models
+    reference_encoder: str = "openai/whisper-large-v3-turbo"
+    reference_decoder: str = "meta-llama/Llama-3.1-8B-Instruct"
+    reference_checkpoint: str = "WillHeld/DiVA-llama-3-v0-8b"
+    init_from_submodel: bool = True
+
+    # Connector Config
+    pre_audio_prompt = property(lambda self: get_prefix(self.reference_decoder)[0])
```
dlwh, line 109: cache_property or who cares?

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+def diva_connector_only(model):
+    frozen_tree = jax.tree_util.tree_map(lambda _: False, model)
+    return eqx.tree_at(
+        lambda tree: (tree.query_tokens, tree.projection.weight, tree.projection.bias, tree.connector),
+        frozen_tree,
+        (True, True, True, True),
+    )
+
+
+class DivaModel(eqx.Module, ModelWithHfSerializationMixin[DivaConfig]):
```
dlwh, line 164: maybe link to paper or something?

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+        if attn_mask is not None:
+            causal_mask = causal_mask & attn_mask
+        k_encoder, k_connector, k_decoder, k_head = maybe_rng_split(key, 4)
+
+        # Encode Audio With Whisper Encoder
+        audio_features = self.encoder(mel, key=k_encoder)
+
+        # Convert to Virtual LLM Tokens
+        virt_whisper_tokens = self.connector.transformer(
+            (self.query_tokens + self.query_position_embeds).broadcast_axis(OtherAxes),
```
dlwh, line 255: hrm i wouldn't think this should be necessary

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+        text = self.decoder.transformer(text_embeds, attn_mask=causal_mask, key=k_decoder)
+
+        push_forward_padding = hax.argsort(input_ids != pad_token_id, "position")
+        input_ids_right_pad = text_tokens[{"batch": hax.arange(Batch), "position": push_forward_padding}]
+        return (
+            audio["position", -1],
+            text[
+                {
+                    "batch": hax.arange(Batch),
+                    "position": (hax.sum(text_tokens == pad_token_id, "position") * -1) - 1,
```
dlwh, line 301: you just broke my brain

src/levanter/models/diva.py
```diff
@@ -0,0 +1,342 @@
...
+        # Compute Distillation Contrastive Loss
+        # Since the weight matrix is frozen, equivalent but faster than KL Div
+        audio_pred, text_pred, virtual_embeds, real_embeds = self(
+            example.audio, example.tokens, example.attn_mask, key=key
+        )
+        diff_distill = audio_pred - text_pred
+        kl_proxy_loss = hax.dot(diff_distill, diff_distill, axis="embed") ** 0.5
+
+        # Compute Contrastive Loss on Input
+        # Correct for Normal Autoregressive Loss Mask
```
dlwh, line 328: do you want to check that attn mask is causal or eh?

dlwh (approved):
Nice! Few minor comments. I don't understand everything but overall seems good to me!

Helw150:
Still investigating a bit of stuff here - used this code to reproduce the original DiVA model with Llama 3 8B - but hitting some weirdness with Llama 3.1 8B where the resulting model has a lot of repetitions.

Hypotheses:
- Some mismatch in RoPE still?
- Some issue where the 3.1 model needs multiple tokens of distillation?

src/levanter/models/diva.py
Helw150: Yes, good catch.

src/levanter/models/diva.py
Helw150: I'll double check. I've somewhat forgotten why I added these explicit broadcasts.

dlwh:
hrm happy to pair if that would be helpful. We can definitely investigate the rope thing. It's a constant pain

dlwh:
it looks like rope is exactly the same for llama3 and 3.1 so it's probably not that, unless you haven't merged main in month or two. I did fix a bug in #740

PR #782, opened 2024-11-02
Fix Llama 3 Tests

Helw150:
HuggingFace seems to have changed a few things around leading the Llama 3 roundtrip tests to hit errors. AFAICT, the Torch tests aren't running in CI so this just fixes the regression!

image


dlwh (approved):

dlwh:
somehow attention tests are hanging but i don't think that's you. Thanks!

PR #783, opened 2024-11-03
backport tokenize-independelty-then-copy from marin

dlwh:


percyliang (commented):
Do we need the version in marin after this or can we just use the levanter version?

dlwh:
this one should subsume the one in marin

PR #785, opened 2024-11-05
fix wandb key

dlwh:
turns out ray doesn't merge things when you use .options, which... what.


src/levanter/infra/ray_tpu.py
```diff
@@ -583,6 +589,16 @@ def _make_unique_job_id(client, run_id):
     return job_id
 
 
+def _deep_merge_envs(env1, env2):
```
percyliang, line 592: This is not called?

dlwh: oh sorry, i had pycharm inline it but it kept the declaration

PR #792, opened 2024-11-07
fix epochs in type signature, fix type checker

dlwh:


TheQuantumFractal (approved):
Seems reasonable to me

PR #795, opened 2024-11-08
Fix transformer-engine attention import

jennifgcrl:
Renamed upstream

### DIFF
--- a/src/levanter/models/attention.py
+++ b/src/levanter/models/attention.py
@@ -310,8 +310,8 @@ def _te_flash_attention(
     precision: PrecisionLike = None,
     block_size: Optional[int] = None,
 ):
-    from transformer_engine.jax.fused_attn import fused_attn  # noqa: F401
-    from transformer_engine.jax.fused_attn import AttnBiasType, AttnMaskType  # noqa: F401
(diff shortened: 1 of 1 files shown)


dlwh:
Thank you! I've been procrastinating fixing this

src/levanter/models/attention.py
```diff
@@ -358,14 +358,13 @@ def _te_flash_attention(
...
-        q=q_,
-        k=k_,
-        v=v_,
+        qkv=(q_, k_, v_),
         bias=fused_attn_bias,
         mask=fused_attn_mask,
         seed=prng,
         attn_bias_type=attn_bias_type,
         attn_mask_type=attn_mask_type,
+        qkv_layout=QKVLayout.BSHD_BSHD_BSHD,
```
dlwh, line 367: do you happen to know what the fastest layout is on nvidia? maybe it doesn't matter so much

dlwh (approved):

src/levanter/models/attention.py
jennifgcrl: I sure don't! If I end up benchmarking this I'll let you know.

PR #796, opened 2024-11-08
Relax restriction on scipy to be compatible with more recent numpy

jennifgcrl:
Correct me if I'm wrong but I think the <=1.12.0 bound on scipy is now vestigial? In any case, it's causing conflicts with other packages (e.g., most recent jax) that require a too new version of numpy for scipy to handle.


dlwh:
Yep, definitely vestigial! There was like a week where JAX didn't work without the bound and I forgot to fix it! Probably we can just delete it since we don't use scipy directly

dlwh:
Hi @jennifgcrl , thanks for raising this! I'm just gonna delete the dependency (in #797) instead, which should be at least as good. (If I'm wrong, let me know!)

PR #798, opened 2024-11-12
Misc fixes from sweep (disable blocked CE by default)

dlwh:
blocked CE is worse except it enables larger batch sizes


dlwh:
(a few changes are also in #801 and #800 )

percyliang (approved):

PR #799, opened 2024-11-12
bulk delete using STS

dlwh:


dlwh:
pre-commit unrelated

scripts/gcs_bulk_delete.py
```diff
@@ -0,0 +1,144 @@
...
+EMPTY_BUCKET = "levanter-empty"
+
+
+def schedule_gcs_deletion_job(project_id, gcs_bucket_name, path_to_delete):
+    """Schedules an STS job to delete all files in a GCS path and waits for completion."""
+
+    client = storage_transfer_v1.StorageTransferServiceClient()
+
+    # Define the transfer job
+    transfer_job = storage_transfer_v1.types.TransferJob(
```
percyliang, line 22: Confused - is this a transfer job or a deletion job?

scripts/gcs_bulk_delete.py
```diff
@@ -0,0 +1,144 @@
...
+    pbar = tqdm(desc=f"Waiting for Transfer Job :: {job_name}", unit="B", unit_scale=True)
+    while time.time() - start_time < timeout:
+        if (time.time() - start_time) % poll_interval == 0:
+            # Prepare the filter string to get the operations for the job
+            filter_string = f'{{"project_id": "{project_id}", "job_names": ["{job_name}"]}}'
+
+            # List transfer operations for the job
+            transfer_operations = operations_client.list_operations("transferOperations", filter_string)
+
+            # Get the latest operation
```
percyliang, line 80: latest by date - is this list guaranteed to be sorted?

scripts/gcs_bulk_delete.py
```diff
@@ -0,0 +1,144 @@
...
+        bucket_name, path_prefix = match.groups()
+        return bucket_name, path_prefix
+    else:
+        raise ValueError(f"Invalid GCS URL format: {gcs_url}")
+
+
+if __name__ == "__main__":
+    # Check for correct usage
+    if len(sys.argv) != 2:
+        print("Usage: python gcs_bulk_delete.py gs://bucket_name/path/to/delete")
```
percyliang, line 129:
```suggestion
        print(f"Usage: python {sys.argv[0]} gs://bucket_name/path/to/delete")
```

scripts/clean_old_checkpoints.py
```diff
@@ -74,9 +74,15 @@ def list_deletable_directories(base_dir):
 
 # Usage example:
 if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser(description="List directories that can be deleted.")
+    parser.add_argument("base_dir", help="The base directory to clean up.", type=str, nargs="+")
+    parser.add_argument("--age", help="The age in days of the checkpoints to delete.", type=int, default=30)
```
abhinavg4, line 81: Maybe: 

The age of the checkpoints to delete (in days)

scripts/gcs_bulk_delete.py
abhinavg4: I think it's a transfer job to an empty bucket

abhinavg4: Actually, it's a transfer from the Empty Bucker to the destination bucket, making sure `delete_objects_unique_in_sink` is true, which means "Whether objects that exist only in the sink should be deleted. "

abhinavg4 (approved):

scripts/gcs_bulk_delete.py
dlwh: yes, it's a dumb trick for STS. I'll add a comment

PR #801, opened 2024-11-12
tracker.finish to deal with subprocess stuff

dlwh:
we use subprocess in marin to invoke levanter, but subprocesses don't wait on other subprocesses somehow, and so wandb doesn't get a chance to finish. This solves this

### DIFF
--- a/src/levanter/main/train_lm.py
+++ b/src/levanter/main/train_lm.py
@@ -268,6 +268,9 @@ def main(config: TrainLmConfig):
             checkpointer = trainer.config.checkpointer.create(trainer.run_id)
             checkpointer.wait_until_finished()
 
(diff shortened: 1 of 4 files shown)


src/levanter/main/train_lm.py
```diff
@@ -268,6 +268,9 @@ def compute_log_probs(model, example):
             checkpointer = trainer.config.checkpointer.create(trainer.run_id)
             checkpointer.wait_until_finished()
 
+    # This isn't necessary except when Levanter is run in a subprocess (as happens w/ ray)
```
percyliang, line 271: In the sense that this gets called automatically?

percyliang (approved):

src/levanter/main/train_lm.py
dlwh: yes

PR #802, opened 2024-11-13
Nit: typing

jennifgcrl:
Fixes a silly type checker complaint. (plus black wanted to reformat some things?)


dlwh (approved):
weird. we run pre-commit in CI, but sometimes stuff slips through

dlwh:
(thanks!)

PR #803, opened 2024-11-13
Support multiple supervisde evals, some cleanup around that

dlwh:


src/levanter/data/sharded_datasource.py
```diff
@@ -184,31 +184,6 @@ def gcs_glob(pattern: str) -> list[str]:
     return matching_urls
 
 
-def datasource_from_chat_jsonl(
```
dlwh, line 187: this just moved to text

src/levanter/data/text.py
```diff
@@ -722,28 +796,69 @@ def _prepare_supervised_example(ex: dict, tokenizer: PreTrainedTokenizerBase, Po
...
+
+                tags = [name] + tags
+            else:
+                name = "default"
+            out[name] = (ds, tags)
+
+    return out
+
+
+def mk_supervised_dataset(
```
TheQuantumFractal, line 852: Seems like we are replacing this with mk_supervised_datasets? Do we still need this function then?

dlwh: it's still used in sft.py atm but i'd like to remove it

ahmeda14960 (approved):
Overall LGTM!  I think it's probably better that we get this merged into main, and I can change the revise sft PR accordingly.

src/levanter/data/text.py
TheQuantumFractal: Sounds good

PR #811, opened 2024-11-19
document WSD-S stuff, add cycles

dlwh:


src/levanter/optim/config.py
```diff
@@ -20,17 +20,21 @@
...
+    weight_decay: float = 0.1
 
     min_lr_ratio: float = 0.1
-    warmup_ratio: Optional[float] = None  # Deprecated. fraction of training steps to use as warmup
     """The lr scheduler operates on 4 stages: [warmup] - {[stable] - [decay]} x haps - [cooldown]"""
     warmup: float = 0.01
     """fraction of training steps to use as warmup, or steps to use. 0.0 means no warmup"""
     stable: float = 0.00
-    """fraction of training steps to use as stable, or steps to use. 0.0 means no stable"""
-    cooldown: float = 0.0
```
blahBlahhhJ, line 32: so we don't want a cooldown phase any more?

src/levanter/optim/config.py
```diff
@@ -20,17 +20,21 @@
...
     """fraction of training steps to use as cooldown, or steps to use. 0.0 means no cooldown"""
+    rewarmup: float = 0.0
+    "If using a cycle, how much of the cycle to use as re-warmup. 0.0 means no re-warmup."
+    cooldown: Optional[float] = None
+    """Deprecated, as its semantics are confusing."""
+    cycles: int | None = None
```
blahBlahhhJ, line 35: This is just another way to do wsds with equal cycle length correct?

src/levanter/optim/config.py
```diff
@@ -20,17 +20,21 @@
...
+    """ Number of cycles to use. If None or 1, use a single cycle. Overriden by haps."""
+
     lr_schedule: str = "cosine"  # constant, cosine, linear
     haps: Optional[list[int]] = None
```
blahBlahhhJ, line 39: Should we refactor this to some other name? cuz in the paper switched to wsd-s?

src/levanter/optim/config.py
dlwh: yes, it's just simpler

src/levanter/optim/config.py
dlwh: cooldown was redundant I think. it decayed from the final lr of the decay phase (which was almost always .1 max lr) to ... .1 max lr. So it was useless except when using constant lr or inv sqrt. constant is covered by stable, we don't really use inv sqrt

src/levanter/optim/config.py
dlwh: yeah. that's a good idea. do you have a name in mind?

blahBlahhhJ: should we just combine into the `cycles` variable? Do something like "if it's a list then... if it's an int then..."

dlwh: that's reasonable

src/levanter/optim/config.py
```diff
@@ -20,20 +20,24 @@
...
     """fraction of training steps to use as cooldown, or steps to use. 0.0 means no cooldown"""
+    rewarmup: float = 0.0
+    "If using a cycle, how much of the cycle to use as re-warmup. 0.0 means no re-warmup."
+    cooldown: Optional[float] = None
+    """Deprecated, as its semantics are confusing."""
+    cycles: int | None | list[int] = None
+    """ Number of cycles to use. If None or 1, use a single cycle. Overriden by haps."""
+
     lr_schedule: str = "cosine"  # constant, cosine, linear
     haps: Optional[list[int]] = None
```
blahBlahhhJ, line 39: mark as deprecated or just delete this? it's fairly new and I don't think people are using it.

blahBlahhhJ (approved):

src/levanter/optim/config.py
```diff
@@ -139,27 +143,39 @@ def mask_fn(model):
...
-            warmup = optax.linear_schedule(0.0, self.learning_rate, warmup_steps)
-            schedules.append(warmup)
-            boundaries.append(warmup_steps)
+        previous_end = 0.0
 
-        for start, end in zip(self.haps[:-1], self.haps[1:]):
+        for cycle, (start, end) in enumerate(zip(cooldown_points[:-1], cooldown_points[1:])):
             cycle_steps = end - start
+            if cycle == 0:  # warmup
+                warmup_steps = _convert_ratio_or_steps(self.warmup, cycle_steps)
```
blahBlahhhJ, line 168: should this be `total_steps` or `cycle_steps`?

blahBlahhhJ: I think it's easier to handle `cycle_steps` cuz it ensures `warmup_steps <= cycle_steps`. But I also think `self.warmup` is a global schedule thing not a cycle schedule thing.

blahBlahhhJ: makes sense.

PR #813, opened 2024-11-20
Fix lr schedule

blahBlahhhJ:


dlwh:
oops yeah definitely that was the problem

PR #815, opened 2024-11-20
fix for token bug that skips EOS

ahmeda14960:
This will actually add EOS if the tokenizer doesn't have it (HF is not good for this)


src/levanter/data/text.py
```diff
@@ -196,7 +196,7 @@ async def get_batch(self, indices: Sequence[int]) -> Sequence[T_co]:
         len = await self.wait_until_len_at_least(max(indices) + 1)
         if len is not None and len < max(indices) + 1:
             raise ValueError("Requested indices beyond the end of the dataset")
-        offsets = np.array(indices, dtype=np.int64) * self.seq_len
```
dlwh, line 199: this needs to be np.int64

ahmeda14960: Sorry not sure how I did this, I just fixed it.

dlwh (approved):

PR #827, opened 2024-12-02
Add segment ids/sequence masking

dlwh:
Should have done this a long time ago, but it didn't seem that important.

Turns out it's a small but non-zero win:

image


blahBlahhhJ (approved):
LGTM

PR #841, opened 2024-12-12
switch to doing lm eval harness evals in bf16

dlwh:


src/levanter/eval_harness.py
```diff
@@ -615,12 +631,12 @@ def log_report_to_tracker(prefix: str, report: dict, tracker: Optional[levanter.
...
 
 
-def lm_eval_harness(config: LmEvalHarnessConfig, tokenizer, EvalBatch, axis_resources):
+def lm_eval_harness(config: LmEvalHarnessConfig, tokenizer, EvalBatch, axis_resources, mp: jmp.Policy | None):
     tasks_to_run = config.to_task_dict()
 
     def lm_eval_harness(step: StepInfo, force=False):
-        if step.step == 0 and not force:
-            return  # don't run eval on the first step
+        # if step.step == 0 and not force:
```
ahmeda14960, line 638: did you mean to comment this out?

ahmeda14960 (approved):
you should double check the commented out stuff, but other than that LGTM

src/levanter/eval_harness.py
dlwh: oops

PR #842, opened 2024-12-13
Fix grad accum when using `loss_mask`

Aphoh:
Implement the fix described [here](https://unsloth.ai/blog/gradient). I don't think this is the cleanest way to do it, but my idea was to make a mixin for batch elements that allow them to specify when they have a specific `num_elements` that the microbatch losses should be divided by.

### DIFF
--- a/src/levanter/models/loss.py
+++ b/src/levanter/models/loss.py
@@ -1,4 +1,5 @@
 import functools
+import logging
 from typing import Optional
 
 import equinox
@@ -8,6 +9,9 @@ import jax.numpy as jnp
 import haliax as hax
 from haliax import NamedArray
 from haliax.nn import cross_entropy_loss_and_log_normalizers
+
+
+logger = logging.getLogger(__name__)
 
 
 def maybe_fused_next_token_loss(
@@ -20,6 +24,7 @@ def maybe_fused_next_token_loss(
     loss_mask: Optional[NamedArray] = None,
     reduction: Optional[hax.ReductionFunction] = hax.mean,
     reduction_axis: Optional[hax.AxisSelection] = None,
+    batch_num_elements: Optional[int] = None,
     logsumexp_weight: Optional[float] = None,
     block_size: Optional[int] = None,
     dtype: Optional[jnp.dtype] = jnp.float32,
@@ -36,6 +41,7 @@ def maybe_fused_next_token_loss(
         loss_mask (Optional[NamedArray]): Mask to apply to the loss.
         reduction (Optional[hax.ReductionFunction]): Reduction function.
         reduction_axis (Optional[hax.AxisSelection]): Axis to apply reduction.
+        batch_num_elements (Optional[int]): The number of elements in the batch. When passed, it is used to reduce the loss.
         logsumexp_weight (Optional[float]): Weight for logsumexp penalty.
         block_size (Optional[int]): Size of each block for processing.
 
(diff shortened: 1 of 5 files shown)


src/levanter/models/loss.py
```diff
@@ -20,6 +24,7 @@ def maybe_fused_next_token_loss(
     loss_mask: Optional[NamedArray] = None,
     reduction: Optional[hax.ReductionFunction] = hax.mean,
     reduction_axis: Optional[hax.AxisSelection] = None,
+    batch_num_elements: Optional[int] = None,
```
dlwh, line 27: imho i think the thing to do is to use reduction=None inside grad accum and do the reduction separately. Also, can't you infer batch_num_elements from loss_mask.sum?

src/levanter/trainer.py
```diff
@@ -549,18 +549,28 @@ def obj_fun(trainable_model):
...
+        reduce = ReductionType.MEAN
+        if isinstance(batch, NumElementsBatch) and mbs != self.TrainBatch.size:
+            batch_kwargs[
+                "batch_num_elements"
+            ] = batch.num_elements()  # tell the loss function how many elements are in the batch
+            batch_kwargs[
+                "reduction"
+            ] = hax.sum  # the loss fn should sum the loss and divide by the number of elements, not average
+            reduce = ReductionType.SUM  # we're already normalizing the loss
         grad_fn = microbatched(
```
dlwh, line 564: let's have the grad_fn return both the total (masked) loss and the number of elements (and the gradient?). WDYT?

dlwh (commented):
thanks for tackling this! I think we can push more of the logic into grad_accum instead and that will be a bit cleaner. WDYT?

src/levanter/models/loss.py
Aphoh: @dlwh We can't get that from `loss_mask.sum` since the `loss_mask` we get is for the microbatch and not the whole batch. Instead of computing a mean loss per micro batch an averaging the mean losses, we need to sum all the microbatches dividing by the true number of losses in the batch.

src/levanter/trainer.py
Aphoh: So we'd change the signature of `ComputeLossFunction` to return something like
`(model, input) -> (loss, extras)` where `extras` has the number of elements in the microbatch in it?

I also find myself wanting to log things this way (by returning elements in an `extras` struct) so this would be a nice way to add that in.

The only concern is that if the number of microbatch elements or the losses are particularly large we could lose precision in the value of the loss right? 
When computing `(loss1 + loss2 + ... ) / num_elems` v.s. `loss1/num_elems + loss2/num_elems + ...`, I would kinda expect the second to be more accurate. Is this a reasonable concern or am I prematurely optimizing?

Aphoh: Turns out the concern about precision is valid in certain regimes for fp16 but probably not for fp32

image

I'll go ahead and try to work this more into the loss functions/grad accum!

dlwh: there's actually a standard way of dealing with this which I've encapsulated as [RunningMean](https://github.com/stanford-crfm/levanter/blob/07b3f1639ec4b2f05779733109a2906378e1af5a/src/levanter/utils/stat_utils.py#L13). The basic idea is to maintain the mean and the count rather than the sum and the count. This solves most precision problems usually. (You can actually do variance too but we don't need it)

Aphoh: @dlwh I tried to encapsulate this in this last commit. I'm pretty sure it doesn't work right since it doesn't pass the `grad_accum` tests (trees don't match). Could you give it a quick peek and let me know if you see anything?

Aphoh:
passing tests, I think this is ready for a peek @dlwh

PR #848, opened 2025-01-02
Fix Transformers seq length override 

ahmeda14960:
Right now we are unable to override the base config loaded from transformers models when training. This normally isn't an issue but causes OOM when doing SFT fine-tuning and using Llama 3 8B base since the model config has a very large context length of 131K tokens.

This might not be the most elegant solution but has fixed the problem AFAIK.


dlwh (approved):
this is how it should be, thanks

PR #850, opened 2025-01-07
Add sequence packing for lm-eval-harness

dlwh:
Speedup was less than I hoped, but I think that can be improved with better packing strategies (and also removing some batching overhead. the bottleneck is now data loading??)

### DIFF
--- a/scripts/gcs_bulk_delete.py
+++ b/scripts/gcs_bulk_delete.py
@@ -1,13 +1,10 @@
 import re
 import sys
 import time
-from datetime import datetime
 
 import google.auth
 from google.api_core import operations_v1
 from google.cloud import storage_transfer_v1
-from google.type.date_pb2 import Date
-from google.type.timeofday_pb2 import TimeOfDay
 
 
 EMPTY_BUCKET = "levanter-empty"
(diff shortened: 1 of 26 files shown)


scripts/gcs_bulk_delete.py
nikil-ravi, line 1: was this intended to be part of the PR? Just checking

src/levanter/data/packing.py
```diff
@@ -0,0 +1,205 @@
+# Implements sequence packing
```
nikil-ravi, line 1: Include a brief high-level description of the algo used for packing currently?

src/levanter/data/packing.py
```diff
@@ -0,0 +1,205 @@
...
+            tokens = hax.named(ids, self.Pos)
+            segment_ids = hax.named(segment_ids, self.Pos)
+            loss_mask = hax.named(loss_mask, self.Pos)
+
+            attn_mask = AttentionMask.causal().with_segment_ids(segment_ids)
+
+            return LmExample(tokens=tokens, loss_mask=loss_mask, attn_mask=attn_mask)
+
+
+@dataclass
```
nikil-ravi, line 78:
```suggestion
@dataclass(frozen=True)
```

nikil-ravi (approved):

nikil-ravi:
Curious whether we know the rough ratio of padding:non padding tokens with this packaging strategy?

scripts/gcs_bulk_delete.py
dlwh: nah, but i don't care enough to take it out :-)

dlwh:
It's about 90% real tokens (as opposed to ~90% padding before)

PR #853, opened 2025-01-09
Added EMA of running weights to the current codebase

WhenWen:
The syntax works like 
'''
--trainer.use_ema True
--trainer.ema_beta 0.995
'''

The current issue is that the current code isn't compatible with the previous version of trainer state.


dlwh:
imho the ema stuff should go in the optimizer state, wdyt?

WhenWen:
Sounds reasonable to me. This should improve compatibility as well. I will try to implement this.

WhenWen:
After trying the implementation, implementing EMA within optimizers seems to limit the generality of this feature. Currently, I have to define specific optimizers like AdamWithEMA while this optimizer is intrinsically the same as Adam. I feel that perhaps it is a better idea to keep the EMA checkpoint in the trainer state because of this.

dlwh:
i see what you're saying, but like, Optax doesn't really have an "adam" right? It's a bunch of transformations and this is just one more? Let me see

dlwh:
closing in favor of #862

PR #863, opened 2025-01-22
add mup configuration

blahBlahhhJ:


src/levanter/main/train_lm.py
```diff
@@ -167,6 +168,41 @@ def main(config: TrainLmConfig):
 
         state = trainer.initial_state(training_key, model_init=lambda: config.model.build(Vocab, key=model_key))
 
+        if config.model.use_mup:
```
dlwh, line 171: if/when we merge this, let's clean htis up (put it in its own file, make the mup_axes configurable, etc.

PR #866, opened 2025-01-24
Allow for tokenizers/preprocessors to change batch size

Aphoh:
Small  change that keeps changes how the counting for the number of batches is done. Instead of assuming $n$ examples from shard iterator means $n$ examples after tokenizing/processing, it gets the length from the output of the tokenizer. I had a use case that involved combining multiple samples together during the processing stage, and ran into a bug where the number of batches stored in the offsets was incorrect.

### DIFF
--- a/src/levanter/store/cache.py
+++ b/src/levanter/store/cache.py
@@ -1375,9 +1375,9 @@ def _tokenize_one_shard_group(
             tokenized = _canonicalize_batch(tokenized)  # type: ignore
             this_prepared = writer._tree_store.batch_preparer(tokenized)
 
-            this_batch_size += len(batch)
-            rows_this_shard += len(batch)
-            total_rows += len(batch)
+            this_batch_size += len(tokenized)
+            rows_this_shard += len(tokenized)
+            total_rows += len(tokenized)
 
             if prepared_batch is None:
                 prepared_batch = this_prepared


dlwh:
thanks for this. I think it's not quite right in the case of preemption. In particular, we use open_shard_at_row with this value when we resume tokenization. What we would need to do is save both "rows_in" and "rows_out" and only use rows_in for open_shard_at_row. Does that make sense?

Aphoh:
@dlwh ah yeah totally! I'll try to get that worked in.

src/levanter/store/cache.py
```diff
@@ -473,7 +473,11 @@ def _monitor_metrics(self):
 class CacheLedger:
     # NB: unlike the old cache, the mere existence of a ledger doesn't mean the cache is finished
     total_num_rows: int
-    shard_rows: Dict[str, int]
+    """Number of outputted rows in the cache"""
+    shard_rows_in: Dict[str, int]
```
dlwh, line 477: sorry, can we rename this one back to just shard_rows (leave comment) so that we don't invalidate all other caches

dlwh (commented):
sorry, one small rename then i'm happy

src/levanter/store/cache.py
Aphoh: Ah gotcha yeah... Do I need to do some other logic to make `shard_rows_out` an optional key during deserialization?

dlwh: i'm pretty sure if you give it a default value it will be fine. I'll check it against a cache before merging

PR #868, opened 2025-01-29
Supporting varied mixtures over training

kothasuhas:
## Description
Currently, LM mixture dataset can only handle a static mixture over the course of training. This PR enables varying this mixture over datasets over the course of training. The user can now specify a list of stages and the sequence index at which each should start. 

Internally, we identify a training block to its stage, which defines its mixing weights. To efficiently translate a data point's index within a block to a respective source dataset, we precompute prefix sums that track how many data points are seen by previous stages.

## Fixes Issues
https://github.com/stanford-crfm/marin/issues/81

## Unit test coverage
There are new unit tests in `test_varying_mixture.py` to ensure that the varying mixture behaves as expected.

## Known breaking changes/behaviors
The design enables traditional usage of the MixtureDataset class. However, some of the private quantities are different (i.e. the expected counts per block now depends on the block and is not a member variable). To my knowledge, these variables are not accessed outside of tests.

## Additional context
I have some changes I want to make Marin to enable usage of this new functionality, though these updates are modular and can be seperate PR's. I have spot-checked that training proceeds as expected with this test. This is my first PR so feedback is appreciated :))

### DIFF
--- a/src/levanter/data/mixture.py
+++ b/src/levanter/data/mixture.py
@@ -183,7 +183,7 @@ class MixtureDataset(AsyncDataset[T]):
         stage_starts = np.array([start for start, _ in self.weight_stages])
         return max(0, np.searchsorted(stage_starts, block_start, side="right") - 1)
 
-    @alru_cache
+    @alru_cache(maxsize=32)
     async def _get_block(self, index: int) -> Optional[np.ndarray]:
         stage = self._get_stage_for_block(index)
         if not self.randomize_blocks:
--- a/src/levanter/data/mixture.py
+++ b/src/levanter/data/mixture.py
@@ -1,6 +1,6 @@
 import asyncio
 import warnings
-from typing import Mapping, Optional, Sequence, TypeVar
+from typing import List, Mapping, Optional, Sequence, Tuple, TypeVar
 
 import jax
 import numpy as np
@@ -30,12 +30,12 @@ class MixtureDataset(AsyncDataset[T]):
     according to the weights.
 
     Creating a random-access MixtureDataset is challenging because we need to keep track of the current index of each
-    dataset. So solve this, we instead use "block-deterministic" mixtures, where the number of samples from each dataset
-    in each block is always identical (and we shuffle the order of the dataset ids in each block).
+    dataset. To solve this, we instead use "block-deterministic" mixtures, where the number of samples from each dataset
+    in each block is always identical (and we shuffle the order of the dataset ids in each block). To handle the case where the dataset mixture changes over time, we use a list of stages and precompute statistics to accurately compute the index of each dataset in each block.
 
     Args:
         datasets: A dict of datasets, where the key is the name of the dataset and the value is the dataset itself
-        weights: weights for each dataset
+        weights: Weights for each dataset. This can be provided in a list of stages, where each stage is a tuple of (start_index, weights).
         stop_strategy: strategy for stopping the iteration, by default RESTART_STRATEGY. (Currently only RESTART_STRATEGY is supported)
             - FIRST_STOP_STRATEGY: stop when one dataset has been exhausted
             - ALL_STOP_STRATEGY: stop when all datasets have been exhausted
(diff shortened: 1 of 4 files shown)


src/levanter/data/mixture.py
```diff
@@ -30,12 +30,12 @@ class MixtureDataset(AsyncDataset[T]):
...
     Creating a random-access MixtureDataset is challenging because we need to keep track of the current index of each
-    dataset. So solve this, we instead use "block-deterministic" mixtures, where the number of samples from each dataset
-    in each block is always identical (and we shuffle the order of the dataset ids in each block).
+    dataset. To solve this, we instead use "block-deterministic" mixtures, where the number of samples from each dataset
+    in each block is always identical (and we shuffle the order of the dataset ids in each block). To handle the case where the dataset mixture changes over time, we use a list of stages and precompute statistics to accurately compute the index of each dataset in each block.
 
     Args:
         datasets: A dict of datasets, where the key is the name of the dataset and the value is the dataset itself
-        weights: weights for each dataset
+        weights: Weights for each dataset. This can be provided in a list of stages, where each stage is a tuple of (start_index, weights).
```
dlwh, line 38:
```suggestion
        weights: Weights for each dataset. This can be provided in a list of stages, where each stage is a tuple of (start_step, weights).
```

src/levanter/data/mixture.py
```diff
@@ -78,25 +99,39 @@ def __init__(
 
         self.stop_strategy = stop_strategy
 
-        self._counts_per_block = self._compute_expected_counts_per_block(block_size)
-        # precompute a list of ids for each block
-        # the ids contain both the dataset index and the index within the dataset
-        self._unpermuted_ids = self._compute_unpermuted_ids(self._counts_per_block)
+        # Compute counts and unpermuted_ids for each stage
```
dlwh, line 102: maybe pull this into a method

src/levanter/data/mixture.py
```diff
@@ -209,9 +259,6 @@ async def getitem_async(self, index: int) -> T:
         return await dataset.getitem_async(dataset_index)
 
     async def _remap_indices(self, ds, indices_into_ds):
-        """
-        Handles wrap around for datasets that have finite length
-        """
```
dlwh, line 214: revert? or maybe you don't like that comment?

src/levanter/data/mixture.py
```diff
@@ -167,8 +217,8 @@ async def get_batch(self, indices: Sequence[int]) -> Sequence[T]:
         assert len(indices) == len(blocks) == len(block_ids)
 
         for batch_index, (idx, block, block_id) in enumerate(zip(indices, blocks, block_ids)):
-            index_within_block = idx % self.block_size  # which element of the block to get
-            id = block[index_within_block]  # for this block, which dataset+base dataset offset
+            index_within_block = idx % self.block_size
+            id = block[index_within_block]
```
dlwh, line 221: revert?

src/levanter/data/text.py
```diff
@@ -1218,7 +1231,7 @@ def shuffle_ds(ds, key):
             weights=self.train_weights,
             stop_strategy=self.stop_strategy,
             key=mix_key,
-            block_size=2048,
+            block_size=self.mixture_block_size,
```
dlwh, line 1234: oops!

dlwh (approved):
awesome! thanks for knocking this out so quickly

src/levanter/data/mixture.py
kothasuhas: cursor shenanigans, reverted

src/levanter/data/mixture.py
kothasuhas: cursor shenanigans part 2, reverted

src/levanter/data/mixture.py
kothasuhas: changed. also added clarification that this corresponds to the sequence index at which you want to change the distribution, not the batch index. this method doesnt get to know the batch size and i think that generality is good (eventually if you want batch size curricula like most real LMs)

kothasuhas: actually because of this, im thinking its better to keep it as "start_sequence_index"?

src/levanter/data/mixture.py
kothasuhas: done

kothasuhas:
thanks for review! only final request is to name the variable as start_seq_index instead of start_step since step is often conflated with batch indices. and for maximum generality, i wanted this method to work without knowing the batch size. if this is good, ill merge

dlwh:
awesome thanks!

PR #873, opened 2025-01-30
Actually do sequence packing

ahmeda14960:


src/levanter/data/packing.py
```diff
@@ -111,9 +111,7 @@ def pack_prompt_completions(
     """
     Packs a list of prompt completions into LmExamples using the SequencePacker
     """
-
```
dlwh, line 114: preference would be to revert changes to this file

src/levanter/data/text.py
```diff
@@ -975,6 +975,38 @@ def preprocess_chat_example(batch, tokenizer: PreTrainedTokenizerBase, should_ap
...
+    source = config.get_shard_source("train")
+    if source is None:
+        raise ValueError("No training data source found")
+
+    # Set up example structure matching supervised case
+    output_exemplar = {"input_ids": np.zeros((0,), dtype=np.int32), "sources_len": np.zeros((0,), dtype=np.int32)}
+
+    input_ids = tokenizer("hi there")["input_ids"]
+    should_append_eos = input_ids[-1] != tokenizer.eos_token_id
+    logger.info(f"Manual EOS Needed: {should_append_eos}")
```
dlwh, line 991: rm?

src/levanter/main/sft.py
```diff
@@ -216,7 +252,98 @@ def train(config: SFTConfig):
...
+        An iterator yielding PromptCompletion objects
+    """
+    # AsyncDataset already has a current_len method that returns current length or None
+    length = asyncio.run(cached_dataset.async_len())
+
+    if length is None:
+        raise ValueError("Dataset length cannot be None")
+
+    for batch_indicies in batched(range(length), 4096):
+        examples = asyncio.run(cached_dataset.get_batch(batch_indicies))
```
dlwh, line 276: indices

src/levanter/main/sft.py
```diff
@@ -216,7 +252,98 @@ def train(config: SFTConfig):
...
+            ids = example["input_ids"].tolist()
+            if len(ids) > Pos.size:
+                ids = ids[: Pos.size]
+
+            if len(ids) <= sources_len:
+                continue
+
+            try:
+                yield PromptCompletion(ids=ids, prompt_length=sources_len, segment_id=batch_indicies[i])
+            except ValueError:
```
dlwh, line 293: what's erroring here? let's at least log these

src/levanter/main/sft.py
```diff
@@ -216,7 +252,98 @@ def train(config: SFTConfig):
...
+        Pos,
+        prompt_completion_iterator,
+        max_segments_per_example=max_pack_size,
+        pad_token=tokenizer.pad_token_id,
+        max_buffered_examples=16,
+    )
+
+
+"""
+Helper function to create a dummy instance with the same shape as the batch.
```
dlwh, line 311: can you relocate / combine these with the other versions?

src/levanter/data/text.py
ahmeda14960: I removed the check for the padding token, but we still need this because if the cached dataset doesn't exist then I want to tokenize and cache it and I would need to check if we should append an EOS

src/levanter/main/sft.py
ahmeda14960: changed

src/levanter/main/sft.py
ahmeda14960: Added some logging

src/levanter/data/text.py
dlwh: i just meant the but fair enough

src/levanter/main/sft.py
dlwh: they can be top level, putting them in data loader seems good to me. I need to add dummy instances to normal training as well

ahmeda14960: Okay done

dlwh:
@ahmeda14960 can you fix unit tests (may just need to move/rename configs)

PR #874, opened 2025-02-01
Refactor `run_on_pod` actors and cancellation logic

cathyzbn:
Manual Worker Termination on central 2: http://localhost:8265/#/jobs/98010000


src/levanter/infra/ray_tpu.py
```diff
@@ -448,92 +384,27 @@ def _handle_ray_error(tpu_info: _TpuInfo, e: RayError):
...
-        os.unlink("/tmp/libtpu_lockfile")
-    except FileNotFoundError:
-        pass
-    except PermissionError:
-        logger.warning("Failed to remove lockfile")
-        try:
-            os.system("sudo rm /tmp/libtpu_lockfile")
-        except Exception:  # noqa
-            pass
+# TODO(cathy): this is unused, do we need it?
```
dlwh, line 387: oh yeah i don't think we need it anymore. I had it before we started forking a training process

src/levanter/infra/ray_tpu.py
```diff
@@ -63,73 +63,85 @@ class TpuFailed(_TpuRunResult):
...
+                        ray.kill(actor)
+                    except Exception:
+                        logger.exception(f"Failed to kill slice actor {actor}")
+                self.slice_actors = None
+            ray.actor.exit_actor()
+    
+    return TPUHeadNodeActor
+
+def make_tpu_slice_actor(tpu_type: str):
+    @ray.remote(resources={f"TPU-{tpu_type}-slice": 1})
```
dlwh, line 129: this isn't a thing. the relevant resource types are:

* `TPU-{tpu_type}-head`: special resource given to the 0th worker of a tpu pod slice
* `{name of tpu slice}`: special resource given to all workers in a particular tpu slice
* (plus the TPU resources for a given tpu worker)

So the types of actors are:

* 1 head node actor for each tpu slice
* K worker actors owned by the head node actor

(A multislice job will have multiple HeadActors)

src/levanter/infra/ray_tpu.py
```diff
@@ -138,61 +150,53 @@ def do_run(self, remote_fn, coordinator_ip, slice_id, num_slices) -> _TpuRunResu
...
+            self.queue = None
+
+            if success:
+                return value
+            else:
+                value.reraise()  # Reraise the exception with original traceback
 
-    coordinator_ip = slice_infos[0][2]
+        def kill(self):
+            if self.process is not None:
```
dlwh, line 189: add logging here

dlwh (commented):
yeah this is basically the right direction. a few thoughts

PR #877, opened 2025-02-03
ray_tpu: Catch preemption better, don't hang if there's a c level abort

dlwh:


kothasuhas:
When I run this on eu west 4b, I find that the job terminates by itself! Specifically, it launches the job, catches the issue at the end of training, retries it 4 times, and eventually succeeds.

Job log: http://localhost:8265/#/jobs/bf000000

kothasuhas (approved):

dlwh:
well, i'll take it. still don't understand why it's crashing at all but crashes are better than hangs

PR #880, opened 2025-02-05
add support for BatchSchedules

dlwh:


src/levanter/schedule.py
```diff
@@ -0,0 +1,122 @@
...
+        if schedule[i].until < schedule[i - 1].until:
+            if i == len(schedule) - 1 and schedule[i].until == -1:
+                continue
+            raise ValueError(f"Schedule is not sorted at index {i}")
+
+
+@dataclass
+class BatchSegment:
+    start: int  # The training step at which this batch size starts.
+    until: int | float  # The training step at which this batch size stops. If -1 or inf, the segment is open-ended.
```
Helw150, line 47: Maybe support just -1 or inf for simplicity sake.

src/levanter/schedule.py
```diff
@@ -0,0 +1,122 @@
...
+
+    This class also tracks cumulative data offsets and can return the "local" indices for a step and a sharding
+    """
+
+    def __init__(self, schedule: int | IntSchedule):
+        self.schedule = schedule
+
+        # Precompute the segments of the schedule.
+        if isinstance(schedule, int):
+            self.segments = [BatchSegment(0, BIG_INT, schedule, 0)]
```
Helw150, line 66: Why BIG_INT and not Inf here? Given the lower level class lists support for inf, seems odd (while unlikely to ever come up) to use this internally.

src/levanter/schedule.py
```diff
@@ -0,0 +1,122 @@
...
+                    prev_until = until
+
+    def batch_size_at_step(self, step: int) -> int:
+        """
+        Return the batch size (number of samples) at the given training step.
+        """
+        for seg in self.segments:
+            if seg.start <= step < seg.until:
+                return seg.value
+        # If the step is beyond all defined segments, use the last segment's batch size.
```
Helw150, line 94: Feels worth throwing a warning here.

src/levanter/schedule.py
Helw150: Doubly so with the BIG_INT mixed in. Some calls support until being natively -1, while the below overrides -1 to BIG_INT and then assumes that in the logic.

src/levanter/data/loader.py
```diff
@@ -31,23 +33,60 @@
...
+        of the data. In general, each machine only loads the data that it needs to process.
 
-        TODO: document this
+        (At the moment, it's not smart enough to load parts of examples if examples are sharded across machines,
+        but whole examples are handled correctly.)
 
         Args:
-            Batch (hax.Axis): The batch axis
+            batch_axis_name (str | Axis): The name of the batch axis
+            batch_size (int | IntSchedule | None): The size of the batch or a schedule for the size of the batch
```
Helw150, line 89: Hmmm, I think this risks being confusing since batch_axis_name sometimes sneakily contains the batch_size too.

Is it to un-haliax-y to only support `batch_axis_name (str)` and `batch_size (int | IntSchedule)`?

src/levanter/data/loader.py
```diff
@@ -61,40 +100,51 @@ def __init__(
         self.axis_resources = axis_resources
         self.data_store = data
         self.mesh = mesh
-        self.Batch = Batch
+
+        if batch_size is None:
+            assert isinstance(batch_axis_name, hax.Axis)
+            self.batch_axis_name = batch_axis_name.name
```
Helw150, line 106: This is what led me to the comment above - just feels yucky enough to do `...name.name` that it could possibly be confusing as well down the line.

src/levanter/data/loader.py
```diff
@@ -223,25 +285,64 @@ def get_data(indices):
...
-            indices_this_batch_this_process = [indices_this_batch[i] for i in self.dl._local_indices]
-            indices_for_this_batch_of_batches.extend(indices_this_batch_this_process)
+            global_offset = self.dl.scheduler.global_data_offset_by_step(bn)
+            offsets.append(global_offset)
+            local_indices_for_device = self.dl.local_data_indices_by_device_for_step(bn)
+
+            distinct_local_indices_this_batch = set()
+            for indices in local_indices_for_device.values():
+                for local_index in indices:
+                    distinct_local_indices_this_batch.add(local_index)
```
Helw150, line 315: What are instances where local_indices_for_device would have duplicates and it isn't a bug?

src/levanter/schedule.py
dlwh: oh yeah i was gonna support inf but it got annoying so I started to remove it but didn't finish

src/levanter/schedule.py
dlwh: was switching to big_int

src/levanter/data/loader.py
dlwh: eh, i thought about this. it was kind of an attempt at backwards compat but I guess there's no reason

src/levanter/data/loader.py
dlwh: tensor parallel, sequence parallel

Helw150: Got it - makes sense

src/levanter/data/loader.py
dlwh: i'll just commit

dlwh: err, int | IntSchedule | Axis

Helw150:
LGTM assuming tests all go through!

Helw150 (approved):

PR #882, opened 2025-02-09
Support Irregular/nondivisible batch sizes by padding

dlwh:
Also add support for padding the final batch

We use zeros for this padding, which works out b/c a 0's loss_mask means no contribution to the loss

### DIFF
--- a/src/levanter/data/loader.py
+++ b/src/levanter/data/loader.py
@@ -176,8 +176,8 @@ class DataLoader(Iterable[Ex]):
         out = _round_to_nearest_multiple(size, self._data_axis_size)
 
         warnings.warn(
-            f"Rounding batch size {size} to {out} to be divisible by data axis size {self._data_axis_size}."
-            f"\n This results in a per-device batch size of {out // self._data_axis_size}"
+            f"Padding batch size {size} to {out} to be divisible by data axis size {self._data_axis_size}."
+            f"\n This results in a per-device batch size of {out // self._data_axis_size}. (Extra data is zeros.)"
         )
 
         return out
--- a/src/levanter/data/loader.py
+++ b/src/levanter/data/loader.py
@@ -106,7 +106,7 @@ class DataLoader(Iterable[Ex]):
             self.scheduler = BatchSchedule(batch_size)
 
         self._batch_sharding = hax.partitioning.sharding_for_axis(self.batch_axis_name, axis_resources, mesh)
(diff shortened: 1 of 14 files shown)


Helw150:
So I can double check, what flow should I be looking at to confirm the 0's always get ignored in the loss?

Right now, I was following:
- CausalLmDataset (ignore_index defaults to None. no explicit loss mask is constructed. Calls LmExample [here](https://github.com/stanford-crfm/levanter/blob/1927f93eb58ed5cf61044ef70e6263138ce0abfc/src/levanter/data/text.py#L247))
- LmExample (ignore_index defaults to None. if ignore_index is None and loss_mask is None, the loss mask is constructed as causal)
- next_token_loss just uses the loss mask as given

Helw150:
NVM, I see. When you create the 0's tree, this includes creating an explicit zero's loss mask.

src/levanter/data/loader.py
```diff
@@ -110,12 +106,32 @@ def __init__(
...
-            self._ex_leaves, self._ex_structure = jax.tree_flatten(
-                blocking_wait(self.data_store.getitem_async(0)), is_leaf=is_named_array
-            )
+            current_len = blocking_wait(self.data_store.current_len())
+            if current_len <= 0:
+                logger.warning("Data store currently has no data. We will block until data is available.")
+
+            initial_example = blocking_wait(self.data_store.getitem_async(0))
+            self._ex_leaves, self._ex_structure = jax.tree_flatten(initial_example, is_leaf=is_named_array)
+            self._padding_example = _make_padding_example(initial_example)
```
Helw150, line 123: I'm not following this right now - I don't see where `self._padding_example` actually gets used?

tests/test_new_loader.py
```diff
@@ -265,15 +295,59 @@ def test_loader_with_batch_scheduler(model_axis_size):
...
+                assert len(batch) == 32
+
+        # total steps: 10 * 8 + 10 * 16 = 240, (1007 - 240) // 32 = 23
+        assert step == 20 + 23
+
+        # last batch should be padded
+        assert len(batch) == 32
+        # ensure all the padded examples are all 0's
+        num_padding = 32 - (1007 - 240) % 32
+        np.all(batch[-num_padding:] == 0)
```
Helw150, line 353: Needs an assert right?

src/levanter/data/loader.py
dlwh: line 330

PR #890, opened 2025-02-23
int8 quantization from aqt

blahBlahhhJ:
Waiting for https://github.com/stanford-crfm/haliax/pull/118

TODO: need to bump haliax version in levanter denpendency.


src/levanter/trainer.py
```diff
@@ -361,6 +361,7 @@ def init_state_and_model(model_init, training_key):
                 is_trainable=is_trainable,
                 mp=self.mp,
                 fp8=self.fp8,
+                int8=self.config.int8,
```
dlwh, line 364: it'd probably be better if we made a QuantizationConfig or something

src/levanter/trainer_state.py
```diff
@@ -110,6 +117,9 @@ def init(
         if fp8 is not None:
             model = fp8_linear_layers(model, fp8)
 
+        if int8:
+            model = int8_linear_layers(model)
```
dlwh, line 121: assert not both maybe, or let's just refactor out a common config

dlwh (approved):

PR #891, opened 2025-02-25
Gcs hf

ahmeda14960:
PR to allow directly loading hf models from hf, using initialize_from_hf=True just set `model_name_or_path` equal to a gcs hf path


dlwh (approved):

dlwh:
test failure isn't your fault. not sure what is going on

PR #893, opened 2025-02-26
Less contention in tokenization

dlwh:
Hopefully avoiding the $$$ we've seen tokenizing


tests/test_newdataset.py
```diff
@@ -37,14 +37,16 @@ async def test_list_async_dataset_appends_and_finalizes_correctly():
```
RohithKuditipudi, line 37: This test seems to be for a different issue (i.e., is shuffle working) than contention in tokenization?

RohithKuditipudi (approved):

tests/test_newdataset.py
dlwh: yeah sorry it snuck in. dont' care enough to fix it.

PR #897, opened 2025-02-27
viz: add ability to compare models, change the color map etc

dlwh:
Also adds tool tip showing argmax for each position, and attempts to escape/encode newlines, and multiple spaces so weird whitespace issues get picked up

Normal viz:

image

Diff viz:

image


Helw150 (approved):
LGTM!

PR #899, opened 2025-03-03
Add SFT mixture script

ahmeda14960:
Nice script that lets us run sft over mixture of datasets. We can also emulate epoching by setting all weights to 0 except for dataset of interesting and setting num_train_steps accordingly

### DIFF
--- a/config/sft_llama3_mixture.yaml
+++ b/config/sft_llama3_mixture.yaml
@@ -86,7 +86,7 @@ trainer:
   fsdp_axis: "embed"
   batch_axis: "batch"
   checkpointer:
-    base_path: "gs://levanter-checkpoints/marin/llama_3.1_mixture_total/seed_0/"
+    base_path: "gs://marin-us-central2/checkpoints/llama_3.1_mixture_total/seed_0/"
 
 optimizer:
   learning_rate: 5e-6
@@ -96,7 +96,7 @@ optimizer:
   warmup: 0.03
 
 hf_save_steps: 1000
-hf_save_path: "gs://levanter-checkpoints/marin/llama_3.1_mixture_total/hf/seed_0/"
+hf_save_path: "gs://marin-us-central2/checkpoints/llama_3.1_mixture_total/hf/seed_0/"
 
 initialize_from_hf: True
 model_name_or_path: "meta-llama/Llama-3.1-8B"
--- a/config/sft_llama3.1_tulu3.yaml
+++ b/config/sft_llama3.1_tulu3.yaml
@@ -43,7 +43,7 @@ trainer:
   fsdp_axis: "embed"
   batch_axis: "batch"
   checkpointer:
-    base_path: "gs://levanter-checkpoints/marin/llama_3.1_tulusft/"
+    base_path: "gs://marin-us-central2/checkpoints/llama_3.1_tulusft/seed_0/"
 optimizer:
   learning_rate: 5e-6
   weight_decay: 0.0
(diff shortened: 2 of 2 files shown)


config/sft_llama3_mixture.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+optimizer:
+  learning_rate: 5e-6
+  weight_decay: 0.0
+  min_lr_ratio: 0.0
+  lr_schedule: "linear"
+  warmup: 0.03
+
+hf_save_steps: 1000
+hf_save_path: "gs://levanter-checkpoints/marin/llama_3.1_mixture_total/hf/seed_0/"
```
dlwh, line 99: don't use levanter-checkpoints anymore. prefer one of the marin buckets

src/levanter/callbacks.py
```diff
@@ -323,11 +323,31 @@ def update_pbar(step: StepInfo):
 def profile(path: str, start_step: int, num_steps: int, create_perfetto_link: bool) -> Callable[[StepInfo], None]:
     def profiler_callback_fn(step: StepInfo):
         # -1 b/c step is the finished step
+        if step.step < start_step - 1:
+            logger.info(f"Profiling scheduled to start at step {start_step}, current step: {step.step + 1}")
```
dlwh, line 327: this is kinda spammy. maybe only do it every minute or 10 steps or something?

src/levanter/callbacks.py
```diff
@@ -338,29 +358,56 @@ def profiler_callback_fn(step: StepInfo):
...
+                logger.info("Profile artifact logged successfully")
+            except Exception as e:
+                logger.error(f"Failed to log profile artifact: {e}")
+
+            logger.info("Calling barrier_sync")
+            try:
+                barrier_sync(timeout=60.0)  # Increase timeout to 60 seconds
+                logger.info("Profiling callback complete")
+            except Exception as e:
+                logger.warning(f"Barrier sync after profiling failed, but continuing training: {e}")
```
dlwh, line 392: hrm, did you hit this? this is probably a bad idea

src/levanter/callbacks.py
```diff
@@ -338,29 +358,56 @@ def profiler_callback_fn(step: StepInfo):
...
 def _flush_while_waiting(event):
     def flush_stdout():
         sys.stdout.flush()
-        time.sleep(5)
+        # Initial shorter wait to reduce lag
+        time.sleep(2)
         while not event.is_set():
-            print("Waiting...", flush=True)
-            time.sleep(5)
+            print(f"Worker {jax.process_index()} flushing logs...", flush=True)
```
dlwh, line 405: logger please

src/levanter/main/sft_mixture.py
```diff
@@ -0,0 +1,306 @@
...
+
+class DatasetType(str, Enum):
+    """Type of dataset to use"""
+
+    HUGGINGFACE = "huggingface"  # Use HF dataset
+    CHAT_JSONL = "chat_jsonl"  # Use JSONL files with chat format
+
+
+@dataclass
+class SFTMixtureConfig:
```
dlwh, line 54: i'd prefer if we had one main instead of this, but let's save that for the pre-train/sft merger

dlwh (approved):

src/levanter/callbacks.py
ahmeda14960: added by cursor, reverting

PR #900, opened 2025-03-05
RayTpuManager

cathyzbn:


src/levanter/infra/ray_tpu.py
```diff
@@ -25,8 +25,12 @@
 
 # CF https://gist.github.com/allenwang28/e3400b9e9212b50aa1cda55ebeccea60
 
+from levanter.ray_tpu.src.ray_tpu import RayTpuManager
+
 logger = logging.getLogger("ray")
 
+_tpu_manager = RayTpuManager()
```
dlwh, line 32: i think i'd rather not initialize a global here unless  it's a pretty simple state

src/levanter/infra/ray_tpu.py
```diff
@@ -65,220 +69,52 @@ class TpuRunError(_TpuRunResult):
     error: Exception
 
 
-def run_on_pod(remote_fn: RemoteFunction | Callable, tpu_type: str) -> ray.ObjectRef:
+def _cancel_future(future):
```
dlwh, line 72: rename to _cancel_subsequent_future or _cancel_future_suppress or something

dlwh: and maybe move to closer to bottom

src/levanter/infra/ray_tpu.py
```diff
@@ -65,220 +69,52 @@ class TpuRunError(_TpuRunResult):
...
-    @ray.remote(resources={f"TPU-{tpu_type}-head": 1})
+    @ray.remote
     def do_run(remote_fn) -> _TpuRunResult:
         logging.basicConfig(level=logging.INFO)
-        num_hosts = ray.util.accelerators.tpu.get_current_pod_worker_count()  # -> 4
-        remote_fn, tpu_name = _redecorate_remote_fn_for_tpu(remote_fn, num_hosts)
+        remote_fn = _forkify_remote_fn(remote_fn)
+        if not isinstance(remote_fn, RemoteFunction):
+            remote_fn = ray.remote(remote_fn)
+        handles = _tpu_manager.remote(remote_fn, topology={tpu_type: num_slices}, multislice=True)
```
dlwh, line 98: should it just be `multislice=num_slices>1`?

dlwh: handle is usually used for actors. are these actors?

src/levanter/infra/ray_tpu.py
```diff
@@ -65,220 +69,52 @@ class TpuRunError(_TpuRunResult):
...
-        futures = [remote_fn.remote() for _ in range(num_hosts)]
         try:
-            out = ray.get(futures)
+            out = ray.get(handles)
             logger.info("TPU job finished")
             return TpuSuccess(info, out)
         except RayError as e:
-            _cancel_all_futures(futures)
+            logger.exception(f"Ray error {e}. Killing futures for this slice")
+            _cancel_future(handles)
```
dlwh, line 107: if they're actor handles, you have to kill them

src/levanter/infra/ray_tpu.py
```diff
@@ -302,72 +137,44 @@ def run_on_pod_multislice_resumable(
...
             num_preemptions += 1
             logger.warning(f"Preempted {num_preemptions} times, {e}")
             continue
-        except ray.exceptions.RayTaskError as e:
-            for f in futures:
-                try:
-                    ray.cancel(f)
-                except Exception:
-                    logger.exception("Failed to kill job after primary failure")
+        except (ray.exceptions.RayTaskError, Exception) as e:
```
dlwh, line 148: if we're catching Exception we don't need TaskError right?

src/levanter/infra/ray_tpu.py
```diff
@@ -302,72 +137,44 @@ def run_on_pod_multislice_resumable(
...
+            try:
+                ray.cancel(future)
```
dlwh, line 150: i don't think we need to cancel this? it's already failed if we're here right?

dlwh (commented):
generally seems good if it works!

PR #901, opened 2025-03-10
add devcontainer config for nvidia

garymm:
Helps automate the process described in "GPU Docker Dev" on supported
editors.

### DIFF
--- a/docs/dev/GPU-Docker-Dev.md
+++ b/docs/dev/GPU-Docker-Dev.md
@@ -3,8 +3,12 @@
 This guide assumes you've gone through the Docker setup section of the [Getting Started GPU](../Getting-Started-GPU.md) guide.
 We assume you're familiar with Docker and have a basic understanding of how to use it.
 
+## Option 1: Development Container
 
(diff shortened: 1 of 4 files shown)


docs/dev/GPU-Docker-Dev.md
```diff
@@ -32,7 +36,7 @@ Now you should be able to run training jobs in this container and it will use th
 
 ```bash
 python src/levanter/main/train_lm.py \
-    --config_path config/gpt2_small.yaml
+    --config_path config/gpt2_nano.yaml
```
garymm, line 39: small requires GCP access, nano does not, so this is a lower bar. Not sure if you want to require GCP access for this step. If so perhaps we should document how to set that up inside the contatiner before this command.

dlwh: no this is great. thanks!

dlwh (approved):

dlwh:
I don't really know anything about dev containers but will trust you on this one!

PR #905, opened 2025-03-10
fix pre-commit warnings and errors

garymm:
top-level `default_stages` uses deprecated stage names (commit) which will be removed in a future version.

repo `https://github.com/pre-commit/pre-commit-hooks` uses deprecated stage names (commit, push) which will be removed in a future version.

src/levanter/store/jagged_array.py:294:64: E226 missing whitespace around arithmetic operator


dlwh:
Thanks!

PR #906, opened 2025-03-10
rewrite deprecated jax.tree_ calls

garymm:


dlwh (approved):
thanks, appreciate it

PR #908, opened 2025-03-10
set asyncio_default_fixture_loop_scope

garymm:
Fixes this warning:
```
PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
```

TBH I haven't evaluated the other options but this is going to be the
default value soom and the tests still pass :-).


dlwh:
Thanks!

PR #911, opened 2025-03-12
Push CausalLM creation into DatasetConfig, fix some breakages from HF api changes

dlwh:
Also removed forgetful causal masking for now since we never use it

### DIFF
--- a/src/levanter/data/text.py
+++ b/src/levanter/data/text.py
@@ -20,7 +20,6 @@ import numpy as np
 import regex
 import tensorstore as ts
 from draccus import field
-from jax.random import PRNGKey
 from jaxtyping import PRNGKeyArray
 from tokenizers import normalizers
 
@@ -30,7 +29,6 @@ from haliax import Axis
 from levanter.data import AsyncDataset
 from levanter.data.dataset import MappedAsyncDataset
 from levanter.data.mixture import MixtureDataset, StopStrategy, rescale_mixture_schedule_for_batch_schedule
-from levanter.models.attention import AttentionMask
 from levanter.models.lm_model import LmExample
 from levanter.schedule import BatchSchedule
 from levanter.store.cache import CacheOptions, TreeCache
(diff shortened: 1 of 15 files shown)


dlwh:
Also, deleting epoch stuff from train_lm since it was getting in the way and i don't think really did what we wanted.

tests/test_llama.py
```diff
@@ -73,7 +72,7 @@ def test_llama_rotary_embedding():
     levanter_emb = DefaultRotaryEmbeddingsConfig().build(HeadSize=HeadSize, Pos=Pos)
     levanter_output = (levanter_emb.cos, levanter_emb.sin)
 
-    hf_rope = HFLlamaRotaryEmbedding(dim=hidden_dim, max_position_embeddings=seq_len, device=device)
+    hf_rope = HFLlamaRotaryEmbedding(config=llama_config.to_hf_config(32000), device=device)
```
dlwh, line 75: HF broke api, as they do

dlwh:
tests are failing because HF is having trouble

src/levanter/data/text.py
```diff
@@ -1046,21 +1026,23 @@ class LMDatasetConfig(LMDatasetSourceConfig, LMTaskConfig):
...
-        ds: AsyncDataset[np.ndarray] | None = self.token_seq_dataset("train", seq_len, monitors)
+        ds: AsyncDataset[np.ndarray] | None = self.token_seq_dataset("train", Pos.size, monitors)
 
         # add epoch flag here.
         if ds is None:
             raise ValueError("No training set!")
 
+        assert ds is not None
+
         if epochs:
```
ahmeda14960, line 1046: shouldn't we also get rid of the if epochs check here?

ahmeda14960 (approved):

src/levanter/data/text.py
dlwh: meh i'm fine with it, but probably

Issue #907, opened 2025-03-10
add Hyena model architecture


garymm:
@dlwh mentioned this may be useful for some users, so filing an issue. I plan to work on this.

Paper: [Hyena Hierarchy: Towards Larger Convolutional Language Models](https://arxiv.org/abs/2302.10866)
Official implementation in PyTorch: https://github.com/HazyResearch/safari

PR #913, opened 2025-03-16
Add Hyena Operator and GPT2 Hyena

garymm:
HyenaOperator is an alternative to Attention.
Gpt2Hyena is like Gpt2, but using HyenaOperator instead of attention.

Fixes: #907

### DIFF
--- a/src/levanter/models/gpt2_hyena.py
+++ b/src/levanter/models/gpt2_hyena.py
@@ -1,0 +1,229 @@
+"""The GPT2 architecture, but with Hyena instead of Attention / Transformer."""
+
+import dataclasses
+from dataclasses import dataclass
+from typing import Dict, Optional, Type
+
+import equinox as eqx
+import jax.random as jrandom
+from jaxtyping import PRNGKeyArray
+
+import haliax as hax
+import haliax.jax_utils
+import haliax.nn as hnn
+from haliax import Axis, NamedArray
+from haliax.jax_utils import named_call, shaped_rng_split
+from haliax.nn.scan import Stacked
+from haliax.state_dict import ModuleWithStateDictSerialization
+
+from levanter.models.attention import AttentionMask
+from levanter.models.gpt2 import Gpt2Embeddings, Gpt2Mlp
+from levanter.models.hyena import HyenaConfig, HyenaOperator
+from levanter.models.lm_model import LmConfig, LmHeadModel
+from levanter.utils.activation import ActivationFunctionName
+
+
+@LmConfig.register_subclass("gpt2_hyena")
+@dataclass(frozen=True)
+class Gpt2HyenaConfig(LmConfig):
+    seq_len: int = 1024
+    hidden_dim: int = 768
+    num_layers: int = 12
+    num_heads: int = 12
+
+    # how much to scale the embedding dim for the mlp layer
+    mlp_scale: int = 4
+
+    initializer_range: float = 0.02
+    # dropout doesn't really help so we 0 it out by default
+    embed_pdrop: float = 0.0
+    resid_pdrop: float = 0.0
+    hyena_pdrop: float = 0.0
+    layer_norm_epsilon: float = 1e-5
+    activation_function: ActivationFunctionName = ActivationFunctionName.GELU_NEW
+
+    gradient_checkpointing: bool = True  # better to just always use this
+    gradient_checkpointing_block_size: int = 5
+
+    use_bias: bool = True
+
+    # Hyena-specific parameters
+    hyena_order: int = 2  # depth of the Hyena recurrence
+    hyena_filter_order: int = 64  # width of the FFN parametrizing the implicit filter
+    hyena_inner_factor: int = 1  # inner dimension multiplier
+    hyena_short_filter_order: int = 3  # length of the explicit input convolutional filter
+    hyena_outer_mixing: bool = False  # whether to use outer mixing
+
+    # Hyena filter parameters
+    hyena_emb_dim: int = 3  # dim of input to MLP, augments with positional encoding
+
+    # Hyena modulation parameters
+    hyena_fast_decay_pct: float = 0.3
+    hyena_slow_decay_pct: float = 1.5
+    hyena_target: float = 1e-2
+    hyena_modulate: bool = True
+    hyena_shift: float = 0.0
+
+    # Additional Hyena options
+    hyena_post_order_ffn: bool = False  # Apply a dense layer between steps of the recurrence
+    hyena_return_state: bool = False  # Whether to return state information
+
+    # Axes
+    Pos = property(lambda self: Axis(name="position", size=self.seq_len))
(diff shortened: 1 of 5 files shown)


src/levanter/utils/activation.py
garymm, line 1: If you like this I can open a separate PR to use it throughout the repo and then rebase this PR on that.

src/levanter/models/gpt2_hyena.py
garymm, line 1: I could potentially refactor Gpt2 to allow for less code duplication, but then it would be more abstract and less self-contained. LMK what you prefer.

src/levanter/models/hyena.py
```diff
@@ -0,0 +1,540 @@
+"""An implementation of the Hyena operator.
+
+Paper: [Hyena Hierarchy: Towards Larger Convolutional Language Models](https://arxiv.org/abs/2302.10866)
+Official implementation in PyTorch:
+https://github.com/HazyResearch/safari/blob/541902aca88cb11af4d816ac762f3303e4ff8eea/src/models/sequence/hyena.py
+
+Current diffences from the official impl:
+- We don't support inner_factor.
+- We don't support post_order_ffn.
+- We don't support num_heads (the PyTorch impl's support for multiple heads seems incomplete, or at least
```
garymm, line 10: I could be wrong about this. LMK if you can tell if their code actually supports multiple heads, and if so please explain it to me and I will try to port it over.

garymm (commented):
I have run the nano configuration which basically just ensures that it doesn't crash. I haven't verified if it actually learns. I will work on that next, but thought it might be valuable to put this up for review now.

garymm:
Currently running the following on 2xH100, will let it run over night:

```yaml
data:
  id: dlwh/wikitext_103_detokenized

model:
  type: gpt2_hyena
  hidden_dim: 512
  num_layers: 8
  seq_len: 1024


trainer:
  tracker:
    tags: [ "wikitext", "gpt2_hyena"]

  mp: p=f32,c=bfloat16
  model_axis_size: 1
  checkpointer:
    keep:
      - every: 50
    save_interval: 5m

  per_device_parallelism: -1
  train_batch_size: 64

optimizer:
  learning_rate: 6E-4
  weight_decay: 0.1
  min_lr_ratio: 0.1
```

dlwh (commented):
will look more later. a bit overwhelmd

src/levanter/models/gpt2_hyena.py
dlwh: i think the copy paste is better probably.

src/levanter/models/hyena.py
dlwh: doesn't it support num_heads:

https://github.com/HazyResearch/safari/blob/541902aca88cb11af4d816ac762f3303e4ff8eea/src/models/sequence/hyena.py#L317-L321

dlwh: oh right you said you don't really speak einsum. 

In their code:

```
rearrange(uc, 'b (ho v) (z l) -> b ho v z l', 
            z=self.num_blocks, 
            ho=self.num_heads, 
            v=self.head_dim * (self.order + 1)
```

amounts to:

```
b, d, seqlen = uc.shape
ho = self.num_blocks
v = self.head_dim * (self.order + 1)
z = self.num_blocks
l = seqlen // z
uc.reshape((b, ho, v, z, l)
```

so it reshapes the embedding dim into (heads, head size)

src/levanter/models/hyena.py
```diff
@@ -0,0 +1,542 @@
...
+  I don't understand it).
+- We more efficiently support multiple blocks. I believe the PyTorch version is inefficient: It uses
+  the full sequence length in the PositionalEmbedding.
```
dlwh, line 13: but shouldn't it?

src/levanter/models/hyena.py
```diff
@@ -0,0 +1,542 @@
...
+            Embed: Axis of positional embedding
+            key: Optional random key (not used)
+        """
+        seq_len = PosPerBlock.size
+
+        # Ensure emb_dim is valid
+        if Embed.size <= 1:
+            raise ValueError("emb_dim must be greater than 1")
+
+        # The time embedding fed to the filteres is normalized so that t_f = 1
```
dlwh, line 105:
```suggestion
        # The time embedding fed to the filters is normalized so that t_f = 1
```

src/levanter/models/hyena.py
```diff
@@ -0,0 +1,542 @@
...
+        # Calculate number of frequency bands
+        bands = (Embed.size - 1) // 2
+        Band = Axis("band", bands)
+
+        # To compute the right embeddings we use the "proper" linspace
+        t_rescaled = hax.linspace(PosPerBlock, start=0, stop=seq_len - 1)
+        w = 2 * jnp.pi * t_rescaled / seq_len
+
+        f = hax.linspace(Band, start=1e-4, stop=bands - 1)
+        z_complex = hax.exp(-1j * f.broadcast_axis(PosPerBlock) * w.broadcast_axis(Band))
```
dlwh, line 117: i'm not checking this math so I'm assuming this is fine

garymm (commented):
Thanks for the helpful review!

src/levanter/models/hyena.py
```diff
@@ -0,0 +1,542 @@
...
+
+        return x
+
+
+class MLPTrainableActivation(eqx.Module):
+    """
+    hax.nn.MLP with a trainable activation function.
+    """
+
+    activation: Callable = eqx.field(static=False)  # this is the main difference from hax.nn.MLP
```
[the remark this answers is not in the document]
dlwh: yep you're definitely right!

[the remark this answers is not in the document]
dlwh: maybe add to the docstirng that it's a shared activation?

src/levanter/models/hyena.py
```diff
@@ -0,0 +1,542 @@
...
+        l_filter = min(input_length, Pos.size)
+
+        # Input projection from [Embed] to [(order+1) * Embed]
+        u = self.in_proj(u, key=key_in_proj)
+
+        # trying to keep the variable names from the official impl.
+        # I think uc stands for "u convolved".
+        uc = self.short_filter(u).rename({"out_channels": EmbedOrderPlus1})
+        # Note: after the short filter, uc."position" axis has slightly larger size due to padding.
+        # Hence we use Pos.name rather than Pos to slice, otherwise haliax complains the axis sizes
```
dlwh, line 512: what do you think about this behavior? I think I don't like it anymore and am thinking of changing it

src/levanter/utils/activation.py
dlwh: yeah i like! We can replace the ACT2FN map. (I do want lower case to still work I guess, but it's not a big deal)

dlwh (approved):
lgtm!

dlwh:
i have a fix for test failures if you merge main

PR #922, opened 2025-03-26
use ActivationFunctionEnum in more places

garymm:
More strongly typed than a str, so will help detect errors earier.
Also centralizes things in utils rather than GPT2 code.


garymm:
Waiting on https://github.com/stanford-crfm/levanter/pull/913 to be merged, then this will be rebased and only the last commit should have diffs vs main.

dlwh (approved):

dlwh:
thanks!

PR #932, opened 2025-04-01
add feistel networks as an option, make them default in API

dlwh:
(but not in the config, for now)


src/levanter/data/_prp.py
```diff
@@ -5,17 +5,80 @@
...
+        Args:
+            indices: An integer or an array of integers to be permuted.
+
+        Returns:
+            The permuted value(s).
+        """
+        pass
+
+    @staticmethod
+    def make(kind: PermType, length: int, prng_key, **kwargs) -> "Permutation":
```
ahmeda14960, line 54: maybe clarify the type for prng_key?

src/levanter/data/_prp.py
```diff
@@ -5,17 +5,80 @@
...
+            **kwargs: Additional arguments for the specific permutation type.
+
+        Returns:
+            An instance of the specified permutation type.
+        """
+        if kind == "feistel":
+            return FeistelPermutation(length, prng_key, **kwargs)
+        elif kind == "linear":
+            return LcgPermutation(length, prng_key)
+        else:
```
ahmeda14960, line 70: cursor says the else is unreachable because you used a literal type for PermType, I'm guessing any other inputs would throw an error at run time anyway? not too important

src/levanter/data/_prp.py
```diff
@@ -67,3 +133,120 @@ def __call__(self, indices):
...
+
+    def __init__(self, length: int, prng_key, rounds: int = 5):
+        self.length = length
+        self.m = next_power_of_two(length)  # m is a power-of-two >= length
+        self.rounds = rounds
+        self.bits = self.m.bit_length() - 1  # because m == 1 << bits
+        # Split the bits into two halves.
+        self.left_bits = self.bits // 2
+        self.right_bits = self.bits - self.left_bits
+        self.R_mask = (1 << self.right_bits) - 1
```
ahmeda14960, line 154: I think we also need a check to make sure that rounds > 0 and that the length is greater than 1 here

ahmeda14960: we can also just change the pemtuation so that if the length is <= 1 we always do LcgPermutation

src/levanter/data/_prp.py
```diff
@@ -67,3 +133,120 @@ def __call__(self, indices):
...
+        # Convert JAX PRNG to numpy RNG and generate round keys in range [0, 2^(right_bits))
+        self.rng = np.random.Generator(np.random.PCG64(jrandom.randint(prng_key, (), 0, 2**30).item()))
+        self.keys = self.rng.integers(0, 1 << self.right_bits, size=rounds, dtype=np.uint64)
+
+    def _F(self, right: np.ndarray, key: np.uint64) -> np.ndarray:
+        """A simple round function that mixes the right half.
+
+        Operates modulo 2^(right_bits).
+        """
+        return (right * np.uint64(2654435761) + key) & self.R_mask
```
ahmeda14960, line 165: Claude says the below is safer for preventing overflow and should still give a valid permutation (not sure how much overflow is a concern here but oh well )

```suggestion
       # Apply mask before multiplication to prevent potential overflow
        masked_right = right & self.R_mask
        return (masked_right * np.uint64(2654435761) + key) & self.R_mask
```

src/levanter/data/_prp.py
```diff
@@ -67,3 +133,120 @@ def __call__(self, indices):
...
+        if np.any(indices < 0) or np.any(indices >= self.length):
+            raise IndexError(f"Index {indices} is out of bounds for length {self.length}")
+
+        # Embed indices into the larger domain.
+        x = indices
+        # Apply cycle-walking: reapply the Feistel network until the result is in [0, length).
+        out = self._feistel(x)
+        mask = out >= self.length
+        cycles = 0
+        while np.any(mask):
```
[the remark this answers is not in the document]
ahmeda14960: WDYT? Weirdly copilot didn't suggest an actual number but if it is possible for this to cycle forever maybe good to add a runtime warning. Something like 1000 cycles?

ahmeda14960 (approved):

src/levanter/data/_prp.py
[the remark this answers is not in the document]
dlwh: yeah i actually already did i guess i didn't push. ChatGPT tells me the chance of this happening is theoretically < 2^-300 but could be bugs so

src/levanter/data/_prp.py
dlwh: Python doesn't care about types :-)

src/levanter/data/_prp.py
dlwh: sure

src/levanter/data/_prp.py
ahmeda14960: doesn't using the literal allow python to check the value through static type checking?

dlwh: sure but if you pass it a string through the kwargs it will still hit it. It's better to be defensive

PR #936, opened 2025-04-08
change eval_lm to log much more information...

dlwh:
don't support loading two kinds of models in one go

### DIFF
--- a/src/levanter/analysis/tree_stats.py
+++ b/src/levanter/analysis/tree_stats.py
@@ -1,0 +1,102 @@
+from typing import Any
+
+import jax
+from jax import numpy as jnp
+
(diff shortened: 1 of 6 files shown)


src/levanter/analysis/tree_stats.py
```diff
@@ -0,0 +1,102 @@
...
+
+import haliax.nn
+from haliax import NamedArray, is_named_array
+from haliax.jax_utils import is_jax_array_like
+
+from levanter.tracker.histogram import Histogram
+from levanter.utils import jax_utils
+
+
+def summary_statistics_for_tree(
```
dlwh, line 14: this is just a move from watch.py

PR #939, opened 2025-04-11
Fix some footguns in JaggedArray

dlwh:
renames the data/offsets/shapes field of JaggedArray to _data etc, and makes the properties `data` etc trim to the number of actually used entries.

This is because _data etc are set to have very long logical lengths (which makes cache construction easier) but this leads to a footgun if you want to say, read all offsets into ram.

### DIFF
--- a/src/levanter/store/jagged_array.py
+++ b/src/levanter/store/jagged_array.py
@@ -91,9 +91,11 @@ class JaggedArrayStore:
     High latency, but high throughput.
     """
 
-    offsets: ts.TensorStore  # offsets of the start of each array, except that index[0] is the number of arrays
-    data: ts.TensorStore
-    shapes: Optional[ts.TensorStore]  # (len(offsets), len(data.shape)-1)
+    # Note that _offsets, _data and _shapes have very very large length and you shouldn't read them directly.
+    # Use the properties offsets, data, and shapes to get the actual data and avoid footguns.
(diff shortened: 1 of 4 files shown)


tests/test_jagged_array.py
```diff
@@ -9,173 +9,180 @@
 from levanter.store.jagged_array import JaggedArrayStore, PreparedBatch
 
 
-class TestJaggedArrayStore:
```
dlwh, line 12: most of the change in this file is me removing this unnecessary class

dlwh: (the rest is using _data in a few places)

src/levanter/store/jagged_array.py
```diff
@@ -431,14 +453,14 @@ def __getitem__(self, item):
             start, stop, step = item.indices(len(self))
             # for now, just read the data into a list
 
-            return [self[i] for i in range(start, stop, step)]
```
ahmeda14960, line 434: out of curiosity, did you just change this to because get_batch_sync is faster? It should be same logic the same right

ahmeda14960 (approved):

src/levanter/store/jagged_array.py
dlwh: yes get_batch is dramatically faster. Tensorstore has a high latency overhead to gcs but if you batch it it's pretty fast

PR #944, opened 2025-04-16
Remove old unneeded functions

dlwh:


src/levanter/data/text.py
```diff
@@ -1,6 +1,5 @@
 import abc
 import asyncio
-import copy
```
ahmeda14960, line 3: I think we still need this for _mask_overlap function at least

ahmeda14960 (approved):

PR #958, opened 2025-05-01
Add MoE support (Mixtral)

blahBlahhhJ:


src/levanter/models/mixtral.py
```diff
@@ -0,0 +1,625 @@
...
+        # self attention and skip connection
+        residual = x
+        x = self.input_layernorm(x)
+        attn_output = self.self_attn(x=x, mask=mask, key=k_attn)
+        x = residual + attn_output
+
+        # MLP and skip connection
+        residual = x
+        x = self.post_attention_layernorm(x)
+        mlp_output = self.shared_mlp(x, key=k_mlp) if self.shared_mlp is not None else 0
```
dlwh, line 475: nit: don't love the one line here. would rather just use statement form.

dlwh (approved):

PR #959, opened 2025-05-07
Hybrid norm

WhenWen:
Implemented hybrid norm and input embedding norm on top of the original llama architecture. 

These two norms can be enabled by setting the value of the following two flags.

--   hybrid_norm: bool = False
--   input_embedding_norm: bool = False


src/levanter/models/llama.py
```diff
@@ -45,6 +45,8 @@ class LlamaConfig(HFCompatConfig):
             Note that num_heads must be divisible by this number. Defaults to 32.
         activation_function (str, optional): activation function for the hidden layer. Defaults to "silu".
         rope_scaling (Dict, optional): dict containing the scaling configuration for the Rotary Positional Embedding.
+        hybrid_norm (bool, optional): whether to use hybrid normalization with additional layer norms after attention and MLP. Defaults to False.
+        input_embedding_norm (bool, optional): whether to use layer normalization after input embeddings. Defaults to False.
```
dlwh, line 49: i think we need to fail on HF export or issue a big warning since these config options don't export. wdyt?

WhenWen: I agree. Let me add this feature today.

WhenWen: The error msg is added

dlwh (approved):

PR #961, opened 2025-05-12
annoyed the type checker didn't catch this

dlwh:


kothasuhas (commented):
totally my bad! can we just remove this? i think this was an accident from some previous fiddling but is unnecessary now, doesn't need to be commented out

kothasuhas (approved):

PR #972, opened 2025-05-23
Add basic model sampling/inference support

neel04:
## Intro
While inferencing is not a core focus of `levanter` (and better left to various other inference backends) having the ability to do decently fast sampling is highly useful for debugging and more importantly, evals.

This (draft) PR aims to setup the basic structure to accomodate more contributions in the future and further expand levanter's capabilities.

## Current Progress

Currently, I have implemented 2 samplers here - a "naïve" greedy sampler which chooses the highest probability token, and a slightly more feature-rich greedy sampler which includes: Top-k, Nucleus sampling, repetition penalty and temperature.

Performance right now is not a core focus. The sampler is JIT-able, and a `lax.scan` wrapped caller is also provided in the form of `scan_generate` however their performance hasn't been tested as I'm on M2 Mac and `jax-metal` doesn't play well with anything right now.

I have also pushed a minimal example in `scripts/simple_inference.py` which loads a model from HF, converts it to being levanter friendly and does prompt completion. Hopefully, it can serve as documentation on how to use the sampler.

Criticisms and opinions, especially around the abstraction are more than welcome :) This is a first-pass attempt and rewrites will have to be done.

---

## To-do (future PRs)
- [ ] `KV-cache`-ing
- [ ] Streaming
- [ ] Multi-turn interactive conversations
- [ ] Hook into Eleuther's eval harness for generation tasks


src/levanter/generator.py
```diff
@@ -0,0 +1,296 @@
...
+    """Struct to encapsulate arguments for the Greedy Sampler."""
+
+    repetition_token_window: int = 4
+    repetition_penalty: float = 0.8
+    top_k: int = 50
+    top_p: float = 0.8
+
+
+class Generator:
+    hf_config: Config
```
dlwh, line 78: our main use case is going to be evals during training, so it would be better if it took an LmHeadModel instead (or a subclass that supported efficient inference or something)

src/levanter/generator.py
```diff
@@ -0,0 +1,296 @@
...
+        model: LmHeadModel[Config],
+        key: PRNGKeyArray = PRNGKey(0),
+    ) -> None:
+        self.hf_config = hf_config.config  # type: ignore
+        self.tokenizer = tokenizer
+        self.model = model
+        self.key = key
+
+    def _tokenize(self: Self, input_str: str) -> Tuple[NamedArray, ...]:
+        self.tokenizer.pad_token = self.tokenizer.eos_token
```
dlwh, line 99: don't love the silent mutation here. big warning and only do it if we need to

src/levanter/generator.py
```diff
@@ -0,0 +1,296 @@
...
+            p=nucleus["batch", 1],
+            axis=axis,
+        )
+
+        return token
+
+    def scan_generate(
+        self: Self,
+        args: SamplingArgs,
+        sampler: Literal["greedy_samping"] = "greedy_samping",
```
dlwh, line 174:
```suggestion
        sampler: Literal["greedy_sampling"] = "greedy_sampling",
```

src/levanter/generator.py
```diff
@@ -0,0 +1,296 @@
...
+            token = gen_sampler(args, logits, last_tok_idx)
+
+            new_prompt_ids = prompt_ids.at["position", last_tok_idx + 1].set(token)
+            new_attn_mask = attn_mask.at["position", last_tok_idx + 1].set(1)
+
+            return (new_prompt_ids, new_attn_mask, last_tok_idx + 1), token
+
+        start = time.perf_counter()  # to propogate timings
+
+        (gen_prompt_ids, _, _), _ = jax.lax.scan(f=_scan_body, init=initial_carry, xs=None, length=args.num_tokens)
```
dlwh, line 201: whole thing should be in jit or you're not going to get any performance wins

src/levanter/generator.py
```diff
@@ -0,0 +1,296 @@
...
+            output_tokens=prompt_ids.array[0].tolist(),
+            num_tokens=args.num_tokens,
+            decoded_output=model_output,
+            total_time_taken=time.perf_counter() - start,
+        )
+
+    def generate(
+        self: Self,
+        args: SamplingArgs,
+        sampler: Literal["greedy_samping"] = "greedy_samping",
```
dlwh, line 219: etc

src/levanter/generator.py
```diff
@@ -0,0 +1,296 @@
...
+        prompt_ids, attn_mask = self._tokenize(args.prompt._to_string())
+
+        gen_sampler = self._get_sampler(sampler, args)
+
+        model = eqx.filter_jit(self.model)
+        last_tok_idx = hax.sum(attn_mask) - 1  # attn_mask is guranteed binary
+
+        start = time.perf_counter()
+
+        for _ in tqdm(range(args.num_tokens)):
```
dlwh, line 230: i prefer tqdm_loggable

dlwh (commented):
generally seems like a good direction! I won't feel comfortable merging until we see it working and the perf isn't completely embarrassing

PR #973, opened 2025-05-24
[jules] refactor: Preserve inner optimizer state on skip

dlwh:
refactor: Preserve inner optimizer state on skip

Adds a SkipStepConfig to allow optimizers to skip steps if the loss and/or gradient norms are too high, based on a rolling window of statistics.

Key changes:
- Introduced `SkipStepConfig` dataclass in `levanter.optim.config` with `rolling_interval_length` and `sigma_factor` parameters.
- Added `skip_step: Optional[SkipStepConfig]` field to `OptimizerConfig`.
- Implemented `SkipStepConfig.wrap(inner_optimizer)` which returns a new `optax.GradientTransformation`. This wrapper:
    - Maintains a rolling history of losses and gradient norms in its state.
    - Expects `loss` as a keyword argument in the `update()` function's `**extra_args`.
    - Calculates a `step_factor` (0.0 or 1.0). If the current loss or gradient norm exceeds `sigma_factor` standard deviations from the rolling mean (after an initial warmup period), `step_factor` becomes 0.0.
    - Scales gradients by `step_factor` before passing them to the `inner_optimizer`, effectively skipping the update if `step_factor` is 0.0.
- Integrated this wrapping mechanism into `AdamConfig.build()`.
- Added comprehensive unit tests in `tests/test_optimizer_config.py` to verify the behavior of `SkipStepConfig`, including various skipping scenarios, state management, and integration with optimizer building. Previously, zero gradients were passed to the inner optimizer, which would still modify its state. With this change:
- If a step is skipped, the `update_fn` in `SkipStepConfig.wrap` now returns the original inner optimizer state from before the update.
- Zero gradients are still returned as the update for the skipped step.
- I updated the unit tests to verify that the inner optimizer state is indeed preserved during a skip.

This change makes the skipping behavior more aligned with a 'true skip', preventing any modification to the optimizer's internal accumulators when the conditions for skipping are met.


blahBlahhhJ:
Nice! Now MFU for MoE should increase by ~3% I think.

PR #974, opened 2025-05-24
Initial implementation of power scheduler

cats-marin:
Implementation based on this paper. https://arxiv.org/abs/2408.13359. 

@dlwh @Helw150 I'm not sure if this is the best way to go about implementing this as is. I had to add five more parameters to the parameter list of the OptimizerConfig class to implement Power Scheduler, as, based on my understanding, there is a tight coupling between the optimizer and the scheduler. Should there be an effort to decouple the optimizer and the scheduler, so that we don't have to pass schedule-specific parameters directly into OptimizerConfig?

### DIFF
--- a/src/levanter/optim/config.py
+++ b/src/levanter/optim/config.py
@@ -18,7 +18,7 @@ import levanter.tracker
 from levanter.utils.jax_utils import leaf_key_paths
 
 
-@dataclass
+@dataclass(frozen=True)
 class LrScheduleContext:
     warmup_steps: int
     decay_steps: int
--- a/src/levanter/optim/config.py
+++ b/src/levanter/optim/config.py
@@ -71,20 +71,21 @@ class InvLrSchedule(LrSchedule):
 
 
 @LrSchedule.register_subclass("power")
+@dataclass
 class PowerLrSchedule(LrSchedule):
     # Power Scheduler: A Batch Size and Token Number Agnostic Learning Rate Scheduler (Shen et al., 2024)
     # https://arxiv.org/abs/2408.13359
     # The scheduler and default hyperparameters are intended for use with maximal update parametrization
     # (mup), as described in the paper. The scheduler may work without mup.
 
-    a: float = 4.6
-    """Learning rate amplitude for the power learning rate schedule. Must be a positive number."""
-    b: float = -0.51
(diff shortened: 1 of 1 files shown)


dlwh:
yeah it's probably time to make a more configurable schedule thing...

dlwh:
ok, you asked for it! so scheduler refactor is going to be a bit delicate just since we need to support old-style configs because I don't want to break too much compatibility (and we probably want to keep the simple case simple).

I'm going to describe it in its full glory, but I think we can split it into chunks, and you only really need to do the first one.

In an ideal world (based on our experience with Marin 8B), I think the schedule system would look something like

```python
lr_schedule: str | LrSchedule | list[ScheduleStep[LrSchedule]]
```

If lr_schedule is a string, use old behavior. Otherwise, use new behavior.

## Chunk 1: LrSchedule

This will enable

```python
lr_schedule: str | LrSchedule
```

LrSchedule should probably look a bit like how we do the OptimConfig, which is to say a draccus plugin registry.

We should make a LrSchedule type that is a PluginRegistry and add the kinds we already have.

LrSchedule should probably have all settings needed to specify it, except for the schedule length, which should be figured out from context.

normal Cosine would be specified as either:

```yaml
optimizer:
     lr_schedule: "cosine"
     warmup: ...
     learining_rate: ... 
```

or

```yaml
optimizer:
    lr_schedule:
          type: cosine
          learning_rate: <peak lr>
          exponent: ...
```

for more fine-grained control...

## Chunk 2: Staged Schedules

We should support staged schedules too:

```yaml
optimizer:
   lr_schedule:
       # warmup
        - start: 0
           value:
               type: linear 
               initial_lr: 0.0
               final_lr: 1e-4
       # stable
        - start: 1000 
          value:
              type: constant
              learning_rate: 1e-4
        # decay
        - start: 100000
           value:
               type: inv
```

This is obviously a pain to specify for the common case but sometimes we need the control


This is a big lift, so lmk if this is too daunting! Like I said, we only really need the first chunk right now

cats-marin:
I implemented chunk 1. I passed in ctx: LrScheduleContext for LrSchedule to avoid the user being able to re-declare learning_rate and min_lr_ratio, which may lead to hidden bugs. 

I will probably revisit and implement chunk 2 some other times.

dlwh (approved):
this is great, thanks!

PR #980, opened 2025-05-30
Safetensor gcs

dlwh:
@erfanzar

### DIFF
--- a/src/levanter/compat/fsspec_safetensor.py
+++ b/src/levanter/compat/fsspec_safetensor.py
@@ -146,56 +146,11 @@ def _make_async_read_fn(reader: _AsyncFsspecReader, meta):
     return read
 
 
-def _make_async_write_fn(writer: _AsyncFsspecReader, meta):
-    dtype = SAFETENSORS_DTYPE_MAP[meta["dtype"]]
-    shape = meta["shape"]
-    base_offset = meta["offset"][0]
-    itemsize = dtype.numpy_dtype.itemsize
-
-    async def write(domain: ts.IndexDomain, array: np.ndarray, _params):
-        slices = domain.index_exp
-        requested_shape = tuple(s.stop - s.start for s in slices)
-
-        # Compute the flat byte offset for each element
-        byte_writes = []
-
-        for idx in np.ndindex(requested_shape):
-            global_idx = tuple(s.start + i for s, i in zip(slices, idx))
-            flat_index = np.ravel_multi_index(global_idx, shape)
-            byte_offset = base_offset + flat_index * itemsize
-            byte_data = array[idx].tobytes()
-            byte_writes.append((byte_offset, byte_data))
-
-        # Coalesce adjacent writes
-        byte_writes.sort()
-        coalesced = []
-        cur_start, cur_buf = byte_writes[0][0], byte_writes[0][1]
-        for off, data in byte_writes[1:]:
-            if off == cur_start + len(cur_buf):
-                cur_buf += data
-            else:
-                coalesced.append((cur_start, cur_buf))
-                cur_start, cur_buf = off, data
-        coalesced.append((cur_start, cur_buf))
-
-        await asyncio.gather(*[writer.fs._pipe_file(writer.gcs_path, data, start=start) for start, data in coalesced])
-
-    return write
-
-
 async def load_tensor_dict(gcs_path: str, cache_size=128) -> dict[str, ts.TensorStore]:
     """
 
     Loads a safetensors file from GCS (or other fsspec file system) into a dictionary of TensorStore objects.
     This allows for efficient sharded reads.
-
-    Writes are technically supported, but are not very efficient.
-
-    Args:
-        gcs_path:
-        cache_size:
-
-    Returns:
 
     """
     reader = _AsyncFsspecReader(gcs_path, cache_size=cache_size)
@@ -204,10 +159,8 @@ async def load_tensor_dict(gcs_path: str, cache_size=128) -> dict[str, ts.Tensor
     tensor_map = {}
     for key, meta in metadata.items():
         read_fn = _make_async_read_fn(reader, meta)
-        write_fn = _make_async_write_fn(reader, meta)
         ts_arr = ts.virtual_chunked(
             read_function=read_fn,
-            write_function=write_fn,
             dtype=SAFETENSORS_DTYPE_MAP[meta["dtype"]],
             shape=meta["shape"],
             loop=asyncio.get_event_loop(),
--- a/src/levanter/compat/fsspec_safetensor.py
+++ b/src/levanter/compat/fsspec_safetensor.py
@@ -8,7 +8,7 @@ import numpy as np
 import tensorstore as ts
 
 
-class AsyncLRUCache:
+class _AsyncLRUCache:
     def __init__(self, maxsize=128):
         self.maxsize = maxsize
         self.cache = OrderedDict()
@@ -32,14 +32,14 @@ class AsyncLRUCache:
             return result
 
 
-class AsyncGCSReader:
+class _AsyncFsspecReader:
     def __init__(self, gcs_path: str, cache_size=128):
         self.gcs_path = gcs_path
         protocol = fsspec.core.split_protocol(gcs_path)[0]
         if protocol is None:
             protocol = "file"
         self.fs = fsspec.filesystem(protocol, asynchronous=True, anon=False)
-        self.cache = AsyncLRUCache(cache_size)
+        self.cache = _AsyncLRUCache(cache_size)
 
     async def read_range(self, start: int, length: int) -> bytes:
         key = (start, length)
@@ -55,7 +55,7 @@ class AsyncGCSReader:
         return await self.cache.get_or_wait(key, fetch)
 
 
-async def load_safetensors_metadata(reader: AsyncGCSReader):
+async def _load_safetensors_metadata(reader: _AsyncFsspecReader):
     """Parses metadata directly from a safetensors file using fsspec."""
     header_len_bytes = await reader.read_range(0, 8)
     (header_len,) = struct.unpack("<Q", header_len_bytes)
@@ -78,7 +78,7 @@ async def load_safetensors_metadata(reader: AsyncGCSReader):
     return tensors
 
 
-def coalesce_reads(offsets, itemsize, max_gap=128):
+def _coalesce_reads(offsets, itemsize, max_gap=128):
     """Group byte offsets into contiguous or nearly-contiguous ranges."""
     offsets = sorted(set(offsets))
     ranges = []
@@ -96,7 +96,7 @@ def coalesce_reads(offsets, itemsize, max_gap=128):
     return ranges
 
 
-def make_async_read_fn(reader: AsyncGCSReader, meta, tensor_key):
+def _make_async_read_fn(reader: _AsyncFsspecReader, meta):
     dtype = SAFETENSORS_DTYPE_MAP[meta["dtype"]]
     shape = meta["shape"]
     base_offset = meta["offset"][0]
@@ -119,7 +119,7 @@ def make_async_read_fn(reader: AsyncGCSReader, meta, tensor_key):
             flat_indices.append((file_offset, dest_flat_index))
 
         byte_offsets = [off for off, _ in flat_indices]
-        ranges = coalesce_reads(byte_offsets, itemsize)
+        ranges = _coalesce_reads(byte_offsets, itemsize)
 
         # Fetch coalesced chunks
         chunks = await asyncio.gather(*[reader.read_range(start, end - start) for start, end in ranges])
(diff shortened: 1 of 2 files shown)


src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+        path: str,
+        *,
+        fs: Optional[AbstractFileSystem] = None,
+        cache_size: int = 32,
+    ):
+        protocol, fs_path = fsspec.core.split_protocol(path)
+        if fs is None:
+            if protocol is None:
+                protocol = "file"
+            if protocol == "gs":
```
rjpower: Why is this part necessary?

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+
+        lock = self._locks.setdefault(key, asyncio.Lock())
+        async with lock:
+            cached = self._cache.get(key)
+            if cached is not None:
+                return cached
+
+            end = start + length
+            if self._async_mode and hasattr(self._fs, "_cat_file"):
+                data = self._fs._cat_file(self._path, start=start, end=end)
```
rjpower: Ditto, what's the advantage of _cat_file vs cat_file?

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+            async_mode = True
+        else:
+            filesystem = fs
+            resolved_path = fs_path if protocol is not None else path
+            try:
+                current_loop = asyncio.get_running_loop()
+            except RuntimeError:
+                current_loop = None
+            fs_loop = getattr(filesystem, "loop", None)
+            async_flag = bool(getattr(filesystem, "asynchronous", False))
```
rjpower: Is the idea there are some filesystems that aren't async compatible and you want to shim them? Consider maybe creating a quick wrapper above for this specific behavior?

class ReadOnlySyncFSWrapper:
   """Transforms sync fsspec filesystems into an asynchronous compatible interface using threads."""


but this is simple enough i suppose it's fine to do it all here.

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+            async_mode = async_flag and (fs_loop is None or fs_loop is current_loop)
+
+        self._fs: AbstractFileSystem = filesystem
+        self._path = resolved_path
+        self._cache_size = cache_size
+        self._cache: Dict[Tuple[int, int], bytes] = {}
```
rjpower: Shouldn't cache hits be rare (assuming block size is reasonable?) IIUC, even if your blocksize was 1MB, a 32B model would only be 32k reads...?

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+                data = await loop.run_in_executor(None, _read_sync)
+
+            if len(self._cache) >= self._cache_size:
+                self._cache.pop(next(iter(self._cache)))
+            self._cache[key] = data
+            self._locks.pop(key, None)
+            return data
+
+
+async def _read_metadata(reader: _AsyncFsspecReader) -> Dict[str, _TensorRecord]:
```
rjpower: Probably worth a link to https://huggingface.co/docs/safetensors/en/index#format

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+
+_PBAR_THRESHOLD_BYTES = 256 * 1024**2
+_PBAR_READ_STEP_BYTES = 64 * 1024**2
+
+
+@lru_cache(maxsize=1)
+def _default_chunk_size_bytes() -> int:
+    """Lazily resolve the default chunk size from the environment."""
+
+    raw = os.environ.get("LEVANTER_GCS_CHUNK_SIZE")
```
rjpower: Why so big by default?

src/levanter/compat/hf_checkpoints.py
```diff
@@ -559,11 +575,43 @@ def _load_shards(self, id: str, index_file: str, rev: Optional[str], dtype) -> d
...
+
+        if not shard_files:
+            raise FileNotFoundError(f"No HF-ish checkpoint files found in {url}")
 
+        for shard_file in shard_files:
+            shard_path = os.path.join(path, shard_file)
+            if not fs.exists(shard_path):
+                raise FileNotFoundError(f"Shard file {shard_path} not found")
+
+            if loader is _load_safe_tensors:
```
rjpower, line 607: This probably deserves a comment ("torch loader doesn't support remote filesystems")

src/levanter/compat/hf_checkpoints.py
```diff
@@ -633,19 +671,11 @@ def load_pretrained(
...
-            contexts.enter_context(local_cpu_mesh())
+        # TODO: in an ideal world, we would only load the part of the array we needed, but
+        # AFAICT neither torch state dicts nor safetensors support this.
+        state_dict = self.load_state_dict(ref, dtype)
 
-        with contexts:
-            # TODO: in an ideal world, we would only load the part of the array we needed, but
-            # AFAICT neither torch state dicts nor safetensors support this.
-            state_dict = self.load_state_dict(ref, dtype)
+        print(list(state_dict.keys()))
```
rjpower: nit remove?

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+
+        def _read_sync(target: bytearray, with_progress: bool) -> None:
+            desc = f"Reading safetensor chunk {chunk.chunk_id}"
+            reader = self._reader
+            filesystem = reader.filesystem
+            with filesystem.open(reader.path, "rb") as f:  # type: ignore[arg-type]
+                f.seek(chunk.byte_start)
+                remaining = chunk.size
+                offset = 0
+                if with_progress:
```
rjpower: shouldn't progress be over the full set of files? this feels like it would be noisy if we're reading in parallel?

if you wanted progress seems better to define some kind of LoaderCtx and let the loader call chunk_done() or whatever.

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+        if buffer is None:
+            raise RuntimeError(f"Chunk {chunk.chunk_id} has been released and cannot be rematerialised")
+        logger.debug(
+            "Process %d observed completed chunk %d",
+            self._process_index,
+            chunk.chunk_id,
+        )
+        return buffer
+
+    async def _read_chunk_bytes(self, chunk: _ChunkSpec) -> bytes | bytearray:
```
rjpower: Yeah it feels like making the Async wrapper would be preferable. As it is I'd worry that we've got a bug in the sync version or async version that we won't observe in testing.

It seems weird that you've defined an async reader but then you're checking it's async here.

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,551 @@
...
+            )
+            return existing
+
+        event = self._chunk_events.get(chunk.chunk_id)
+        if event is None:
+            event = asyncio.Event()
+            self._chunk_events[chunk.chunk_id] = event
+
+        if not event.is_set():
+            lock = self._chunk_locks.setdefault(chunk.chunk_id, asyncio.Lock())
```
rjpower: Probably a race on the threads that try to put the lock into the dict. You could create all the locks up front to avoid this.

OTOH, it looks like the usage is just to take the list of chunks/tensors to load, why not just make that the public interface and then you can dedup there? Then you could get rid the release_buffer stuff and the locking, and just have:

load_tensors(list[TensorId]) -> dict[TensorId, np.array]

src/levanter/compat/fsspec_safetensor.py
dlwh: i dunno, gotta read it all. I can change it to like 128M I guess

src/levanter/compat/fsspec_safetensor.py
dlwh: isn't. will fix

src/levanter/compat/fsspec_safetensor.py
dlwh: yeah, i probably let codex have too much free reign here. I can have it refactor it later. HFFileSystem doesn't implement the async interface

src/levanter/compat/hf_checkpoints.py
dlwh: oops

src/levanter/compat/fsspec_safetensor.py
dlwh: fair point

src/levanter/compat/fsspec_safetensor.py
dlwh: I'm gonna just make it all sync. this is unnecessary

src/levanter/compat/fsspec_safetensor.py
dlwh: yeah gonna siemplify it

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,290 @@
...
+    "U16": np.dtype("uint16"),
+    "U32": np.dtype("uint32"),
+    "U64": np.dtype("uint64"),
+    "BOOL": np.dtype("bool"),
+}
+
+
+ShardingFunction = Callable[[Tuple[int, ...]], Optional[jax.sharding.Sharding]]
+
+DEFAULT_CHUNK_SIZE_BYTES = int(os.environ.get("LEVANTER_FSSPEC_CHUNK_BYTES", 2 * 1024**3))
```
rjpower, line 42: I guess I'm confused about this part. If we need to read the whole file anyway, then we can just pre-allocate 1 giant buffer and launch a threadpool to fill it up in memory.

If we don't have to read the whole file, this chunk size is way to big.

I suspect you want a chunk size like 64MB to get more parallelism out of GCS though.

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,290 @@
...
+    """Wrap a synchronous AbstractFileSystem to provide async methods using a thread pool."""
+
+    def __init__(self, fs: AbstractFileSystem):
+        super().__init__()
+        self._fs = fs
+        import concurrent.futures
+
+        default_workers = min(64, max(8, (os.cpu_count() or 8) * 2))
+        max_workers = int(os.environ.get("LEVANTER_FSSPEC_MAX_WORKERS", default_workers))
+        self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
```
rjpower: This could just use the default, since you gate on the semaphore below anyway?

src/levanter/compat/fsspec_safetensor.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def _build_chunks(tensors: Iterable[TensorRecord], chunk_limit: int) -> List[ChunkSpec]:
+    if chunk_limit <= 0:
+        raise ValueError("chunk_limit must be positive")
+
+    sorted_records = sorted(tensors, key=lambda t: (t.file_path, t.byte_start))
+    chunks: List[ChunkSpec] = []
+
+    current: List[TensorRecord] = []
```
rjpower, line 142: You know this already but we'll likely spend our time reading the giant tensors out as single chunks. Doesn't seem worth optimizing that this point though.

src/levanter/compat/hf_checkpoints.py
```diff
@@ -199,23 +201,36 @@ def _load_torch(path, dtype):
...
         keys = list(f.keys())
         for key in tqdm(keys, total=len(keys), desc="Loading weights"):
             tensor_slice = f.get_slice(key)
-            d[key] = _maybe_shard_best_effort(tensor_slice, dtype)
+            d[key] = _shard_best_effort(tensor_slice, dtype)
 
     return d
 
 
+def _sharded_load_tensorstore_async(path, dtype, fs):
```
rjpower: _sharded_load_safetensors_async?

src/levanter/compat/hf_checkpoints.py
```diff
@@ -559,51 +572,80 @@ def _load_shards(self, id: str, index_file: str, rev: Optional[str], dtype) -> d
...
-                    fs.get(model_path, tmp.name)
+        if not shard_files:
+            for model_file in [SAFE_TENSORS_MODEL, PYTORCH_MODEL]:
+                model_path = os.path.join(path, model_file)
+                if fs.exists(model_path):
+                    shard_files = [model_file]
+
                     if model_file == SAFE_TENSORS_MODEL:
-                        return _load_safe_tensors(tmp.name, dtype)
+                        loader = _load_safe_tensors
```
rjpower, line 648: nit, _load_safe_tensors is now a sentinel for "_load_sharded_safe_tensors". maybe just delete it in favor of the sharded version? (This is confusing below when you test loader ==_load_sharded...)

src/levanter/compat/hf_checkpoints.py
```diff
@@ -559,51 +572,80 @@ def _load_shards(self, id: str, index_file: str, rev: Optional[str], dtype) -> d
...
 
+        for shard_file in shard_files:
+            shard_path = os.path.join(path, shard_file)
+            if not fs.exists(shard_path):
+                raise FileNotFoundError(f"Shard file {shard_path} not found")
+
+            if loader is _load_safe_tensors:
+                shard_state_dict = _sharded_load_tensorstore_async(shard_path, dtype, fs)
+            else:
+                if not warned:
```
rjpower: Maybe we can just move this tempfile logic into the torch loader itself, and then we don't need this if statement anymore? e.g. we won't have remote vs local paths per se anymore. we'll just use the sharded loader everywhere, and the torhc loader will pull remote paths over as needed.

rjpower (approved):
Thanks, much better!

src/levanter/compat/hf_checkpoints.py
dlwh: oh right. originally this was going to make virtual ts views but that was slow / i don't understanad async

src/levanter/compat/fsspec_safetensor.py
dlwh: we don't **really** need to read in the whole tensor on every worker, but every time i try to read just shards on 4 machines it's worse than doing it this way (which is why this code exists in a bit of a frankensttate. I can't quite give up on the dream of reading just the data we need, or something)

src/levanter/compat/fsspec_safetensor.py
dlwh: done

PR #982, opened 2025-06-03
add update clipping

dlwh:


src/levanter/optim/config.py
```diff
@@ -399,6 +400,12 @@ class AdamConfig(OptimizerConfig):
     beta2: float = 0.95
     epsilon: float = 1e-8
     max_grad_norm: Optional[float] = 1.0
+    update_rms_clipping: Optional[float] = None
+    """
+    If set, this will use RMS clipping on the update, a la Adafactor or StableAdamW (https://arxiv.org/pdf/2304.13013)
```
WhenWen, line 406: I think we are not actually implementing StableAdamW? We don't have the g_t^2 / u_t term if I am not mistaken. Is what we are doing closer to a plain update clipping?

WhenWen (approved):
Everything except the comment about StableAdamW looks good

src/levanter/optim/config.py
dlwh: yeah i'll fix comment

PR #998, opened 2025-06-18
Add Qwen3 support

dlwh:


Helw150 (approved):
LGTM. I also tested a roundtrip locally with Qwen3-0.6B to make sure loading pretrained models works (below sliding window length) as a sanity check, but that's probably too heavy to merge.

dlwh:
eh i think we can add it behind `skip_if_no_torch`

PR #999, opened 2025-06-18
Update Clipping

dlwh:
also contains #1000 and #1001 which i'll aim to merge first


WhenWen (approved):
The clipping part looks good to me. I left a comment in #1000

PR #1000, opened 2025-06-18
double check that we still want to delete checkpoints because deleting

dlwh:


src/levanter/checkpoint.py
```diff
@@ -213,13 +213,30 @@ def on_step(self, info, force: bool = False):
...
-                    self._rm_checkpoint(last_checkpoint)
+                    # check if we still want to delete it. Sometimes we like to replace the metadata of the last
+                    # checkpoint. It'd be nice if the process weren't manual, but this is a good compromise
+                    try:
+                        last_metadata = _load_metadata(last_checkpoint)
+                        if last_metadata.get("is_temporary", False):
+                            logger.info(
+                                f"Deleting old temporary checkpoint {last_checkpoint} after saving new checkpoint."
+                            )
+                            # we can delete the last temporary checkpoint now
```
WhenWen, line 230: Why are we deleting if the is_temporary is False? Should it be the reverse?

dlwh: hrm I think this is correct. we're deleting if is_temporary is true and not if it's False or missing?

WhenWen: Sorry I misunderstood

WhenWen (approved):
LGTM

PR #1003, opened 2025-06-21
refactor rotary embeddings in prep for inference

dlwh:


config/backpack.yaml
Helw150, line 1: RIP

src/levanter/layers/attention.py
```diff
@@ -1183,8 +1196,10 @@ def __call__(self, x: NamedArray, mask: Optional[NamedArray | AttentionMask], *,
 
         # Apply rotary position embeddings if configured
         if self.config.rope is not None:
-            rot_embs = self.config.rope.build(self.config.HeadSize, q.resolve_axis("position"))
-            q, k = rot_embs(self.config.HeadSize, q, k)
+            pos_ids = pos_ids or hax.arange(x.resolve_axis("position"))
+            rot_embs = self.config.rope.build(self.config.HeadSize)
```
Helw150: Since all the args for this are now available at init time, it feels like this should be built inside `init` rather than inside `__call__`

src/levanter/models/olmo.py
```diff
@@ -331,8 +333,10 @@ def __call__(self, x: NamedArray, mask: Optional[NamedArray | AttentionMask], *,
         v = v.rearrange((..., "kv_heads", "position", "head_size"))
 
         # Apply rotary position embeddings
-        rot_embs = self.config.rope.build(self.config.HeadSize, q.resolve_axis("position"))
-        q, k = rot_embs(self.config.HeadSize, q, k)
+        pos_ids = pos_ids or hax.arange(x.resolve_axis("position"))
+        rot_embs = self.config.rope.build(self.config.HeadSize)
```
Helw150, line 337: Same comment on moving this build function to init since it's static at inference time.

Helw150: (Side note: should we have a TODO to make OlmoAttention inherit from attention config?)

Helw150 (approved):
LGTM, a few nits.

src/levanter/models/olmo.py
dlwh: honestly i'd rather not to keep the generic Attention from getting too unwieldy. better to do some copy/paste

PR #1019, opened 2025-06-28
AdamC Implementation

Helw150:
This viral twitter paper claims to solve the problems we vaguely introduced Z-Loss to solve. 

https://arxiv.org/pdf/2506.02285

Seems worth a shot given that Kaiyue saw the norm growth for basically every optimizer and it's relatively a simple add.

### DIFF
--- a/src/levanter/optim/config.py
+++ b/src/levanter/optim/config.py
@@ -434,7 +434,7 @@ class AdamConfig(OptimizerConfig):
 
     adamc_weight_decay: bool = False
     """
-    If set, uses the AdamC correctted weight decay which keeps
+    If set, use the AdamC corrected weight decay, which keeps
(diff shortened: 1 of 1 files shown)


dlwh (approved):

src/levanter/optim/config.py
```diff
@@ -432,6 +432,15 @@ class AdamConfig(OptimizerConfig):
     See https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/optim/skip_step_optimizer.py
     """
 
+    adamc_weight_decay: bool = False
+    """
+    If set, uses the AdamC correctted weight decay which keeps
```
percyliang:
```suggestion
    If set, use the AdamC corrected weight decay, which keeps
```

PR #1020, opened 2025-06-30
Yarn

WhenWen:
Implemented 'yet another rope embedding' from https://arxiv.org/abs/2309.00071.

The hyperparameter follows the same meaning in huggingface implementation https://github.com/huggingface/transformers/blob/ccf2ca162e33f381e454cdb74bf4b41a51ab976d/src/transformers/modeling_rope_utils.py#L226

### DIFF
--- a/tests/test_rotary.py
+++ b/tests/test_rotary.py
@@ -100,7 +100,7 @@ def test_yarn_rotary_embedding():
     assert q_rotated.axes == q.axes
 
     # Test that the output is different from input (rotary embeddings should modify the input)
-    assert not hax.allclose(q, q_rotated, rtol=1e-6, atol=1e-6)
+    assert not hax.all(hax.isclose(q, q_rotated, rtol=1e-6, atol=1e-6))
 
     # Test HF config conversion
     theta, config = yarn_config.to_hf_config()


src/levanter/layers/rotary.py
```diff
@@ -180,6 +180,93 @@ def to_hf_config(self) -> tuple[float, dict]:
...
+            half_dim = self.HeadDim.size // 2
+            head_dim = self.HeadDim.size
+            HeadHalfSize = self.HeadDim.resize(self.HeadDim.size // 2)
+            inv_freq: NamedArray = 1.0 / (
+                self.config.theta ** (hax.arange(HeadHalfSize, step=2, dtype=jnp.float32) / head_dim)
+            )
+
+            # YaRN β-ramp
+            def _find_dim(n_rot: float):
+                return (half_dim * math.log(self.config.original_max_position_embeddings / (n_rot * 2 * math.pi))) / (
```
Helw150, line 200: Why use `math` here rather than `jnp`? Is there a compile time advantage I'm missing?

Helw150 (approved):
LGTM, though I'd have a preference to use jnp.pi, etc. rather than have the math import unless I'm missing something that makes this nicer for Jit.

dlwh (approved):

src/levanter/layers/rotary.py
WhenWen: Have some bug initially with jnp. Will keep it this way now

PR #1022, opened 2025-07-01
Adding PSGD-QUAD to optimizer collection

evanatyourservice:
This implements PSGD-QUAD optimizer for levanter in a similar fashion to PSGD-Kron.

QUAD has a few benefits over Kron, mostly associated with its triangular-solve-free, matmul-only preconditioner update.
- fast; preconditioner update can be done every train step
- more robust to low precision and ill-conditioned gradients
- more robust to having diagonal preconditioners, max_skew_dense (memory_save_mode in Kron) default is now 1.0 instead of float('inf'), making memory use more similar to adam.


dlwh:
test failure is on me

evanatyourservice:
ok this is looking good on my side if yall are still interested in merging :)

PR #1031, opened 2025-07-04
Simplify installation docs using uv

marcelroed:
This updates the installation docs based on the uv setup defined in #1030.

We now recommend `uv` instead of `conda` or `virtualenv`. We also recommend using `uv run` for development.

#1030 should be merged first, since this PR is branched off from that one.


dlwh (approved):
thanks!

PR #1042, opened 2025-07-08
Fixes for ray tests with uv

dlwh:


src/levanter/trainer.py
```diff
@@ -204,7 +204,7 @@ def loss_fn(self):
         def fn(model, *batch, **batch_kwargs):
             with hax.axis_mapping(self.compute_axis_mapping):
                 model = self.mp.cast_to_compute(model)
-                return _ensure_scalar(self._raw_loss_function(model, *batch, **batch_kwargs))
+                return _ensure_scalar(self._raw_loss_function(model, *batch, **batch_kwargs).mean())
```
dlwh, line 207: unrelated change to make vmapping the loss work again

pyproject.toml
```diff
@@ -121,6 +123,30 @@ markers = [
 ]
 asyncio_default_fixture_loop_scope = "function"
 
+
```
dlwh: i don't love the way this is set up RN but I'd rather fix it later

.github/workflows/run_entry_tests.yaml
marcelroed, line 1: Remove uv venv

marcelroed: And also you don't really need `uv sync` if you're using `uv run`

marcelroed: Further, it's probably better to install Python using `uv` instead of the other way around.

https://docs.astral.sh/uv/guides/integration/github/#installation

The python installation doesn't need to be added to the action, if you do `uv sync` it'll figure it out.

Also worth flipping the caching flag to save precious actions time: https://docs.astral.sh/uv/guides/integration/github/#caching

How do you test this setup currently? Happy to do it if I can get set up with whatever's needed.

dlwh: pushing repeatedly

marcelroed: lmao

dlwh: it's not possible to nicely run  Github Ci actions that do jax cpu tests on linux on an Apple Silicon Macbook...

marcelroed: I think you can with a container, but yeah it's not the nicest thing in the world

dlwh: you can't because you need AVX instructions which docker on mac still doesn't support

dlwh: (you could rebuild jaxlib w/o avx but that is way too much work)

marcelroed: I think they fixed that https://github.com/docker/for-mac/issues/6620

dlwh: empirically it does not

dlwh: ok looks like i needed to upgrade docker one more time. 4.38 didn't work but 4.43 does

marcelroed: pog

dlwh:
just gonna merge

PR #1084, opened 2025-07-21
Include Kimi style muon learning rate scaling

WhenWen:
As our scaling law in the speedrun in Marin is more pessimistic than the Kimi's version, we want to sanity check whether it is because we are using different per-module learning rate scaling.

### DIFF
--- a/src/levanter/optim/muon.py
+++ b/src/levanter/optim/muon.py
@@ -37,6 +37,9 @@ class MuonConfig(OptimizerConfig):
     epsilon: float = 1e-8
     muon_epsilon: float = 1e-8
     max_grad_norm: float = 1.0
+    # Kimi scale the learning rate for every d_1 * d_2 module by 0.2 * jnp.sqrt{\max{d_1, d_2}}, instead of the jnp.sqrt{\max{1, d_1/d_2}} as in the original nanogpt speedrun.
+    # When this scaling is enabled, it is recommended to use learning rate and weight decay similar to adam
+    use_kimi_scaling: bool = False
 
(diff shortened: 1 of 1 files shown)


src/levanter/optim/muon.py
```diff
@@ -133,7 +136,10 @@ def transform_linear_layer(layer: haliax.nn.Linear):
             array = layer.weight.array
             updated_weight_array = zeropower_via_newtonschulz5(array, steps=steps, eps=muon_eps)
 
-            scale = jnp.sqrt(jnp.maximum(1, updated_weight_array.shape[0] / updated_weight_array.shape[1]))
+            if not use_kimi_scaling:
+                scale = jnp.sqrt(jnp.maximum(1, updated_weight_array.shape[0] / updated_weight_array.shape[1]))
```
Helw150:
```suggestion
                scale = jnp.sqrt(jnp.maximum(1, updated_weight_array.shape[0] / updated_weight_array.shape[1])) # sqrt(Out/In)
```
Quick comment to enhance readability.

Aside, but in theory it would be nice if we could use named axes here to make sure this always corresponds to Out / In. 

I double checked that for all the MLPs, we have out first=True so this should be correct, but this feels like the type of code that NamedAxes are supposed to make typesafe.

dlwh: realistically optimizer people want to do their matrix stuff so we should support the flattening thing, but I agree we should double check on in/out. here it's probably easiest to just check if out_first=True

WhenWen: Will do

PR #1089, opened 2025-07-22
Improve resource management in ray_tpu

yifanmai:
Improvements:

- Create a new `SliceResource` dataclass that contains the actor, slice info, and placement group.
- Use `SliceResource` to keep the actor and slice info together, and eliminates the need to pass both the actors list and slice info dict. This also eliminates the possibility of the list and dict getting desynchronized.
- Moved ownership of the placement group to the `run_on_pod()` process.
- Moved all calls for creating and deleting placement groups to the `run_on_pod()` process. This prevents placement groups from failing to be created or deleted due to actor or network failures.
- Added timeouts to every action in the resource acquisition phase.
- Renamed a variable named `slice` becuase `slice` is a stdlib function name.
- Refactored and simplified resource cleanup logic.
- Changed actor shutdown logic to attempt to gracefully shut down the actor.

Bugfixes:

- Removed [calls](https://github.com/stanford-crfm/levanter/blob/279af9fa4afbc729357d8a82c38677904cd35a94/src/levanter/infra/ray_tpu.py#L656) to non-existent `shutdown()` method on `SliceActor`.
- Fixed [incorrect condition](https://github.com/stanford-crfm/levanter/blob/279af9fa4afbc729357d8a82c38677904cd35a94/src/levanter/infra/ray_tpu.py#L585) that could cause too few slices to be created.
- Fixed [resource leak](https://github.com/stanford-crfm/levanter/blob/279af9fa4afbc729357d8a82c38677904cd35a94/src/levanter/infra/ray_tpu.py#L593-L599) when new slice actors are created but `get_slice_info.remote()` - the actors are orphaned because they never get added to the pool, but they also never get deleted.

### DIFF
--- a/src/levanter/infra/ray_tpu.py
+++ b/src/levanter/infra/ray_tpu.py
@@ -163,7 +163,6 @@ class SliceInfo:
     num_hosts: int
     ip_address: str
     num_tpus_per_host: int
-    pg: PlacementGroup
 
     def resources_per_job(self) -> dict:
         """
@@ -188,6 +187,13 @@ class SliceInfo:
             num_gpus=0,
             memory=20e9,
         )
+
+
+# Timeouts (in seconds)
+_HEALTH_CHECK_TIMEOUT = 60
+_TERMINATE_ACTOR_TIMEOUT = 300
+_START_ACTOR_TIMEOUT = 300
+_PLACEMENT_GROUP_READY_TIMEOUT = 300
 
 
 def _multislice_info_from_head(head: SliceInfo, slice_id: int, num_slices: int) -> MultisliceInfo:
@@ -289,7 +295,8 @@ class SliceActor:
 
 @dataclass
 class SliceResource:
-    """Represents all resources associated with a slice. Not an actor."""
+    """A collection of all resources associated with a single TPU slice."""
+
     actor: ActorHandle
     """Actor handle for the SliceActor"""
 
@@ -548,11 +555,11 @@ def run_on_pod_ray(
         raise RuntimeError("Unknown error occurred during TPU job") from problem
 
 
-def _stop_actor(actor: ActorHandle, timeout: float = 60) -> None:
+def _stop_actor(actor: ActorHandle) -> None:
     try:
-        ray.get(actor.__ray_terminate__.remote(), timeout=timeout)
+        ray.get(actor.__ray_terminate__.remote(), timeout=_TERMINATE_ACTOR_TIMEOUT)
     except ray.exceptions.RayError as e:
-        logger.warning(f"Failed to gracefully shut down actor in {timeout} seconds; killing it instead: {e}")
+        logger.warning(f"Failed to gracefully shut down actor in {_TERMINATE_ACTOR_TIMEOUT} seconds; killing it instead: {e}")
         ray.kill(actor)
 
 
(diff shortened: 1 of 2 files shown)


dlwh:
ah man i thought i caught the shutdown calls but i was not being careful enoug

src/levanter/infra/ray_tpu.py
```diff
@@ -291,19 +283,28 @@ def get_slice_info(self):
...
             num_tpus_per_host=num_tpus_per_host,
-            ip_address=socket.gethostbyname(socket.gethostname()),
-            pg=pg,
+            ip_address=ip_address,
         )
 
-        ray.get(self.slice_info.pg.ready())
-        return self.slice_info
+
+@dataclass
```
dlwh: frozen? (I'm freezing everything now)

src/levanter/infra/ray_tpu.py
```diff
@@ -358,8 +359,7 @@ def run_on_pod_ray(
     num_preemptions = 0
     attempt = 0
 
-    slice_pool: list[ActorHandle] = []  # pool of SliceActors
-    slice_infos: dict[ActorHandle, SliceInfo] = {}
+    slice_pool: list[SliceResource] = []
```
dlwh, line 362: this is better!

src/levanter/infra/ray_tpu.py
```diff
@@ -567,46 +555,88 @@ def run_on_pod_ray(
...
-            try:
-                slice_infos[a] = ray.get(info_ref)
-                slice_pool.append(a)
-            except ray.exceptions.RayError as e:
-                # this can happen with a logic error or if the actor is unhealthy/preempted/whatever
-                logger.exception("Failed to get new actors ready", exc_info=e)
-                raise
+def _stop_actor(actor: ActorHandle) -> None:
+    try:
+        # This is recommended by https://docs.ray.io/en/latest/ray-core/api/doc/ray.kill.html
```
dlwh, line 559: crazy there isn't a public API method to do this

src/levanter/infra/ray_tpu.py
```diff
@@ -567,46 +555,88 @@ def run_on_pod_ray(
...
+    This function expects to be called repeatedly in an outer retry loop until it succeeds.
+
+    Returns a new pool. Does not mutate `slice_pool`."""
+    # NOTE: Do not add a retry loop, as this function will be run in an outer retry loop.
+    healthy_slices = _prune_dead_slices(slice_pool)
+    del slice_pool  # Defensively prevent mutations to slice_pool
+    if len(slice_pool) >= healthy_slices:
+        return healthy_slices
+    
+    # if we don't have enough slides, create more
```
dlwh:
```suggestion
    # if we don't have enough slices, create more
```

🛝

src/levanter/infra/ray_tpu.py
```diff
@@ -567,46 +555,88 @@ def run_on_pod_ray(
...
+            _release_slice_resource(tpu_slice)
+            continue
+        logger.info(f"Placement group for {tpu_slice.slice_info.slice_name} ready.")
+        healthy_slices.append(tpu_slice)
+        logger.info(f"Added {tpu_slice.slice_info.slice_name} to pool.")
+
+    logger.info(f"Pool ready with {len(healthy_slices)} actors.")
+
+    if len(healthy_slices) < num_slices:
+        raise Exception(f"Wanted {num_slices} slices but only acquired {len(healthy_slices)} slices.")
```
dlwh, line 640: we should be sure this gets treated as a preemption rather than a failure I think?

dlwh: nvm it's fine!

yifanmai: Yup, the outer loop treats an exception here as a preemption.

src/levanter/infra/ray_tpu.py
yifanmai: Made this frozen.

src/levanter/infra/ray_tpu.py
yifanmai: Fixed.

PR #1091, opened 2025-07-23
Implement latent attention option

Helw150:
## Summary
- add `latent_dim` to `AttentionConfig`
- extend `Attention` to optionally project keys/values from a low-rank latent representation
- test latent-dim attention

## Testing
- `pre-commit run --files src/levanter/layers/attention.py tests/test_attention.py`
- `pytest tests/test_attention.py::test_attention_latent_dim -q`
- `pytest tests -m "not entry and not slow and not ray"` *(fails: 52 failed)*

------
https://chatgpt.com/codex/tasks/task_e_688020be882483228f1a45cf720dd6f5


Helw150:
MLA from Kimi and DeepSeek - verified that the implementation looks correct. In theory, XLA should handle the collapse of the latent projection into the K and V projection on it's own, so I'm just going to have faith in the compiler that it brings the efficiency gains.

WhenWen:
> i'd kind of rather have a separate attention module? I'm just wary of this getting too complicated. WDYT?

I also feel that we should have a separate one just because the parameterization is just too different 

Also, the implementation looks sound to me.

Helw150:
Sure, I'll split it out (or ask Codex to). While I'm at it, that lets me prevent QK norm from MLA since that's not something one should do.

WhenWen:
I realize we miss two things here upon checking Deepseek's paper.

1. This is likely minor but the compressed hidden needs to go through a layernorm (see the attached graph)
image


2. MLA's absorbing trick will breakdown with RoPE so they also have this decoupled RoPE trick which I think we missed 

image

src/levanter/layers/attention.py
```diff
@@ -923,10 +983,18 @@ def flatten(axes):
...
+        )
+        h_out = flatten(
+            tuple(ax for ax in pspec_for_axis(d["H"]) if ax is not None) or None
+        )
+        s_out = flatten(
+            tuple(ax for ax in pspec_for_axis(d["S"]) if ax is not None) or None
+        )
+        d_out = flatten(
+            tuple(ax for ax in pspec_for_axis(d["D"]) if ax is not None) or None
+        )
```
dlwh: i feel like you must have a different max_columns set somehow

Helw150: This is just as a result of `uv run pre-commit run --all-files`

Helw150: But I notably have had strangeness in my local ruff so maybe I'm missing something

Helw150: Reverted manually, UV didn't do it this time so I must be missing something in my formatting process.

tests/test_latent_attention.py
```diff
@@ -0,0 +1,39 @@
...
+from levanter.layers.attention import (
+    AttentionMask,
+    MultiHeadLatentAttention,
+    MultiHeadLatentAttentionConfig,
+)
+from levanter.layers.rotary import DefaultRotaryEmbeddingsConfig
+
+
+@pytest.mark.parametrize("q_lora_rank", [None, 2])
+def test_multihead_latent_attention_shapes(q_lora_rank):
```
dlwh, line 17: anything we can do beyond testing it doesn't have shape errors? might not be worth

dlwh (commented):
lgtm but will let @WhenWen approve

tests/test_latent_attention.py
Helw150: Normally, I'd add a round trip with HF - but HF doesn't have MLA implemented in it's main library.

Helw150: The MLA is inside of custom code for DeepSeek and Kimi

dlwh:
test failures look real

Helw150:
Yeah, I realized I also need to add a method to actually integrate this with models too.

Helw150:
Ok, think I have test failures dead here now (though requires TPU test to run to confirm). Will merge this once confirmed and then add separate logic for replacing a models attention mechanism modularly.

PR #1092, opened 2025-07-23
Switch default configs to Llama

dlwh:
## Summary
- make Llama the default training architecture in `train_lm.py`, `eval_lm.py`, `viz_logprobs.py` and export script
- update scripts and CI workflows to use `llama_small_fast.yaml`
- update docs and README to reference Llama configs
- add `llama_small_fast_itest.yaml` for integration tests
- switch unit tests to Llama models

## Testing
- `pre-commit run --files $(git status --porcelain | cut -c4- | tr '\n' ' ')`
- `pytest tests -m "not entry and not slow and not ray" -q` *(failed: cannot import name 'BuiltInKeyEntry' from 'jax._src.tree_util')*

------
https://chatgpt.com/codex/tasks/task_e_68802a6c9cb08331befd5b53f84b086f

### DIFF
--- a/docs/Training-On-Your-Data.md
+++ b/docs/Training-On-Your-Data.md
@@ -254,7 +254,7 @@ Levanter supports starting from an HF pretrained model. To do so, you should set
 ```yaml
 model:
   type: mpt
-initialize_from_hf: "meta-llama/Llama-2-7b-hf"
+initialize_from_hf: "NousResearch/Llama-2-7b-hf"
 use_hf_model_config: true
 ```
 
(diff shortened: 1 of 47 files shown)


docs/Training-On-Your-Data.md
```diff
@@ -254,33 +254,20 @@ Levanter supports starting from an HF pretrained model. To do so, you should set
...
-Then, pass in the token as an environment variable:
-
-```bash
-HUGGING_FACE_HUB_TOKEN=hf...
-```
-
-Pass that in anywhere you're passing in a `WANDB_API_KEY`.
-
-Then, you can use the model like so:
+For Llama 2 specifically (or other gated models), you'll need a few extra steps. Fortunately `NousResearch/Llama-2-7b-hf` is freely accessible on the Hub, so no gating is required. You can initialize like so:
```
Helw150: This kind of seems odd to say. Maybe this? 
```suggestion
While some versions of Llama (and other gated models) require extra steps, `NousResearch/Llama-2-7b-hf` is freely accessible on the Hub, so no gating is required. You can initialize like so:
```

src/levanter/main/sft_mixture.py
```diff
@@ -71,8 +71,8 @@ class SFTMixtureConfig:
     hf_save_steps: int = 0
 
     max_seq_len: int = 2048
-    model_name_or_path: str = "meta-llama/Llama-2-7b-hf"
-    tokenizer: str = "meta-llama/Llama-2-7b-hf"
+    model_name_or_path: str = "NousResearch/Llama-2-7b-hf"
+    tokenizer: str = "NousResearch/Llama-2-7b-hf"
```
Helw150: Any reason we shouldn't use Marin 8B as the default here? It's definitely better than Llama 2 for SFT.

Helw150 (approved):
A few bits, but otherwise LGTM

src/levanter/main/sft_mixture.py
dlwh: good idea!

PR #1097, opened 2025-07-24
Switch from placement groups to tree of actors for TPU

yifanmai:


src/levanter/infra/ray_tpu.py
```diff
@@ -222,44 +193,140 @@ def _multislice_info_to_env_vars(multislice: MultisliceInfo) -> dict[str, str]:
...
+        # if we don't have enough slices, create more
+        logger.info(f"{self.get_actor_pool_name()} actor pool has {len(self._actor_pool)} members, but we want {desired_num_actors}. Creating more.")
+
+        actors = [self.create_actor() for _ in range(desired_num_actors - len(self._actor_pool))]
+
+        actors_and_actor_info_awaitables = [(actor, self.get_actor_info_future(actor)) for actor in actors]
+        logger.info(f"{self.get_actor_pool_name()} actor pool waiting for {len(actors)} new actors to start...")
+        for actor, actor_info_awaitable in actors_and_actor_info_awaitables:
+            try:
+                actor_info: ActorInfoT = ray.get(actor_info_awaitable, timeout=_START_ACTOR_TIMEOUT)  # TODO: make this overridable
```
dlwh, line 272: i think we're ok with this just taking forever fwiw. Like literally days. We use Ray as a janky batch submission thing and it's fine if there's no capacity for a while

src/levanter/infra/ray_tpu.py
```diff
@@ -284,27 +366,93 @@ def get_slice_info(self):
...
+        return not self.is_being_preempted()
+
+    def is_being_preempted(self) -> bool:
+        from levanter.infra.tpus import get_current_tpu_is_preempted
+
+        return get_current_tpu_is_preempted()
+
+    def get_host_info(self) -> TPUHostInfo:
+        self._node_id = ray.get_runtime_context().get_node_id()
+        self._num_tpus = TPUAcceleratorManager.get_current_node_num_accelerators()
```
dlwh: we might still need my hack for v4 and v5p on some versions of ray, but maybe it's fine on >=2.47 and we should just require that

dlwh (approved):
lgtm! I think we should make the slice creation timeout default to being extremely log, on the order of 24 hours.

src/levanter/infra/ray_tpu.py
yifanmai: Changed this to 7 days.

src/levanter/infra/ray_tpu.py
yifanmai: Added the hack back.

PR #1100, opened 2025-07-25
Run on Pod Should Not Run on Pre-emptible instances

Helw150:


yifanmai (approved):

src/levanter/infra/ray_tpu.py
```diff
@@ -337,7 +337,13 @@ def run_on_pod(
     return ray.get(run_on_pod_ray.remote(remote_fn, tpu_type, num_slices, max_retries_preemption, max_retries_failure))
 
 
-@ray.remote(num_cpus=0.01)
+@ray.remote(
+    num_cpus=0,
+    scheduling_strategy=NodeAffinitySchedulingStrategy(
+        node_id=ray_state.list_nodes(filters=[("is_head_node", "=", True)])[0].node_id,
+        soft=False,
```
dlwh, line 344: can we actually make this soft

PR #1116, opened 2025-07-30
Fix bugs in health checks

yifanmai:
Fix bugs:

- Only the last slice was getting health checks
- Health checks were only checked once at the start of the loop
- There was no timeout for waiting on the health checks


dlwh (approved):

yifanmai:
Partially implemented in #1126.

PR #1118, opened 2025-08-01
Add 1-sqrt averaging

dlwh:
## Summary
- add epsilon for EMA decay schedule to keep updates non-zero
- document `type:` key for configuring model averaging
- test epsilon behavior

## Testing
- `pre-commit run --files docs/reference/Configuration.md src/levanter/optim/model_averaging.py tests/test_model_averaging.py`
- `uv run pytest tests/test_model_averaging.py -m "not entry and not slow and not ray"`

------
https://chatgpt.com/codex/tasks/task_e_688bcb0a1700833181e4e940701bbbcb

### DIFF
--- a/src/levanter/optim/model_averaging.py
+++ b/src/levanter/optim/model_averaging.py
@@ -47,25 +47,32 @@ class EmaModelAveraging(ModelAveraging[M]):
 
 
 class EmaDecaySqrtModelAveraging(ModelAveraging[M]):
-    """Hybrid EMA followed by :math:`1 - \sqrt{x}` decay."""
+    """Hybrid EMA followed by :math:`1 - \sqrt{x}` decay.
+
+    This implementation keeps a running total of the weight mass so the
+    average can be queried at any step. After ``decay_steps`` updates the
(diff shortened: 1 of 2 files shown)


src/levanter/optim/model_averaging.py
```diff
@@ -45,6 +46,39 @@ def model_params(self) -> M:
...
+    def _raw_weight(self, step: int) -> float:
+        if step < self.switch_step:
+            return 1.0 - self.beta
+        t = step - self.switch_step
+        frac = jnp.clip(t / self.decay_steps, 0.0, 1.0)
+        return float(1.0 - jnp.sqrt(frac))
+
+    def update(self, new_model: M, step: int) -> "EmaDecaySqrtModelAveraging[M]":
+        w = self._raw_weight(step)
+        new_tot_w = self.tot_weight + w
```
WhenWen, line 72: I feel new_tot_weight will be more consistent with self.tot_weight but this is likely minor

WhenWen (approved):
Looks reasonable to me

PR #1124, opened 2025-08-07
Sliding Window Attention Mask and Attention Sinks

Helw150:
## Summary
- support sliding window attention via `AttentionMask` and materialization helpers
- allow `LmExample` builders to specify sliding window masks
- test causal sliding window masking
- introduce attention sink support through `AttentionWithSink` and `dot_product_attention_with_sink`
- implement sink by appending zero key/value and biasing the final logit so all backends can handle sinks without custom softmax

## Testing
- `uv run pre-commit run --all-files`
- `uv run pytest tests/test_attention.py tests/test_sharded_dataset.py::test_resolve_audio_pointer` *(fails: FileNotFoundError: https://ccrma.stanford.edu/~jos/mp3/trumpet.mp3)*

------
https://chatgpt.com/codex/tasks/task_e_689357f863f08322b2d7589f16a2ed1b

### DIFF
--- a/src/levanter/layers/attention.py
+++ b/src/levanter/layers/attention.py
@@ -1225,6 +1225,13 @@ def _tpu_splash_attention(
                 base_mask = splash_attention_mask.CausalMask(shape=(Sq, Sk))
             else:
                 base_mask = splash_attention_mask.FullMask(_shape=(Sq, Sk))
+            if mask.sliding_window is not None:
+                local_mask = splash_attention_mask.LocalMask(
+                    shape=(Sq, Sk),
+                    window_size=(mask.sliding_window - 1, None),
+                    offset=0,
+                )
+                base_mask = splash_attention_mask.LogicalAnd(base_mask, local_mask)
             # This is going to be a pain to support
             if mask.explicit_mask is not None:
                 raise NotImplementedError("Explicit masks are not yet supported for splash attention")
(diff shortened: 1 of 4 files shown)


src/levanter/layers/attention.py
```diff
@@ -220,6 +220,91 @@ def dot_product_attention(
...
+    sink_bias = sink_bias.broadcast_axis(KPos1)
+
+    if bias is not None:
+        bias = hax.concatenate(KPosPlus, [bias, sink_bias])
+    else:
+        zero_bias_axes = bias_axes_prefix + (KPos,)
+        zero_bias = hax.zeros(zero_bias_axes, dtype=sink_bias.dtype)
+        bias = hax.concatenate(KPosPlus, [zero_bias, sink_bias])
+
+    return dot_product_attention(
```
dlwh, line 286: my guess is the correct thing to do is to get the logz from attention and re-weight attention outputs with that?

dlwh (approved):

src/levanter/layers/attention.py
Helw150: Overall yes, but then I have to get into the low-level impl. of both Splash Attention (for TPU) and Flash Attention (for GPU) unless I'm missing something. Opted for keeping this more modular, but happy to change if you prefer!

Helw150:
Confirmed locally that the equivalence with the torch impl. from the GPT-OSS repo passes!

src/levanter/layers/attention.py
Helw150: Going to go ahead and merge for now, we can come back to this later if attention sinks become more core in our usage.

PR #1126, opened 2025-08-08
Variable multislice

yifanmai:
Allow `num_slices` in `run_on_pod()` to be a list of valid sizes.

When starting a job with variable multitask:

- The scheduler attempts to acquire the maximum number of slices requested
- If this does not work, it uses the largest valid size that is <= the actual number number of slices acquired, and releases unneeded slices

Every 15 minutes, and at least 60 minutes after the last scaling:

- The scheduler attempts to acquire enough slices to go up to the next larger valid size
- If this works, it preempts the job and restarts the job, following the same procedure as the job starting procedure above
- Otherwise, it releases the newly-acquired slices and does not disrupt the job.

Other changes:

- Replace `get_slice_info()` and `get_host_info()` with `get_info()` to simplify things
- When waiting for multiple actors, use `ray.wait()` to wait `TIMEOUT` seconds rather than `len(actors) * timeout`
- Clean up logging
- Handle exceptions thrown by actor exit correctly
- Fix a bug where the main loop could stall forever waiting for health checks
- Shorten start actor timeout


src/levanter/infra/ray_tpu.py
```diff
@@ -162,9 +162,16 @@ class TPUHostInfo:
...
 # Timeouts (in seconds)
 _HEALTH_CHECK_TIMEOUT = 60
-_TEARDOWN_ACTOR_TIMEOUT = 300
-_TERMINATE_ACTOR_TIMEOUT = 300
-_START_ACTOR_TIMEOUT = 7 * 24 * 60 * 60  # 1 week
+_TEARDOWN_ACTOR_TIMEOUT = 60
+_TERMINATE_ACTOR_TIMEOUT = 60
+_START_ACTOR_TIMEOUT = 60
+# NOTE: It is good to have a short timeout here because
+# cluster autoscaling happens outside the this timeout.
```
dlwh: wait is that true though? Autoscaling happens in response to demand, which create_actor is the source of? (I agree the preempted loop mitigates but I don't love it being used for this since the length of time it takes to spin up a node is on the order of ~5-10 min even if capacity is there)

src/levanter/infra/ray_tpu.py
```diff
@@ -255,45 +259,49 @@ def _remove_unhealthy_members_from_actor_pool(self) -> None:
...
+    def _remove_members_from_actor_pool(self, desired_num_actors: int) -> None:
+        logger.info(f"{self.get_actor_pool_name()} actor pool has {len(self._actor_pool)} members; scaling down to {desired_num_actors} members; current members: {[self.get_actor_name_from_actor_info(member.actor_info) for member in self._actor_pool]}")
+        members_to_remove = self._actor_pool[desired_num_actors:]
+        self._actor_pool = self._actor_pool[:desired_num_actors]
+        for member in members_to_remove:
+            logger.info(f"{self.get_actor_pool_name()} actor pool member {self.get_actor_name_from_actor_info(member.actor_info)} stopping.")
+            _stop_actor(member.actor)
+        logger.info(f"{self.get_actor_pool_name()} actor pool scaled down to {len(self._actor_pool)} members: {[self.get_actor_name_from_actor_info(member.actor_info) for member in self._actor_pool]}")
 
     def scale_actor_pool(self, desired_num_actors: int) -> None:
```
dlwh: protected maybe?

src/levanter/infra/ray_tpu.py
```diff
@@ -312,9 +331,83 @@ def get_actor_name_from_actor_info(self, actor_info: SliceInfo) -> str:
...
+        # Want to scale to the next largest desired size that is bigger than the current size.
+        sorted_valid_sizes = sorted(num_slices)
+        current_size = current_size = len(self._actor_pool)
+        larger_sizes = [size for size in sorted_valid_sizes if size > current_size]
+        if not larger_sizes:
+            return False
+        next_larger_size = larger_sizes[0]
+
+        # We start N slice probe actors to probe for available slices.
+        # We shut down these actors after we're done checking.
```
dlwh: is this better than just sending them directly?

src/levanter/infra/ray_tpu.py
```diff
@@ -497,14 +585,13 @@ def run_on_pod_ray(
 
     This function is a Ray remote function that can be called from anywhere in the Ray cluster.
     """
-    if num_slices <= 0:
-        raise ValueError("num_slices must be greater than 0")
+    if not num_slices:
```
dlwh: probably doesn't matter but this doesn't check non-negativity anymore

dlwh (commented):
generally lgtm, though I'm not sure about the timeout changes?

src/levanter/infra/ray_tpu.py
yifanmai: Need to update this comment. If we are successful, we keep the actors. If we're not, we shut down the actors.

src/levanter/infra/ray_tpu.py
yifanmai: wdym?

src/levanter/infra/ray_tpu.py
yifanmai: I can't find documentation for this, but this is what I noticed empirically. When you spawn an actor and there isn't enough resources for it, it will block on trying to get resources from the autoscaler, and I noticed it waiting for much longer than whatever the timeout is here. I can investigate this further before merging.

src/levanter/infra/ray_tpu.py
yifanmai: reverted

src/levanter/infra/ray_tpu.py
yifanmai: Changed the start actor timeout to 30 minutes. Also changed the scale up check interval to 3 hours, and the scale up interval to 12 hours.

src/levanter/infra/ray_tpu.py
yifanmai: or actually I changed this to a more correct validation function

src/levanter/infra/ray_tpu.py
yifanmai: Updated comment

src/levanter/infra/ray_tpu.py
dlwh: `_scale_actor_pool` but I don't care really. Looks like maybe a decision I made?

tests/test_ray_tpu.py
```diff
@@ -183,6 +183,30 @@ def test_multislice_simple_run():
...
+
+    assert results is not None
+    assert len(results) in num_slices  # num_slices * hosts_per_slice (assuming 1 host per v4-8 slice)
+
+    for i in range(len(results)):
+        assert isinstance(results[i], np.ndarray)
+        assert results[i].shape == (4,)
+        if i > 0:
+            # Due to MEGASCALE_SLICE_ID, the PRNG key might differ effectively if the code used it.
+            # simple_jax_fn uses a fixed PRNGKey(0) so all slices should produce identical results.
```
dlwh:
```suggestion
```

dlwh:
thanks!

src/levanter/infra/ray_tpu.py
yifanmai: Done. This was a public method that was called from `run_on_pod_ray()` before, but now `scale_multislice()` is the replacement public method, so this can be private

PR #1139, opened 2025-08-22
Add Trackio tracker

dlwh:
## Summary
- add Trackio tracker with mode control and auto-resume
- disable logging on non-primary processes by default
- run Trackio tests offline and include Trackio in dev dependencies

## Testing
- `pre-commit run --all-files`
- `uv run pytest tests -m "not entry and not slow and not ray"` *(fails: ProxyError: HTTPSConnectionPool errors)*

------
https://chatgpt.com/codex/tasks/task_e_68a686162aa48331bb0d67c61df07daa


src/levanter/tracker/trackio.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import annotations
```
Helw150, line 4: Is there a way we could obviate the need for a separate class which seems subject to bitrot by using the plug-and-play wandb support?

They suggest just doing `import trackio as wandb` which should let us just use our wandb tracker directly?

dlwh: i would actually rather keep it this way since I suspect that it won't actually be fully compatible despite claims (missing features, bugs), and the amount of duplication that isn't boilerplate is pretty low

dlwh: just looking at the issue list: https://github.com/gradio-app/trackio/issues

PR #1140, opened 2025-08-23
Correctly handle num_cores < 8 and fail if no futures are scheduled.

jyc:
Currently, when num_cores for a slice < 8, num_hosts = 0 and we no-op. This causes jobs to report that they succeeded even when remote_fn ends up never getting run.

To fix this, use num_cores = max(1, num_hosts // 8).

Also fail when futures = [] instead of reporting success. (Is that correct?)

### DIFF
--- a/src/levanter/infra/ray_tpu.py
+++ b/src/levanter/infra/ray_tpu.py
@@ -451,7 +451,8 @@ class SliceActor(ResourcePoolManager[TPUHostInfo]):
         if tpe.startswith("v4") or tpe.startswith("v5"):
             num_cores = int(tpe.split("-")[1])
             num_tpus_per_host = 4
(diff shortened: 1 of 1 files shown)


src/levanter/infra/ray_tpu.py
```diff
@@ -451,7 +451,8 @@ def get_info(self) -> SliceInfo:
         if tpe.startswith("v4") or tpe.startswith("v5"):
             num_cores = int(tpe.split("-")[1])
             num_tpus_per_host = 4
-            num_hosts = num_cores // 8
+            # "v5litepod-4" should still create 1 host, not 0
+            num_hosts = max(1, num_cores // 8)
```
dlwh, line 455: oops sorry

dlwh (approved):

PR #1142, opened 2025-08-25
Add PolyNorm used by Motif …

Helw150:
This new SmolLM from a Korean company uses PolyNorm activation (https://arxiv.org/abs/2411.03884) which uses learned weights and biases to allow a trainable activation function to be learned across layers. The original paper and the Motif paper claim that PolyNorm enables more efficient learning than SwiGLU. Not sure if I expect this to be stable at scale, but seems like a good thing to Speedrun!

https://huggingface.co/Motif-Technologies/Motif-2.6B/blob/main/modeling_motif.py


WhenWen (approved):
This implementation looks good to me. I am interested to see how this will work out in speedrun

PR #1143, opened 2025-08-26
Modular Transformer Draft

Helw150:
It's currently quite hard to do Speedrun of architecture changes in Marin since it basically requires you to codevelop with Levanter. This PR is a pass at implementing a more modular Transformer which should allow people to define custom attentions, activations, or full Transformer layers within Marin and pass them modularly to Train LM. Things that are promising in the Speedrun can eventually be mainlined into Levanter!

### DIFF
--- a/src/levanter/models/modular_transformer.py
+++ b/src/levanter/models/modular_transformer.py
@@ -1,0 +1,323 @@
+from __future__ import annotations
+
+import dataclasses
+from dataclasses import dataclass
+from typing import Optional, Type
+
+import equinox as eqx
+import jax.random as jrandom
+from jaxtyping import PRNGKeyArray
+
+import haliax as hax
+import haliax.nn as hnn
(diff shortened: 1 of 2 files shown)


src/levanter/models/modular_transformer.py
```diff
@@ -0,0 +1,341 @@
...
+    ) -> NamedArray:
+        k_attn, k_mlp = maybe_rng_split(key, 2)
+
+        residual = x
+        x = self.input_layernorm(x)
+        attn_output = self.self_attn(x=x, mask=mask, key=k_attn, pos_ids=pos_ids)
+        x = residual + attn_output
+
+        residual = x
+        x = self.post_attention_layernorm(x)
```
WhenWen, line 69: I don't know whether this is consistent with what we used in Levanter but it feels slightly confusing because this is actually a pre(MLP)-norm instead of a post(Attention)-norm

PR #1146, opened 2025-09-05
Fix: Correct implementation in scion.py

Wanyun0327:
Hello! Thank you for the great work and for implementing the optimizer from our paper, "Training Deep Learning Models with Norm-Constrained LMOs".

I noticed a few minor discrepancies between the current implementation and the details described in our paper. I've submitted this pull request to align the code more closely with the original algorithm. Detailed comments and explanations have also been added to the top of the optim/scion.py file.

1. Changed the scaling factor calculation in the `transform_linear_layer` function to use sqrt(d_out/d_in), removing the conditional maximum with 1. 
2. Added the scaling factor 1/d_in in `scale_by_sign`.
3. Added scaling factors `spectral_radius` and `sign_radius` to allow for additional scaling after orthogonalization.
4. Removed weight decay, since Scion does a convex combination with learning rate directly.
5. Removed scion_to_signum_lr as its role is replaced by the scaling factors `spectral_radius` and `sign_radius`.
6. Added BiasRMS_transform that applies RMS normalization to bias parameters.

### DIFF
--- a/src/levanter/optim/scion.py
+++ b/src/levanter/optim/scion.py
@@ -1,3 +1,20 @@
+# Revised by Wanyun Xie (wanyun.xie@epfl.ch)
+# 1. Changed the scaling factor calculation in the `transform_linear_layer` function to use sqrt(d_out/d_in), removing the conditional maximum with 1. 
+# 2. Added the scaling factor 1/d_in in `scale_by_sign`.
+# 3. Added scaling factors `spectral_radius` and `sign_radius` to allow for additional scaling after orthogonalization.
+# 4. Removed weight decay, since Scion does a convex combination with learning rate directly.
+# 5. Removed scion_to_signum_lr as its role is replaced by the scaling factors `spectral_radius` and `sign_radius`.
+# 6. Added BiasRMS_transform that applies RMS normalization to bias parameters.
+
+# Other tips:
+# - Note that we usually set `sign_radius` larger than `spectral_radius`. This is opposite to the original setting `signum_lr = learning_rate * 0.25`.
+# - Tuning lr and spectral_radius/sign_radius is important and can be transferred from a small proxy model. `spectral_radius=50` and `sign_radius=3000` is what we set for modded-nanogpt, not for this one.
+# - Could choose constrained or unconstrained version of Scion by setting `unconstrained` flag.
+# - Not sure how necessary to set different momentum for Spectral and Sign.
+# - Not sure how necessary to set `max_grad_norm` for Sign.
+# - It's okay to still keep Sign for bias parameters, but without 1/d scaling as we suggested in Table 6 of Scion paper https://arxiv.org/abs/2502.07529.
+
+
 import dataclasses
 from dataclasses import dataclass
 from typing import NamedTuple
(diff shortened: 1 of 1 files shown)


WhenWen:
Looking at the PR, is the following understanding correct that 'spectral radius' and 'sign radius' is a reparameterization of the learning rate and weight decay relationship? 

For the linear layer, if we ignore the difference in {sqrt(d_out/d_in)} and \max{1, {sqrt(d_out/d_in)}} for now. And assume the orthogonalized momentum after scaling is $O$ and the weight is $W$.

The original version apply the following update $-\eta O - \eta \lambda W$ for learning rate $\eta$ and weight decay $\lambda$.

The current version apply the following update $- \eta' r O - \eta' W$ for learning $\eta'$ and spectral radius $r$ for constrained Scion and  $- \eta' r O$ for unconstrained Scion

So it seems like there is a matching between the old version and the current version where we choose
1. $\eta = \eta' r$
2. $\lambda = 1/r$ for constrained Scion or  $\lambda = 0$ for unconstrained Scion. 


For the layer where we apply sign, if the signed momentum is $S$ and the weight is $W$.

The original version apply the following update $-\eta S - \eta \lambda W$ for learning rate $\eta$ and weight decay $\lambda$.

The current version apply the following update $- \eta' r S / d_in - \eta' W$ for learning $\eta'$ and sign radius $r$ for constrained Scion and  $- \eta' r O$ for unconstrained Scion

So it seems like there is a matching between the old version and the current version where we choose
1. $\eta = \eta' r / d_in$
2. $\lambda = d_in / r$ for constrained Scion or  $\lambda = 0$ for unconstrained Scion. 


The current default seems to translate to 
1. weight decay of $0.02$ and a learning rate $50\eta'$ for the linear layer 
2. weight decay $0.002 * d_in$ and a learning rate $500 \eta' / d_in$ for the embedding layer. 


As d_in are certainly much larger than $10$, it seems like our original setups where we set the learning rate to be lower for the embedding is not contradictory to the fact that sign radius is higher than spectral radius. 

Given this translation relationship, I am wondering whether we should consider having two ways of setting the hyperparameters for Scion in the codebase: 
1. the current way which is faithful to the paper
2. uses learning rate and weight decay with the same semantic meaning as other optimizers in the library to allow easier hyperparameter transfer from optimizers like Muon. We can incorporate decoupled weight decay for embedding and weight matrices (something similar to https://github.com/stanford-crfm/levanter/pull/1147).

Wanyun0327:
Hi! Thanks for the quick reply.

The relationship you've described among learning rate, weight decay, and radius is correct. It's a good idea to keep two versions. 

I would like to emphasize two key points:
1. It is crucial for Scion to decouple the weight decay or radius for embedding and weight matrices. Using different learning rates for these matrices does not achieve the same effect as using different weight decays.
2. Scaling factors such as $\sqrt\frac{d_{out}}{d_{in}}$ and $\frac{1}{d_{in}}$ are important for preserving the ability to transfer learning rates.

Please let me know if there's anything else I can help with.

Wanyun0327:
Hi! I just noticed a minor error in `Sign_transform`. "Embedding" shouldn't be scaled by $\frac{1}{d_{in}}$ while "lm_head" should be scaled by $\frac{1}{d_{in}}$. Thus, I revised the code accordingly. 

About `sign_radius` for these two layers, they might need to be set separately. I'm not very sure about this since in our modded-nanogpt code, we follow a version that includes only one `sign_radius` since the first and last layers share weights.

Another choice for the first layer is `ColNorm` with the update $\sqrt{d_{out}}\frac{col_j(W_1)}{||col_j(W_1)||_2}$. You could check the detailed selection in Table 6 of Scion paper.

WhenWen (approved):

PR #1147, opened 2025-09-05
Muon: Decouple muon's and its adamw part's weight decay

leloykun:
Decouples muon's and its adamw part's weight decay.

- Learning rates and weight decay are tightly linked and should be sweeped jointly.
- Adam's updates tend to have size `0.23 * adam_lr` RMS while Muon's updates have size `lr` RMS (see Kimi's scaling factor). Different effective update sizes means we should, ideally, use different weight decays for the weights optimized by Muon vs. the weights optimized by AdamW.


WhenWen (approved):
Lgtm! We will move on to add a more fine-grained sweeping for this adam weight decay

leloykun:
Thanks @WhenWen ! It does look like I need another approval from you before I could merge this.

WhenWen (approved):
lgtm. Will test small lm head wd later in Marin

PR #1148, opened 2025-09-08
[WIP] Slightly less ambiitious inference

dlwh:


dlwh:
@codex can you make a substantive review

src/levanter/inference/page_table.py
```diff
@@ -0,0 +1,388 @@
...
+
+    # ------------------------------------------------------------------
+    # Sequence management
+    # ------------------------------------------------------------------
+    @eqx.filter_jit(donate="all")
+    def assign_seq_id_to_seq(self) -> tuple["PageTable", int]:
+        seq_id = hax.argmax(self.seq_lens, "seq").scalar()
+        # Error handling: if there are no sequences available, return INVALID
+        seq_id = hax.where(is_invalid(self.seq_lens["seq", seq_id]), seq_id, INVALID)
+        new_seq_lens = hax.where(
```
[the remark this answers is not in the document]
dlwh: INVALID is 2000000 you silly bot

dlwh:
@codex can you look again. please review very carefully. there's lots to get right/wrong.

src/levanter/models/flash_attention.py
```diff
@@ -175,6 +175,11 @@ def _flash_attention_forward(
     ell = hax.auto_sharded(ell)
 
     is_causal = isinstance(mask, AttentionMask) and mask.is_causal
+    # Non-zero offsets (dynamic or otherwise) are not supported in this implementation
+    if isinstance(mask, AttentionMask) and mask.is_causal and mask.causal_offset is not None:
+        raise NotImplementedError("Non-zero causal offsets are not implemented for flash attention")
```
rjpower, line 180: Dumb question, would it make sense for the attention mask to provide `materialize` and `materialize_with_offset`, and then the attention impl would call the appropriate one?

src/levanter/data/packing.py
```diff
@@ -176,6 +176,9 @@ def per_segment_loss(
     assert packed_example.attn_mask.segment_ids is not None, "segment_ids must be set in the AttentionMask"
 
     segment_ids = packed_example.attn_mask.segment_ids
+    if isinstance(segment_ids, tuple):
```
rjpower, line 179: Is segment_ids always a 1 element tuple in this case?

rjpower: (should we always keep it as a tuple to avoid the checks?)

src/levanter/layers/attention.py
```diff
@@ -1325,38 +1474,37 @@ def init(config: AttentionConfig, *, key) -> "Attention":
 
         return Attention(config, q_proj, k_proj, v_proj, o_proj, q_norm, k_norm, rot_embs)
 
+    def empty_cache(self, Batch: Axis, MaxLen: Axis, *, dtype):
```
rjpower: nit empty_kv_cache for consistency?

src/levanter/layers/attention.py
```diff
@@ -1374,18 +1522,654 @@ def __call__(
...
+
+    is_special_case = jnp.all(new_lengths <= 1)
+
+    updated_keys, updated_values = jax.lax.cond(is_special_case, specialized_path, general_path, operands)
+    updated_lengths = lengths + new_lengths
+
+    return updated_keys, updated_values, updated_lengths
+
+
+class KvPageCache(eqx.Module):
```
rjpower, line 1626: I have I had some trouble understanding the logic for the kvcache vs kvpagecache. I naively expected them to share some interface, but it seems like they're completely separate beasts instead.

Not actionable except as a general thought for how maybe paged decoding could be clearly separated out.

src/levanter/layers/attention.py
```diff
@@ -784,39 +839,79 @@ def materialize(
...
             sliding_window: If provided, restrict each query position to attend only to keys within
                 ``sliding_window`` previous positions.
+            For ``offset == 0`` this is identical to the old ``AttentionMask.causal()``
+            behaviour; larger offsets loosen the restriction so that each query can
+            see ``offset`` additional future tokens.
         """
-        return AttentionMask(is_causal=True, sliding_window=sliding_window)
+        if isinstance(offset, int | Integral):
+            causal_offset = hax.named(offset, ())
+        else:
```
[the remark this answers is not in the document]
dlwh: fake news

src/levanter/models/flash_attention.py
dlwh: hrm, well there's kind of two different notions of offset one might want here. materialize optionally takes a slice of both qpos and kpos and will slice the matrix that way. but you can have other offsets too. I dunno. maybe your idea is the right thing.

src/levanter/data/packing.py
dlwh: yeah probably. was trying to preserve some compat but it's mostly a pain

dlwh: will do a follow up pr

src/levanter/layers/attention.py
dlwh: i'm gonna delete this one actually

src/levanter/layers/attention.py
dlwh: i'm gonna delete pagecache

dlwh: err, i mean kvcache

PR #1158, opened 2025-09-12
WIP: Add JIT-ed UL2R functions

jyc:
- **Task types:** R, X, S (the task mix is given in terms of task types) TODO maybe clean this up & make type either rx/s, then just call these task config names?
- **Task kinds:** R- and X- denoising have the same implementation, just different parameters, so they have the same kind.
  S-denoising is its own kind.
  We encode task parameters for access by JAX as a tensor `task_params[task_idx] = [task_kind, task_token_id, mask_prob, mean_span_length]`.
- `to_ul2r_tokens`: Dispatches to `to_ul2r_rx_tokens` (span corruption) or `to_ul2r_s_tokens`.
- `ul2r_loss_mask`: Creates loss mask (loss computed only on outputs, not inputs/padding)
- `ul2r_block_diagonal_mask`: Creates prefix attention mask (inputs attend bidirectionally within segment)
- `Ul2rDataset`: Packs / reserves space using `GreedyPrepackedDataset`; denoises using `create_ul2r_example`

## Pipeline 

1. Compute denoising lengths
2. Reserve space in GreedyPrepackedDataset: Documents remain contiguous; padding reserved at end only.
3. Create denoising examples in loop: For each segment, `process_segment` rolls it to position 0, applies `to_ul2r_tokens` creating `0...<task_token><denoised>...0...`.
   OR these together: `task_token1 denoised1 task_token2 denoised2 ...`.
   Generate `LmExample` with `ul2r_loss_mask` and `ul2r_block_diagonal_mask` (ORed with causal mask).

## Data 

1. **(NumPy)** `doc1 doc2 ...` from TreeCache
2. **(NumPy)** `doc1 doc2 0...` packed contiguously in `GreedyPrepackedDataset` with reserved space at end
3. **(JAX)** For each segment i, `process_segment` (ul2r.py:749-781):
   - Rolled input: `doc rest`
   - After `to_ul2r_tokens`: `task_token denoised 0...`
   - Rolled to `out_start`: `0... task_token denoised 0...`
4. **(JAX)** OR all segments together: `task_token1 denoised1 task_token2 denoised2 ...`

## JAX

Had to modify implementations to work in JIT-ed functions.
For example, the original `np.pad(np_rng.permutation(...))` pattern doesn't work when `num_items` and `num_segments` depend on example length.

- `random_segmentation_jax` and `random_spans_noise_mask_jax` won't produce identical results to their NumPy equivalents for the same key
- `noise_span_to_unique_sentinel` should behave identically (modulo padding)

`to_ul2r_rx_tokens` corresponds to [this code](https://github.com/stanford-crfm/levanter/pull/220/files#diff-45b8ddf66815ef744e8a94a8778959160234fb7fe54edead964f615f967c575cR166-R177).

## TODOs

- [ ] Try generate some data using `cache_dataset` (debugging config errors right now)
- [ ] Do we need to pass in `KPos` as well as `Pos` (aka QPos?) `text.py` doesn't use it, we would just use it for the segment IDs in AttentionMask?
- [ ] Do we need `eos_id`
- [ ] Do we need to make sure we don't mask out bos/eos_id or have the prefix/continuation consist entirely of them?

---

- [ ] Make prefix attention work with splash attention
- [ ] Implement versions both with and without PrefixLM; disable better attention implementations when prefix is enabled (from @dlwh)
- [ ] Later: avoid `GreedyPrepacked` holding O(n) tensors for entire dataset (iterate over chunks to trade space for time?)

---

*Note: Not attached to any implementation details, names, etc. Feedback from JAX experts very welcome!*

### DIFF
--- a/config/llama2_7b_ul2r.yaml
+++ b/config/llama2_7b_ul2r.yaml
@@ -3,10 +3,10 @@ data:
   configs:
     pile:
       id: dlwh/wikitext_103_detokenized
-      cache_dir: /tmp/marin_cache
+      cache_dir: /tmp/marin_cache_pile
     ul2r:
       id: dlwh/wikitext_103_detokenized
-      cache_dir: /tmp/marin_cache
+      cache_dir: /tmp/marin_cache_ul2r
       format:
         type: ul2r
         task_configs:


src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
...
+    segment_id = jnp.cumsum(first_in_segment)
+    segment_length = jax.ops.segment_sum(
+        jnp.ones_like(segment_id), segment_id, padded_length
+    )
+
+    return segment_length
+
+
+@functools.partial(
+    jax.jit, static_argnames=["mean_noise_span_length", "random_roll", "padded_length"]
```
dlwh: i don't think there's any reason for mean_noise_span_length to be static and if you're going to use `cond` anyway, no reason for `random_roll` to be static

src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
+import functools
+import jax
+import jax.numpy as jnp
+from jax.random import PRNGKey
+from jax.experimental import checkify
+
+
+@functools.partial(jax.jit, static_argnames=["padded_length"])
+def random_segmentation_jax(
```
dlwh: can you link to references for these (t5x or whatever)

src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+        rolled = jnp.where(indices < length, rolled, False)
+        return rolled
+
+    mask = jax.lax.cond(random_roll, apply_roll, lambda m: m, is_noise)
+
+    return mask
+
+
+@functools.partial(jax.jit, static_argnames=["pad_token_id"])
```
dlwh: probably fine for pad token to not be static

src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
...
+def noise_span_to_unique_sentinel_jax(
+    tokens: jnp.ndarray,
+    noise_mask: jnp.ndarray,
+    sentinel_tokens: jnp.ndarray,
+    length: int,
+    pad_token_id: int,
+) -> jnp.ndarray:
+    """
+    Replace each run of consecutive noise tokens with a different sentinel.
```
dlwh, line 300: maybe document that `length` is the true length?

src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
...
+    return result
+
+
+@functools.partial(
+    jax.jit,
+    static_argnames=[
+        "max_length",
+        "pad_token_id",
+        "mean_noise_span_length",
+        "random_roll",
```
dlwh: don't think these need to be static

src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
...
+
+    return loss_mask
+
+
+@jax.jit
+def ul2r_block_diagonal_mask(
+    input_masks: jnp.ndarray, segment_ids: jnp.ndarray
+) -> jnp.ndarray:
+    """
+    - Input tokens can attend to all other input tokens bidirectionally within
```
dlwh: it is probably convenient to have a version that is still causal b/c it will compose nicer with splash attention. (Or we'll need to write our own kernel...)

src/levanter/layers/attention.py
```diff
@@ -678,6 +678,10 @@ def materialize(
             segment_mask = _materialize_segment_mask(self.segment_ids, QPos, KPos, q_slice, k_slice)
             mask = combine_masks_and(mask, segment_mask)
 
+        if self.prefix_mask is not None:
```
dlwh: can you add code to raise if prefix_mask is encountered in splash_attention, flash_attention [which can probably handle this with a bit of work, but not a ton], or the nvidia path

dlwh:
@codex can you look too

src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
...
+    max_segments = jnp.max(segments) + 1  # Number of unique noise spans
+
+    # Check if we have too many noise spans
+    err, num_sentinels = checkify.checkify(
+        lambda: checkify.check(
+            max_segments <= len(sentinel_tokens),
+            f"Too many noise spans: {max_segments} > {len(sentinel_tokens)}",
+        ),
+        errors=checkify.index_checks,
+    )()
```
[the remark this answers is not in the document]
jyc: ah interesting, I guess I can ignore the return but then unless the caller wraps this function with `checkify.checkify` nothing happens? https://docs.jax.dev/en/latest/_autosummary/jax.experimental.checkify.check.html

```
import jax
import jax.numpy as jnp
from jax.experimental import checkify
def f(x):
  checkify.check(x>0, "{x} needs to be positive!", x=x)
  return 1/x
checked_f = checkify.checkify(f)
err, out = jax.jit(checked_f)(-3.)
err.throw()  
  ...
jax._src.checkify.JaxRuntimeError: -3. needs to be positive!
```

I don't see it used anywhere else so maybe not worth it, but will just remove `err` value for now.

src/levanter/data/ul2r.py
jyc: Haha thanks for pointing all these out, good to learn. ~I guess for arguments that will have the same shape in different invocations it's not generally not worth making it static unless that somehow lets the compiler optimize control flow?~ (edit: I think I'm wrong)

jyc: oh hm yea I guess curious about the intuition around static! The docs say:

> Calling the jitted function with different values for these constants will trigger recompilation. Arguments that are not array-like or containers thereof must be marked as static.

but I definitely don't have the full context

src/levanter/data/ul2r.py
```diff
@@ -0,0 +1,342 @@
...
+    inputs = noise_span_to_unique_sentinel_jax(
+        tokens, noise_mask, sentinel_token_ids, length, pad_token_id
+    )
+    targets = noise_span_to_unique_sentinel_jax(
+        tokens, ~noise_mask, sentinel_token_ids, length, pad_token_id
+    )
+
+    indices = jnp.arange(padded_length)
+    input_len = jnp.argmax(inputs == pad_token_id)
+    target_len = jnp.argmax(targets == pad_token_id)
```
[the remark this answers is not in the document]
jyc: huh, nice!

src/levanter/data/ul2r.py
jyc: Ah interesting, so this function should generate the full PrefixLM mask instead of just the block diagonal part for the prefixes? I was thinking we could take this mask and or it with a causal mask but don't know much about splash attention, will look up the implementation in the meantime!

src/levanter/data/ul2r.py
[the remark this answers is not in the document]
dlwh: i don't use checkify so i dunno. i use eqx.error_if

src/levanter/data/ul2r.py
dlwh: well i guess we'd want to `and` it to get causal but yeah i guess that would work

PR #1159, opened 2025-09-12
Implement OpenAI compatible inference service.

rjpower:
Features:

- OpenAI compatible service
- On demand checkpoint reloading

Current limitations:

- Multi-host support

Multiple hosts may or may not work. They'll definitely start their own inference server port, to guarantee maximum chaos.

- Checkpoint sharding is not totally clear

This should be updated to use a global barrier to ensure workers are aligned with the leader etc. This shouldn't be an issue however, as we should always be using the lead node for submitting requests

- Model selection is ignored.
This is intentional, as we will in general be loading models explicitly at startup or via a checkpoint service.

- No cancellation support.

This is a nice to have, for example if we want to cancel some queries when loading new weights or if they are slow.

- Missing performance for multi-turn queries.
Some kind of persistent KV-caching will be necessary for multi-turn (agent) RL testing. See https://github.com/jax-ml/jax-llm-examples/blob/main/serving/serving_jax/serving_loop.py for an example.

### DIFF
--- a/src/levanter/inference/engine.py
+++ b/src/levanter/inference/engine.py
@@ -788,10 +788,10 @@ class InferenceEngine:
         Each Request provides prompt_tokens, decode_params, and n_generations (clones).
         """
         # validate we don't have any sequences with n_generations exceeding max_seqs
-        total_needed = sum(int(r.n_generations) for r in requests)
-        if total_needed > int(self.gen_state.decode_state.page_table.max_seqs):
+        max_needed = max(int(r.n_generations) for r in requests)
+        if max_needed > int(self.gen_state.decode_state.page_table.max_seqs):
             raise ValueError(
-                f"Total sequences needed ({total_needed}) exceeds max_seqs ({self.gen_state.decode_state.page_table.max_seqs})."
+                f"Total sequences needed ({max_needed}) exceeds max_seqs ({self.gen_state.decode_state.page_table.max_seqs})."
                 "Decompose your request into smaller batches or increase max_seqs when building the service."
             )
 
--- a/tests/inference/test_engine.py
+++ b/tests/inference/test_engine.py
@@ -1,18 +1,17 @@
 # Copyright 2025 The Levanter Authors
 # SPDX-License-Identifier: Apache-2.0
 
+import logging
+
+import haliax as hax
 import jax
 import jax.numpy as jnp
-import haliax as hax
-
+import pytest
 from haliax import Axis
-
 from levanter.inference.engine import InferenceEngine, Request
 from levanter.inference.jit_scheduler import SeqDecodingParams
-from levanter.layers.attention import KvPageCache
 from levanter.inference.page_table import PageTable
-import pytest
-import logging
+from levanter.layers.attention import KvPageCache
 
 
 class DummyModel:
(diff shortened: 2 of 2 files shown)


dlwh:
the openai logprob api is kind of maximally annoying

src/levanter/main/sample_lm.py
```diff
@@ -84,90 +633,9 @@ def _load_model(config: SampleLmConfig, Vocab: Axis, *, key) -> LmHeadModel:
...
-                text = tokenizer.decode(seq_outputs, skip_special_tokens=True)
-                print(f"Tokens for sequence {seq_id} (len: {len(seq_outputs)}: {seq_outputs}")
-                print(f"Generated text for {seq_id}: {text}")
-                tokens_with_text = [
-                    f"{tok} ({tokenizer.decode([tok], skip_special_tokens=True)})" for tok in seq_outputs
-                ]
-                print(f"Tokens with text for sequence {seq_id}: {tokens_with_text}")
+    """Start the FastAPI inference server"""
+    server = InferenceServer.create(config)
+    server.serve(host="0.0.0.0", port=8000)
```
dlwh: can we keep sample_lm being batch?

rjpower:
Yeah I'll look into getting that wired up today.

dlwh (approved):

src/levanter/main/sample_lm.py
rjpower: reverted to original.

src/levanter/inference/engine.py
```diff
@@ -767,12 +788,31 @@ def _prefill_prompts(
...
 
-    def generate(self, requests: Sequence[Request]) -> tuple[list[list[int]], int]:
+    def generate(self, requests: Sequence[Request]) -> GenerationResult:
         """Generate tokens for a batch of Requests.
 
         Each Request provides prompt_tokens, decode_params, and n_generations (clones).
-        Returns (outputs_per_sequence, total_generated_tokens).
         """
+        # validate we don't have any sequences with n_generations exceeding max_seqs
+        total_needed = sum(int(r.n_generations) for r in requests)
```
dlwh: in theory this should be fine so long as no r has too many

src/levanter/inference/openai.py
```diff
@@ -0,0 +1,703 @@
...
+    stop: Optional[Union[List[str], str]] = None
+    n: int = 1
+    seed: Optional[int] = None
+    logprobs: bool = False
+
+
+class LogProb(BaseModel):
+    token: str
+    logprob: float
+    bytes: Optional[List[int]] = None
```
dlwh, line 83: this is nonstandard yeah?

src/levanter/inference/openai.py
```diff
@@ -0,0 +1,703 @@
...
+                        break
+
+                with self.model_lock:
+                    self._process_batch(requests)
+            except Exception as e:
+                logger.error(f"Error in inference loop: {e}", exc_info=True)
+
+        logger.info("Inference thread shutting down")
+
+    def _process_batch(self, requests: List[InferenceRequest]):
```
dlwh, line 280: probably really GenerationService/InferenceEngine should just be constantly looping (maybe in this thread) and have queue it polls. I was working on getting there but you are likely to be faster

dlwh: (i'll try to do this since i'm mucking around in this bit, actually)

src/levanter/inference/openai.py
rjpower: This is how it shows up from openai?

https://gist.github.com/rjpower/ec43bd6f95464c7cbb9abe6dc2e4e4ab

src/levanter/inference/engine.py
rjpower: as in we don't need the test or we should be testing for max instead of sum?

just trying to avoid the infinite loop

src/levanter/inference/openai.py
rjpower: i'm fine either way, if it's more performant to have the engine poll the queue that's fine, we'll just want to be careful about the request matching and error tracking.

src/levanter/inference/engine.py
rjpower: changed to max, we'll tweak as needed.

src/levanter/inference/openai.py
rjpower: (well technically it's buried in a contents but i'll add parity checking for the structure next)

dlwh: TIL!

src/levanter/inference/openai.py
dlwh: yeah i think i'll have `submit_request` return a handle etc instead of what we currently do

PR #1161, opened 2025-09-15
Misc improvements to Engine

dlwh:


.agents/projects/inference.md
```diff
@@ -0,0 +1,215 @@
+# OAI-compatible Inference Server (Levanter)
```
dlwh, line 1: nb this is out of date but kind of useful

src/levanter/callbacks/__init__.py
```diff
@@ -151,11 +153,112 @@ def flush_stdout():
...
+        if event is not None:
+            event.set()
+
+        levanter.tracker.current_tracker().log_artifact(path, type="jax_profile")
+        # Log host stats if available
+        if stats_path is not None and os.path.exists(stats_path):
+            try:
+                levanter.tracker.current_tracker().log_artifact(stats_path, type="host_profile")
+            except Exception:
+                pass
```
rjpower: nit logger.warn("Failed to log host profile stats", exc_info=True)

src/levanter/inference/jit_scheduler.py
```diff
@@ -292,35 +319,62 @@ def update_tokens(
...
             def update(state):
-                tkns, lps, cnts, pids = state
+                tkns, lps, cnts, pids, f, should_purge = state
                 pos = cnts["seq", sid].scalar()
                 tkns = tkns.at["seq", sid, "position", pos].set(new_tokens["position", i])
                 if lps is not None:
                     lps = lps.at["seq", sid, "position", pos].set(new_log_probs["position", i])
                 cnts = cnts.at["seq", sid].add(1)
+                # completion checks
+                max_allowed = self.max_num_tokens["seq", sid].scalar()
```
rjpower, line 339: I might be missing something but: do these per sequence checks hurt us at all vs like setting the group max = max([seq in group]) and using a more fixed iteration? 

I guess this is the outer loop-ish so maybe not that critical. I was just wondering if this makes it harder for XLA to batch matmuls etc.

src/levanter/inference/service.py
```diff
@@ -36,9 +38,23 @@ class Request:
...
+    tokens_decoded: int = 0
+    done: bool = False
+    # Optional per-token logprobs (not populated yet)
+    logprobs: list[float] | None = None
+
+
 @dataclass(frozen=True)
-class GenerationServiceConfig:
-    """Configuration for GenerationService memory/layout knobs.
+class EngineConfig:
```
rjpower: Is this common terminology? Seems generic vs InferenceEngine, InferenceManager etc.

rjpower (approved):

src/levanter/inference/service.py
dlwh: It's pretty common. Vllm uses LlmEngine, but sglang is just Engine https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/entrypoints/engine.py#L92 I dunno. I can do InferenceEngine

src/levanter/inference/jit_scheduler.py
dlwh: yeah i figured it's outer loop and doesn't matter. i can probably fully vectorize if you really want

rjpower: Eh, it's fine if it's running quickly, just curious if it would make it hard for XLA to determine the static behavior. But if it's running quickly clearly that's not a problem!

src/levanter/inference/service.py
rjpower: Up to you, it seemed a bit generic to me in the context of Levanter, but not a big deal.

dlwh: I went with InferenceEngine

PR #1169, opened 2025-09-17
Toward managing sequence slot allocations in the host, bug fix for out of order slot ids

dlwh:
I'm trying to reduce number of round trips per decode step to make things faster. One way to do this is to batch up seq allocations and other metadata management into a single transaction. This means maintaining more state host-side. (Where possible, I'd like to do less state management device side.) Right now, that materializes as us maintaining a free list of "slots" host side as well as the mapping `slot <-> (request_id, child)`

In the process, I found a bug where the `allocate_for_seq` logic was very wrong for the case where the seq is were not densely packed. This includes a fix for that.


src/levanter/inference/page_table.py
```diff
@@ -102,24 +140,69 @@ def no_op(self_):
...
+
+        **ASSUMES** that the ``token_slot_ids`` are already grouped by sequence ID, i.e. all tokens for a given sequence
+        are contiguous in the input. The order of sequences in the input does not matter.
+        """
+
+        token_slot_ids = hax.where(token_slot_ids < 0, self.max_seqs, token_slot_ids)
+        # CAREFUL: we don't assume slot_ids are sorted, just contiguous. segment_sum is our friend
+        # NB: segment_sum assumes that segment ids are in the range [0, num_segments)
+        # and returns an array of such length
+        # essentially this means segment_sum et al require sorted, dense segment ids,
```
rjpower: nit, doesn't assume sorted, just dense?

src/levanter/inference/page_table.py
```diff
@@ -430,8 +514,11 @@ class PageBatchInfo(eqx.Module):
     page_indices[0] does not in general correspond to the first sequence in DecodeState, but rather the first sequence
     that has tokens **in this batch**.
 
+    To recover the mapping, using slot_ids
```
rjpower: Feels like something is missing...

rjpower (approved):
Reads easier as well this way.

PR #1170, opened 2025-09-17
Adopt OpenAI types in inference server, batching improvements.

rjpower:
The OpenAI response format is a bit twisty, trying to write our own Pydantic models is destined to result in small inconsistencies.

Fortunately we can import these types directly from the OpenAI Python API.
Unfortunately, they don't define Pydantic models for the request side, so we make do with TypedDict.


dlwh (approved):

dlwh:
test is failing because i try to interpret all config files that start with llama as a train_lm with llama config...

PR #1172, opened 2025-09-19
Reduce roundtrips from prefill

dlwh:
Pushes sequence allocation into the kernel

### DIFF
--- a/src/levanter/inference/engine.py
+++ b/src/levanter/inference/engine.py
@@ -852,7 +852,6 @@ class InferenceEngine:
 
         # _run_prefill returns (GenState, _DecodeOutputs)
         self.gen_state, outputs = new_state
-        # self.gen_state = jax.block_until_ready(self.gen_state)
         return outputs
 
     def _free_page_count(self) -> int:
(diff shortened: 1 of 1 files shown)


src/levanter/inference/engine.py
```diff
@@ -778,31 +866,52 @@ def _pages_needed_for_prompt(self, prompt_len: int) -> int:
...
         max_prefill_size = self.config.max_prefill_size or self.model.Pos.size
-        tokens = np.full((max_prefill_size,), INVALID, dtype=jnp.int32)
-        slot_ids = np.full((max_prefill_size,), INVALID, dtype=jnp.int32)
-        pos_ids = np.full((max_prefill_size,), INVALID, dtype=jnp.int32)
-        offset = 0
-        num_seqs_in_prefill = 0
+        max_seq_len = decode_state.tokens.axis_size("position")
+        max_slots = decode_state.max_seqs
+
+        queue_tokens = np.full((max_prefill_size,), INVALID, dtype=jnp.int32)
```
rjpower, line 878: nit the `np` api with `jnp.int32` is a little odd looking, maybe a convention i'm not aware of though

rjpower (approved):

src/levanter/inference/engine.py
dlwh: oh yeah. will fix

PR #1175, opened 2025-09-21
force 0.7.2 so we can use the fancier paged attention

dlwh:


dlwh:
(apparently broken. need to fix)

rjpower (approved):

rjpower:
Nice, looking forward to trying it out!

PR #1184, opened 2025-09-24
Cleanups for inference.

rjpower:
* Adjust VM counting logic to be consistent with TPU topologies.
* Add more logging to model loading.
* Handle new gemma model configuration: the text model is hidden under text_config now.
* Only write tensorboard records from worker 0 and handle string records appropriately.

### DIFF
--- a/src/levanter/inference/openai.py
+++ b/src/levanter/inference/openai.py
@@ -42,7 +42,6 @@ from levanter.checkpoint import load_checkpoint
 from levanter.compat.hf_checkpoints import HFCheckpointConverter, RepoRef, load_tokenizer
 from levanter.inference.engine import InferenceEngine, InferenceEngineConfig, Request
 from levanter.inference.jit_scheduler import SeqDecodingParams
-from levanter.models.llama import LlamaConfig, LlamaLMHeadModel
 from levanter.models.lm_model import LmConfig, LmHeadModel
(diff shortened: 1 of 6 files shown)


AlienKevin:
@rjpower I'm still getting up to speed with the inference code. How much of the OpenAI api do we support? and is there a way to test the oai api?

rjpower:
We support the basic stuff:

* n (number of generations)
* temperature
* stop tokens
* top-1 logprobs 

Things we don't support:

* model argument (it's ignored)
* streaming
* logit_bias
* frequency_penalty

I just cleaned up the inference repl and added a serving mode to this PR, you can run it with:

uv run --extra=serve src/levanter/main/inference_repl.py --command=serve --checkpoint=timinar/baby-llama-58m --model.type=llama

And then you can use any particular client you want or even curl:

curl http://localhost:8000/v1/chat/completions -H "Content-Type: application/json" -H "Authorization: Bearer $OPENAI_API_KEY" -d '{     "model": "meta-llama/Llama-3.2-1B-Instruct",
    "messages": [{"role": "user", "content": "How many words in the next sentence? \"I like cats!\""}],
    "logprobs": false,
    "max_tokens": 128,
    "temperature": 0.1
  }'

rjpower:
All tests passed manually.

PR #1187, opened 2025-09-25
Support generation-based evals in lm_eval_harness

AlienKevin:
1. Stop sequence is working now (a padding bug has been fixed).
2. Added support for passing apply_chat_template and fewshot_as_multiturn in the eval_harness field of the YAML config. These two fields are required by Llama 3 during their GSM-8K evals.
3. Reproduced Llama 3.1 8B Instruct's GSM-8K (CoT) result using similar lm_eval_harness setup. (reproduced 85.6 vs reported 84.5)
4. Prompts and generation outputs are now saved to lm_eval_harness_results.json for easy inspection.

You can try out the updated lm_eval_harness:
```
uv run --extra tpu python -m levanter.eval_harness --config config/harness/harness_llama3_instruct.yaml
```

You need to additionally install `lm_eval`:
```
uv pip install lm_eval
```

v5p-8 run: https://wandb.ai/kevinxli/levanter/runs/t1j3nmxx
v4-8 run: https://wandb.ai/kevinxli/levanter/runs/80mvwdk9


src/levanter/eval_harness.py
```diff
@@ -101,7 +107,7 @@ class _LmEvalHarnessWorker:
     """
 
     def __init__(
-        self, EvalBatch, EvalPos, model, axis_resources, tokenizer, mp, max_packed_segments, apply_chat_template=False
+        self, EvalBatch, EvalPos, model, axis_resources, tokenizer, mp, max_packed_segments, max_gen_toks=256,
```
BabyChouSr: could we abstract out `max_gen_toks` to `generation_kwargs: dict`? Then, a user can pass in the temperature, max_gen_toks, top_p, top_k via this dictionary more simply.

src/levanter/eval_harness.py
```diff
@@ -382,10 +733,12 @@ def to_dict(self):
 class LmEvalHarnessConfig:
     task_spec: list[TaskConfig | str]
     max_examples: int | None = None
-    max_eval_length: int | None = None
+    max_length: int | None = None
     log_samples: bool = False
     bootstrap_iters: int = 0
     apply_chat_template: bool = False
+    fewshot_as_multiturn: bool = False
+    max_gen_toks: int = 256
```
BabyChouSr: Similar to the comment above, can we abstract out max_gen_toks as a kwarg in `generation_kwargs: dict`.

config/harness/harness_nano.yaml
```diff
@@ -24,6 +26,8 @@ trainer:
   per_device_parallelism: -1
   train_batch_size: 32
 
-  tensor_parallel_axes: ["mlp", "heads"]
+  tensor_parallel_axes: ["mlp", "heads", "kv_head", "vocab"]
+  model_axis_size: 4
```
dlwh, line 30: nano is meant to run on a laptop but i think this is fine

src/levanter/eval_harness.py
```diff
@@ -324,11 +419,267 @@ def loglikelihood(self, requests: list[Instance]) -> list[tuple[float, bool]]:
 
         return result
 
+
+    def tok_encode(
```
dlwh, line 438: make protected?

```suggestion
    def _tok_encode(
```

src/levanter/eval_harness.py
```diff
@@ -324,11 +419,267 @@ def loglikelihood(self, requests: list[Instance]) -> list[tuple[float, bool]]:
...
+        self,
+        string: Union[str, List[str]],
+        left_truncate_len: int = None,
+        add_special_tokens: bool = False,
+        truncation: bool = False,
+    ) -> Union[List[int], List[List[int]]]:
```
dlwh, line 444: quick doc comment?

src/levanter/eval_harness.py
```diff
@@ -324,11 +419,267 @@ def loglikelihood(self, requests: list[Instance]) -> list[tuple[float, bool]]:
...
+            kwargs["max_gen_toks"] = int(kwargs["max_gen_toks"])
+            
+        # Handle seed parameter  
+        if "seed" in kwargs and kwargs["seed"] is not None:
+            kwargs["seed"] = int(kwargs["seed"])
+        
+        # Handle n generations parameter
+        if "n" in kwargs:
+            kwargs["n"] = int(kwargs["n"])
+
```
dlwh, line 727: should we warn on args we don't understand yet?

src/levanter/eval_harness.py
```diff
@@ -324,11 +419,267 @@ def loglikelihood(self, requests: list[Instance]) -> list[tuple[float, bool]]:
...
+                "prompt": prompt_text,
+                "generation": text,
+            })
+        
+        # print(f'{outputs=}')
+
+        return outputs
+
+    @staticmethod
+    def modify_gen_kwargs(kwargs: dict) -> dict:
```
dlwh:
```suggestion
    def _modify_gen_kwargs(kwargs: dict) -> dict:
```

src/levanter/eval_harness.py
```diff
@@ -324,11 +419,267 @@ def loglikelihood(self, requests: list[Instance]) -> list[tuple[float, bool]]:
...
+            stop_tokens = gen_kwargs.get("stop_tokens")
+            if stop_tokens is not None:
+                # stop_tokens is a NamedArray with shape (stop_seq, position)
+                num_stop_seqs = stop_tokens.shape['stop_seq']
+                num_stop_tokens = stop_tokens.shape['position']
+                max_stop_seqs = max(max_stop_seqs, num_stop_seqs)
+                max_stop_tokens = max(max_stop_tokens, num_stop_tokens)
+        
+        # Taken from: config/sampler/sample_llama8b.yaml
+        engine_cfg = InferenceEngineConfig(
```
dlwh, line 605: we should probably make this parametric at some point

dlwh: also, for now, we should probably error out on multihost jax because I don't think Engine works there yet...

src/levanter/eval_harness.py
```diff
@@ -324,11 +419,267 @@ def loglikelihood(self, requests: list[Instance]) -> list[tuple[float, bool]]:
...
+    def modify_gen_kwargs(kwargs: dict) -> dict:
+        """
+        Modify generation kwargs to standardize parameters, similar to vLLM implementation.
+        """
+        # Handle temperature
+        kwargs["temperature"] = kwargs.get("temperature", 0.0)
+        
+        # Handle do_sample parameter like vLLM does
+        do_sample = kwargs.pop("do_sample", None)
+        if do_sample is False and "temperature" not in kwargs:
```
dlwh: if do_sample is FAlse and there's a non-zero temperature, I think we should maybe error? or at least warn?

src/levanter/eval_harness.py
AlienKevin: n_generations is supported by src/levanter/inference/engine.py

src/levanter/eval_harness.py
AlienKevin: tok_encode is an abstract method from the parent TemplateLM so it doesn't need to be prefixed with `_`

src/levanter/eval_harness.py
AlienKevin: Added

src/levanter/eval_harness.py
AlienKevin: Created a unified generation_kwargs for temperature, max_gen_toks, max_length etc

src/levanter/eval_harness.py
```diff
@@ -382,10 +783,44 @@ def to_dict(self):
...
+        return self.generation_kwargs.get("max_gen_toks", 256)
+    
+    def get_effective_max_length(self) -> int | None:
+        """Get the effective max_length, preferring generation_kwargs over the top-level field."""
+        # Prefer max_length from generation_kwargs if set
+        gen_max_length = self.generation_kwargs.get("max_length")
+        if gen_max_length is not None:
+            return gen_max_length
+        # Fall back to top-level max_length field for backward compatibility
+        return self.max_length
```
BabyChouSr: Maybe we should just have `max_length` in one place in the config and not have it in the generation_kwargs.

src/levanter/eval_harness.py
```diff
@@ -602,18 +1038,37 @@ def _actually_run_eval_harness(
...
+        sample_outputs = harness.get_sample_outputs()
+        if sample_outputs:
+            # Add outputs to each benchmark in results
+            for task_name in outputs.get("results", {}):
+                # Get all sample outputs for this task (since we don't track individual tasks yet)
+                all_samples = []
+                for samples in sample_outputs.values():
+                    all_samples.extend(samples)
+                if all_samples:
+                    outputs["results"][task_name]["outputs"] = all_samples
```
BabyChouSr, line 1062: could you add something that logs the prompts, generations, and the solutions to wandb via `log_report_to_tracker`? This will make it easier to inspect the results of the model's responses.

src/levanter/eval_harness.py
```diff
@@ -52,16 +53,21 @@
...
+from levanter.inference.engine import InferenceEngine, InferenceEngineConfig, Request as GenRequest
+from levanter.inference.jit_scheduler import SeqDecodingParams
+from levanter.inference.utils import INVALID
 
 try:
     from lm_eval import evaluator
     from lm_eval.api.instance import Instance
-    from lm_eval.api.model import LM
+    from lm_eval.api.model import TemplateLM
+    from lm_eval.models.utils import handle_stop_sequences, postprocess_generated_text
```
BabyChouSr, line 64: one note about the function `postprocess_generated_text` is that it is not available in our current lm eval harness fork. we will need to rebase off main which i have PR'd but currently testing. Our fork being https://github.com/stanford-crfm
[lm-evaluation-harness](https://github.com/stanford-crfm/lm-evaluation-harness)

BabyChouSr (commented):
amazing work so far! it would be nice to be able to add the wandb upload of the prompts / generations / results for easier visual inspection.

src/levanter/eval_harness.py
dlwh: i meant that there are other ones we don't support like top_p etc

src/levanter/eval_harness.py
```diff
@@ -324,11 +435,301 @@ def loglikelihood(self, requests: list[Instance]) -> list[tuple[float, bool]]:
...
+                            stop_tokens_array, 
+                            ("stop_seq", "position")
+                        )
+                        gen_kwargs["stop_tokens"] = stop_tokens
+                    else:
+                        gen_kwargs["stop_tokens"] = None
+                else:
+                    gen_kwargs["stop_tokens"] = None
+            else:
+                gen_kwargs["stop_tokens"] = None
```
dlwh: maybe a helper for this?

dlwh (approved):
nice!

src/levanter/eval_harness.py
AlienKevin: The prompts and generations are logged implicitly and uploaded to Wandb. See example: https://wandb.ai/kevinxli/levanter/runs/t1j3nmxx/logs

src/levanter/eval_harness.py
AlienKevin: That's a good point. Tho because max_length is needed by both generation and loglikelihood evals, I can leave it at the top level instead of inside generation_kwargs.

src/levanter/eval_harness.py
AlienKevin: Good idea, done

AlienKevin:
@dlwh @BabyChouSr updated

config/harness/eval_llama3.yaml
```diff
@@ -45,7 +45,7 @@ trainer:
   per_device_parallelism: -1
   train_batch_size: 512
 
-  tensor_parallel_axes: ["mlp", "heads"]
+  tensor_parallel_axes: ["mlp", "kv_head"]
```
dlwh: actually "vocab" should probably be a TP axis

dlwh (approved):

PR #1202, opened 2025-09-26
Integrate size threshold for chips to vm on V4 topologies

Helw150:


src/levanter/infra/ray_tpu.py
```diff
@@ -135,7 +135,7 @@ def _parse_tpu_type(tpe: str) -> tuple[str, int]:
     except ValueError:
         raise ValueError(f"Invalid chip count in TPU type: {tpe}")
 
-    return family, chip_count
+    return family, chip_count // 2
```
rjpower: Wow, thank you Google for the consistent naming.

I think this is only for v4, v4-2048 = 1024 chips = 256 hosts.

v5/v6 name = #chips so: v5litepod-64 = 64 chips = 16 hosts

if family == "v4": return family, chip_count // 2
return family, chip_count

?

Helw150: I find this all so confusing docs wise that if you left this to me I would literally just try to spin up every pod type and look at how much of the resource we get on ray and then save the mapping...

so I will do whatever you say as long as the result is 256 for a v4-2048 🫡

rjpower: I guess v5p is the same as v4, sigh. But yeah v5e is name = chips. I'll look into if we can build a table...

PR #1204, opened 2025-09-26
Use an explicit table to track TPU configurations.

rjpower:


dlwh (approved):
thanks! i think it's best we just make this as dumb as possible as dumb as it feels

Issue #1181, opened 2025-09-23
Ragged paged attention kernel results in weird inference results.


rjpower:
I might be doing something wrong but:

```
uv run scripts/ray/dev_tpu.py --tpu-name=power-scratch execute\
  -- "cd submodules/levanter && uv run --extra=serve --extra=tpu src/levanter/main/inference_repl.py --command=complete --args='What is the square root of 17?' --max_tokens=32 --checkpoint=Qwen/Qwen3-4B-Instruct-2507 --model.type=qwen3 2>&1 | grep -v 'Could not open'"
```

Syncing local changes to dev-tpu-power-scratch...
2025-09-23 16:44:56,629 - INFO - Syncing 1334 files (git-tracked, modified, and important untracked)...
2025-09-23 16:45:01,452 - INFO - File sync completed successfully
Running: cd submodules/levanter && uv run --extra=serve --extra=tpu src/levanter/main/inference_repl.py --command=complete --args='What is the square root of 17?' --max_tokens=32 --checkpoint=Qwen/Qwen3-4B-Instruct-2507 --model.type=qwen3 2>&1 | grep -v 'Could not open'
Loading Qwen/Qwen3-4B-Instruct-2507...
2025-09-23 23:45:11,680 - INFO - Loaded tokenizer with vocab size 151669 from Qwen/Qwen3-4B-Instruct-2507
Loading weights: 100%|██████████| 219/219 [00:02<00:00, 91.70it/s] 
Loading weights: 100%|██████████| 174/174 [00:02<00:00, 73.10it/s]
Loading weights: 100%|██████████| 5/5 [00:00<00:00, 100.49it/s]
2025-09-23 23:45:19,511 - INFO - Resizing model from 151936 to 151669 to match tokenizer vocab size
2025-09-23 23:45:25,661 - INFO - Starting inference context...
2025-09-23 23:45:25,662 - INFO - Inference thread started
2025-09-23 23:45:25,662 - INFO - Batch processing thread started
2025-09-23 23:45:25,662 - INFO - Inference service initialized and ready
✓ Loaded Qwen/Qwen3-4B-Instruct-2507
[...]

PR #1208, opened 2025-09-29
Reenable TPU ragged paged attention

AlienKevin:
We can reenable RPA since shape error bug (https://github.com/marin-community/levanter/issues/1176) is already fixed by https://github.com/marin-community/levanter/pull/1192


dlwh:
i think the bigger reason we left it disabled was #1181 but i think that's a different problem

dlwh:
tests are failing... let me see if i have a branch that fixes

PR #1212, opened 2025-09-29
Move model loading out of the `openai` server.

rjpower:
This moves model loading into the user of the server, e.g. the inference repl, rollout worker, or future Marin inference serving implementation. The loading was only used by the inference repl side of things and can be better handled by the user.

### DIFF
--- a/src/levanter/inference/openai.py
+++ b/src/levanter/inference/openai.py
@@ -20,15 +20,12 @@ import uuid
 from dataclasses import dataclass, field
 from typing import Dict, List, Literal, Optional, Union
 
-import equinox as eqx
 import haliax as hax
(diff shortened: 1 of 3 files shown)


rjpower:
I fixed a few pre-commit issues in the lm_eval code, not sure how those snuck in or if we have a pre-commit versioning issue.

rjpower:
For example, if I wanted to evaluate some examples while I was training a model, it seems natural to be able to write:

```
engine = InferenceEngine(model=self.trained_model)
run_inference_samples(engine)
```

Instead of committing a checkpoint first and then reloading it.

AlienKevin:
I see, makes sense!

PR #1219, opened 2025-10-01
Add support for custom eval plugins

eric-czech:
refs: https://github.com/Open-Athena/marin/pull/2

This adds a very simple plugin model for registering custom evals during training.

### DIFF
--- a/docs/Getting-Started-Training.md
+++ b/docs/Getting-Started-Training.md
@@ -185,6 +185,7 @@ Levanter supports custom evaluation plugins that run during training to track sp
 # my_eval_plugin.py
 import jax.numpy as jnp
 import levanter
+from levanter.callbacks import StepInfo
 from levanter.utils.hf_utils import HfTokenizer
 from levanter.eval import EvalPlugin
 
@@ -193,10 +194,10 @@ class PerplexityPlugin(EvalPlugin):
         # Plugin handles its own configuration internally
         self.sample_text = "The quick brown fox jumps over the lazy dog."
 
-    def create_callback(self, *, tokenizer: HfTokenizer, **kwargs):
+    def create_callback(self, *, tokenizer: HfTokenizer, **kwargs) -> Callable[[StepInfo], None]:
         tokens = tokenizer(self.sample_text)["input_ids"]
 
-        def callback(step_info):
+        def callback(step_info: StepInfo) -> None:
             # Simple perplexity calculation on sample text
             eval_model = step_info.eval_model
             logits = eval_model(jnp.array(tokens[:-1]))


docs/Getting-Started-Training.md
```diff
@@ -176,3 +176,44 @@ start training from scratch.
 2. Resume from a WandB run: if you set `--trainer.wandb.resume`, it will resume the corresponding WandB run with the ID `asdf1234`. You can
 find the WandB run ID in the URL of your WandB run page. For more information, please refer to the
 [WandB documentation](https://docs.wandb.ai/guides/runs/resuming).
+
+## Custom Evaluation Plugins
```
eric-czech, line 180: Is there a better place for an example like this?  I didn't see anything obviously better.

src/levanter/eval.py
```diff
@@ -479,3 +482,49 @@ def zeros_like(total: Arrayish, per_tag: Arrayish) -> "_EvalRunningMeans":
...
+
+
+class EvalPlugin(ABC):
+    """Abstract base class for evaluation plugins."""
+
+    @abstractmethod
+    def create_callback(
+        self,
+        *,
+        tokenizer: HfTokenizer,
```
eric-czech, line 504: Passing through the tokenizer and batch size made sense to me here.  I'm not so sure about the device_mesh and axis mappings -- presumably at least the device mesh should be used for doing inference in parallel batches?

Overall, I don't really know what context is best -- guidance is welcome.

src/levanter/eval.py
```diff
@@ -479,3 +482,49 @@ def zeros_like(total: Arrayish, per_tag: Arrayish) -> "_EvalRunningMeans":
...
+            parameter_axis_mapping: Axis mapping for parameter storage
+            batch_size: Evaluation batch size
+
+        Returns:
+            A callback function that takes StepInfo and returns None
+        """
+        ...
+
+
+def load_eval_plugin(config: EvalPluginConfig) -> EvalPlugin:
```
eric-czech, line 525: I thought a little bit about doing this through the draccus [PluginRegistry](https://github.com/marin-community/draccus/blob/13dd58f0c67002add8a188d85ba0e74661039078/draccus/choice_types.py#L129) originally, but I hit a few problems and it seemed like overkill.  That might be worth revisiting, or it could at least make sense to use the namespace package model for plugins (i.e. mentioned [here](https://github.com/marin-community/draccus/blob/13dd58f0c67002add8a188d85ba0e74661039078/draccus/choice_types.py#L133-L134)).  This seems simpler though, so let me know if I'm missing something important those would add.

tests/test_train_lm.py
```diff
@@ -23,29 +60,12 @@ def test_train_lm():
...
-                ),
-                trainer=train_lm.TrainerConfig(
-                    num_train_steps=2,
-                    train_batch_size=len(jax.devices()),
-                    max_eval_batches=1,
-                    tracker=NoopConfig(),
-                    require_accelerator=False,
-                    ray=RayConfig(auto_start_cluster=False),
-                ),
+                model=_create_test_model_config(),
```
eric-czech, line 64: There was enough duplication in this off-target test (and the one below) that some helper fxns seem warranted.  I have carefully checked them to ensure the same logic was preserved (i.e. it was not an AI refactor).

src/levanter/eval.py
rjpower: Having recently written this rant, I can sympathize: https://github.com/marin-community/levanter/issues/1213

FWIW, this seems easier to follow than a registry, since you would clearly observe the fully-qualified classname in the configuration and wouldn't guess at what happens.

I have a bit of a knee-jerk reaction to "programming through config" objects, but I'm willing to accept that this is a matter of taste and I'm just yelling at clouds.

That is, I have nothing bad to say about this PR, which seems reasonable to be in this context, but I tend to be rather more comfortable with, for instance, asking the user to register their plugins in code explicitly rather than via configuration, but again, yelling at clouds.

eric_train_lm.py:
   trainer = Trainer(..., on_eval=[ComputePerplexity()])

eric-czech: > but again, yelling at clouds

Heh not really, I hear you.  For this example:

```
trainer = Trainer(..., on_eval=[ComputePerplexity()])
```

Do you have any ideas on how one could do such a thing if all users are supposed to be separated from direct `Trainer` instantiation by `TrainLmConfig` though?  I like that seemingly all important infrastructure functionality within marin/levanter is gated by some kind of easily serializable input.  I'm biased against [cloudpickle](https://github.com/cloudpipe/cloudpickle)-y distributed systems that try to make everything serializable, e.g. code callbacks, but only because it has caused me so many headaches in the past.

Put differently, what's a third option beyond programming through config and programming with a bunch of dynamic code references flying around?  Maybe all evals should just have to be defined within Levanter, given some name, and then referenced by name in a config?

rjpower: For sure, cloudpickle can go wrong quickly. There's definitely something nice about having things be expressed outside of code. That said, I guess I don't anything too wrong with the idea of having a (small) separate main for your experiment:

```
eric_repo/
  eric_trainer.py
```

```
# eric_trainer.py
def main():
   trainer = Trainer(
     # whatever configuration you want here, you could load from config object if you want
     # or just set things directly in code
     eval_plugins=[PerplexityMetric(...)])
```

You don't need cloud pickle in this case, you would instead run `eric_trainer.py` on all of your workers instead of `train_lm.py`, and code in whatever plugins or hooks you want.

That is, as a user you could choose to make your trainer as configurable as you want, but by having a separate main you remove the need for reflection on the metrics (since the user just does it themself). I acknowledge this would be a detour from how we're doing things now! There's certainly a tradeoff for e.g. doing things like parameter sweeps, where you still need to express that somehow.

eric-czech: Makes sense.  I could use `Trainer` directly -- pretty much all the torch frameworks I know of expose a `Trainer` so it would certainly be familiar.  Some have a layer on top of that, and that's how I think of Marin.

A relevant question then is, should Marin users be able to run their own evals via the Marin interface rather than the Levanter interface?  I don't see a single experiment that uses the Levanter `Trainer` directly now, and IMO custom evals are a thing worth supporting at the Marin level.  Would you not agree?

rjpower: That makes sense to me? David probably has more, and more sensible thoughts on it. But for instance, Calvin added the "hackable transformer" starter recently to Marin:

https://github.com/marin-community/marin/pull/1638

Before we had Marin, there wasn't really a sensible place to put things like this, but maybe now (and esp. with the mono-repo in mind), you could imagine parts of Levanter serving as templates rather than having to do everything.

eric-czech: > Calvin added the "hackable transformer" starter recently to Marin
> you could imagine parts of Levanter serving as templates

Nice!  Relying more and more on templates certainly tracks for me.  Managing the boilerplate associated with that approach has become much more manageable with AI copilots IMO.

docs/Getting-Started-Training.md
```diff
@@ -176,3 +176,45 @@ start training from scratch.
...
+
+class PerplexityPlugin(EvalPlugin):
+    def __init__(self):
+        # Plugin handles its own configuration internally
+        self.sample_text = "The quick brown fox jumps over the lazy dog."
+
+    def create_callback(self, *, tokenizer: HfTokenizer, **kwargs) -> Callable[[StepInfo], None]:
+        tokens = tokenizer(self.sample_text)["input_ids"]
+
+        def callback(step_info: StepInfo) -> None:
```
eric-czech, line 200: How should these callbacks be written to be aware of cluster topology (i.e. use multiple devices) @dlwh?

rjpower: You're typically still running in a TPU mesh context at this point, so you're able to do multi-host operations on the TPU mesh. At this point, however, the plugin has to be responsible for getting the data into the right format and synchronizing across the workers.

For host synchronization you can check out `jax_utils.py` for some helpful functions like multihost_broadcast_sync and barrier_sync.

If you wanted to evaluate on let's say an input batch, you can use something like:

```
sharded_batch = hax.shard(batch, trainer.compute_axis_mapping)
```

If you're using raw Jax arrays for your model, you'll have to manually construct it with the appropriate Sharding to align with the model:

https://docs.jax.dev/en/latest/jax.sharding.html#jax.sharding.Sharding

src/levanter/eval.py
dlwh: yeah that's the impetus behind hackable. I think there's still some value b/c it's easier to do apples to apples comparisons with config flags than two mostly identical but not entirely implementations of things.

We do need to support more kinds of callbacks and I'm happy to say no to pluginregistries and:

1) In Marin, prefer just passing in objects since I think pickle will be ok here?
2) In yaml, prefer fully qualified class names.

docs/Getting-Started-Training.md
```diff
@@ -176,3 +176,45 @@ start training from scratch.
...
+
+```python
+# my_eval_plugin.py
+import jax.numpy as jnp
+import levanter
+from levanter.callbacks import StepInfo
+from levanter.utils.hf_utils import HfTokenizer
+from levanter.eval import EvalPlugin
+
+class PerplexityPlugin(EvalPlugin):
```
dlwh, line 192: really we should just have full callback plugins but this is a good starting point. evals are a bit special

docs/Getting-Started-Training.md
dlwh: Russell basically covered it, but there is a bit of subtlety if you want to do all data loading/processing on one host but do compute on all hosts. (We do this in eval_harness.py, but it is a mess.)

src/levanter/eval.py
dlwh: yeah i think you want all these things. The Mesh is usually available via a global function at this point but everything else is pretty

What do you think the best way to deal with this kind of thing is? For, e.g., our audio models (which exist now!) we wouldn't want to just pass in a tokenizer, but a Processor... 

I've always struggled with how to expose this kind of stuff

src/levanter/eval.py
```diff
@@ -479,3 +482,49 @@ def zeros_like(total: Arrayish, per_tag: Arrayish) -> "_EvalRunningMeans":
         z = RunningMean.zeros_like(total)
         per_tag = RunningMean.zeros_like(per_tag)
         return _EvalRunningMeans(z, per_tag, z, per_tag)
+
+
+@dataclass
+class EvalPluginConfig:
```
dlwh, line 488: i guess i'd prefer:

- For Marin, supporting passing through the plugin object directly. I don't think we need to insist on reflection when it's not necessary
- For the yaml/draccus path, extending draccus to support fully qualified names (which is clearly better and I am a fool)

docs/Getting-Started-Training.md
dlwh: i think a new doc would be good?, probably with a reference in this one and maybe ConfigurationGuide.md

src/levanter/eval.py
jder: Just my 2 cents and without much context, but I agree with the push towards either defining everything in levanter or having eric_trainer.py, where the latter loads up a normal train config and maybe its own separate config and creates/configures/joins these things together in code eg `trainer = trainer_config.build(); trainer.add_callback(my_special_config.build(trainer.tokenizer, …))`

docs/Getting-Started-Training.md
eric-czech: To your knowledge, is it possible to get an incorrect result on model inference when not handling the sharding correctly (e.g. within a TPU mesh context as @rjpower mentioned)?  I can't tell based on the jax/haliax docs if it's possible to get an incorrect result from a model, rather than a correct result computed inefficiently.  

E.g. is it possible that activations/logits get computed from only partial weights?

src/levanter/eval.py
eric-czech: > we wouldn't want to just pass in a tokenizer, but a Processor...

Ahh, makes sense.  I suppose a somewhat obvious approach would be:

1. Support custom callbacks first like you mentioned in https://github.com/marin-community/levanter/pull/1219#discussion_r2438552104
2. Extend the above for evals as callbacks
3. Make modality-specific, eval callback implementations

Nudging people towards using Levanter directly would probably be better IMO.

src/levanter/eval.py
eric-czech: > just passing in objects since I think pickle will be ok here?

Makes sense to me, modulo the usual footguns with allowing code as arguments like accidentally putting things in closures for callbacks that are super slow to pickle/unpickle.

>  you would instead run eric_trainer.py

I'm leaning that way instead, especially given that it probably would make sense to first abstract custom callbacks, and then custom evals, and then maybe modality-specific evals (as @dlwh pointed out).  It feels like that would all be pushing towards API surface area that would be less likely to be maintained well.

PR #1230, opened 2025-10-03
Support iteration-based profiling in InferenceEngine and eval_harness

AlienKevin:
## Problem

The `eval_harness.py` script uses `TrainerConfig` but doesn't create a `Trainer` instance, so profiler callbacks were never set up even when `--trainer.profiler` was enabled. Additionally, the `--trainer.profiler_start_step` and `--trainer.profiler_num_steps` parameters had no effect since eval_harness doesn't have a training loop with discrete steps.

## Solution

This PR implements decode iteration-based profiling for eval_harness by adding a step callback mechanism to `InferenceEngine`:

### InferenceEngine changes (`src/levanter/inference/engine.py`):
- Add `step_callback` parameter to `generate()` method
- Track `decode_iteration` counter in the generation loop
- Call `step_callback` at each decode iteration before processing

### eval_harness changes (`src/levanter/eval_harness.py`):
- Pass profiler configuration dict through to `LevanterHarnessLM` and `InferenceEngine`
- Implement `_handle_profiler_step()` to start/stop JAX profiler at specific decode iterations
- Add `decode_step_callback` that maps iteration numbers to profiler steps
- Set logger level to INFO to ensure profiler messages are visible
- Support step-based profiling for both `generate_until` and `loglikelihood` tasks

## Usage

Users can now profile specific decode iterations during evaluation:

```bash
uv run --extra profiling python -m levanter.eval_harness \
  --config config/harness/harness_llama3_instruct.yaml \
  --trainer.profiler true \
  --trainer.profiler_perfetto_link false \
  --trainer.profiler_start_step 67 \
  --trainer.profiler_num_steps 5 \
  --eval_harness.max_examples 200
```

This will:
- Profile decode iterations 67-72 (5 iterations starting at iteration 67, empirically these iterations have the highest throughput so we measure this region)
- Generate xprof traces in `logs/{run_id}/profiler/`
- Log "Starting profiler at step X" and "Stopping profiler at step Y" messages

With iteration-based profiling, we see clean iterations in the xprof trace:
Screenshot 2025-10-03 at 12 41 40 PM

## Testing

Tested with:
- GSM8K evaluation (\`generate_until\` task) with 200 examples
- Verified profiler starts/stops at correct iterations
- Confirmed valid xprof traces are generated (219MB xplane.pb for 20 iterations)
- Verified profiler messages appear in logs
- Tested with both profiler enabled and disabled modes

## Backwards Compatibility

- When `--trainer.profiler false` or omitted, no profiling overhead
- Existing eval_harness usage without profiler parameters continues to work unchanged


src/levanter/eval_harness.py
```diff
@@ -144,6 +144,7 @@ def __init__(
         max_packed_segments,
         generation_kwargs=None,
         sample_logging_config: SampleLoggingConfig | None = None,
+        profiler_config: Optional[dict] = None,
```
dlwh: i generally prefer data classes

src/levanter/eval_harness.py
```diff
@@ -393,6 +398,49 @@ def _prepare_bucket(self, task_name: str) -> list[dict[str, str]] | None:
...
+
+        # Stop profiler at end_step
+        elif self._current_step == end_step and self._profiler_started:
+            logger.info(f"Stopping profiler at step {self._current_step}")
+            jax.profiler.stop_trace()
+            self._profiler_started = False
+
+            # Log artifact
+            profile_path = self.profiler_config.get("profile_path", "/tmp/profile")
+            levanter.tracker.current_tracker().log_artifact(profile_path, type="jax_profile")
```
dlwh: can we merge this logic with _stop_porifler_if_needed to remove redundancy

src/levanter/eval_harness.py
```diff
@@ -393,6 +398,49 @@ def _prepare_bucket(self, task_name: str) -> list[dict[str, str]] | None:
...
+            self._profiler_started = True
+
+        # Stop profiler at end_step
+        elif self._current_step == end_step and self._profiler_started:
+            logger.info(f"Stopping profiler at step {self._current_step}")
+            jax.profiler.stop_trace()
+            self._profiler_started = False
+
+            # Log artifact
+            profile_path = self.profiler_config.get("profile_path", "/tmp/profile")
```
dlwh: default should maybe be /tmp/levanter_profile or something now that i think about it

src/levanter/eval_harness.py
```diff
@@ -685,7 +742,18 @@ def generate_until(self, requests) -> List[str]:
...
+            """Called at each decode iteration in the engine."""
+            # Use the iteration number as the step for profiling
+            saved_step = self._current_step
+            self._current_step = iteration
+            self._handle_profiler_step()
+            self._current_step = saved_step
+
+        # Pass the callback to the engine if profiling is enabled
+        step_callback = decode_step_callback if self.profiler_config.get("enabled", False) else None
+        result = engine.generate(gen_requests, step_callback=step_callback)
```
dlwh, line 772: don't love step_callback but it's clearlly convenient

PR #1231, opened 2025-10-03
MuP In Levanter

Helw150:
Chained with https://github.com/marin-community/haliax/pull/219


src/levanter/layers/attention.py
```diff
@@ -1474,15 +1477,31 @@ def init(config: AttentionConfig, *, key) -> "Attention":
...
+            reparam_cls=mup.OutputLinearMup if use_mup else mup.LinearStandardParam,  # Q out is not Width Dependent
         )
         k_proj = hnn.Linear.init(
-            In=config.Embed, Out=(config.KVHeads, config.HeadSize), key=k_k, use_bias=use_bias, out_first=True
+            In=config.Embed,
+            Out=(config.KVHeads, config.HeadSize),
+            key=k_k,
+            use_bias=use_bias,
+            out_first=True,
+            reparam_cls=mup.OutputLinearMup if use_mup else mup.LinearStandardParam,  # K out is not Width Dependent
```
dlwh: Wait, this is wrong right? either n_head or d_head will tend to infinity,

dlwh: (More generally, I kinda wonder if it makes more sense to do something like Haliax's axis mapping? Like, some axes are infinite, some are finite. Given some axes, you can classify any linear map as infinite or finite

src/levanter/models/llama.py
```diff
@@ -226,10 +228,12 @@ def attention_config(self) -> AttentionConfig:
             num_kv_heads=self.num_kv_heads,
             head_dim=self.head_dim,
             use_bias=self.use_bias,
+            use_mup=self.use_mup,
             upcast_attn=self.upcast_attn,
             attn_backend=self.attn_backend,
             flash_attention_block_size=self.flash_attention_block_size,
             rope=self.rope,
+            scaling_factor=1.0 if self.use_mup else None,  # Scaling is done by muP
```
dlwh: seems cleaner to change the default behavior in Attention so that None means mup if use_mup and 1/sqrt(d) otherwise

src/levanter/models/llama.py
```diff
@@ -259,11 +263,33 @@ def init(
...
-        gate_proj = hnn.Linear.init(Out=Mlp, In=Embed, key=k_fc, use_bias=use_bias, out_first=True)
-        up_proj = hnn.Linear.init(Out=Mlp, In=Embed, key=k_up_proj, use_bias=use_bias, out_first=True)
-        down_proj = hnn.Linear.init(Out=Embed, In=Mlp, key=k_down_proj, use_bias=use_bias, out_first=True)
+        gate_proj = hnn.Linear.init(
+            Out=Mlp,
+            In=Embed,
+            key=k_fc,
+            use_bias=use_bias,
+            out_first=True,
+            reparam_cls=mup.HiddenLinearMup if use_mup else mup.LinearStandardParam,
```
dlwh: kinda wonder if we should add aliases like

```
def hidden(use_mup=True): return mup.HiddenLinearMup if use_mup else mup.LinearStandardParam
```

and then do mup.hidden() but eh

src/levanter/main/train_lm.py
```diff
@@ -35,12 +35,37 @@
...
+    if requested is not None:
+        if requested and not model_supports:
+            raise ValueError("Model config does not support MuP but TrainLmConfig.use_mup=True was provided.")
+        if requested and not optimizer_supports:
+            raise ValueError("Optimizer config does not support MuP but TrainLmConfig.use_mup=True was provided.")
+
+        if model_supports and model_use_mup != requested:
+            config.model = dataclasses.replace(config.model, use_mup=requested)
+        if optimizer_supports and optimizer_use_mup != requested:
+            config.optimizer = dataclasses.replace(config.optimizer, use_mup=requested)
```
dlwh, line 56: i hate it but i guess it's ok

src/levanter/optim/mup.py
```diff
@@ -0,0 +1,73 @@
...
+            return leaf * lr_scale
+        return leaf
+
+    # Avoid descending into nested ReparamEnabled modules. We need to allow traversal
+    # of the root ``updates`` object, so we key off object identity instead of the type
+    # alone.
+    root_id = id(updates)
+
+    def _is_nested_reparam_enabled(node: Any) -> bool:
+        return isinstance(node, ReparamEnabled) and id(node) != root_id
```
dlwh: hrm, is it even well defined to have nested objects like this? if anything, I'd expect to switch to the LR preferred by the inner module?

dlwh (changes requested):
I'm pretty sure the attention is wrong? Willing to be convinced

src/levanter/layers/attention.py
Helw150: image

Helw150: image
and relevant definition from appendix

Helw150: If n_head tends to infinity, then the same argument of output layers applies to each linear map. If d_head tends to infinity the argument is mildly different than the comment (happy to remove it), but the same muP factor should be used due to the correlation.

dlwh: I read that as the **attention** operation not q_proj and k_proj?

In the mup repo, [they initialize k_proj the same as o_proj](https://github.com/microsoft/mup/blob/main/examples/Transformer/model.py#L296-L304). Interestingly, they do zero-initialize q_proj, which is from Appendix D.2.

src/levanter/models/llama.py
Helw150: SG

src/levanter/models/llama.py
Helw150: Yeah, good call. Will add on the Haliax PR first and then move here!

src/levanter/optim/mup.py
Helw150: Not well defined. Codex added in the prototype and it didn't feel worth removing, but I'll remove now!

src/levanter/layers/attention.py
dlwh: (if @WhenWen and you both agree, I'll relent)

Helw150: They then separately require that you normalize Q^TK by d yourself (https://github.com/microsoft/mup/blob/main/README.md?plain=1#L71-L73)

Which is the the same as just (Q/sqrt(d))^T(K/sqrt(d)) (which is what this does)

Helw150: ah nvm I'm dumb. Yeah, this double counts the divisor

WhenWen: So I guess we are off by d here still? It feels more natural to me to actually change the scaling_factor to d and use mup.HiddenLinearMuP here. 

The argument is that if we want to use QKNorm with MuP down the line, the current implementation will break as the activation scale will be normalized out.

src/levanter/layers/attention.py
```diff
@@ -1462,6 +1463,9 @@ class AttentionConfig:
     """Configuration for QK normalization. If None, no normalization is applied."""
 
     def __post_init__(self):
+        if self.use_mup and self.scaling_factor is None:
+            object.__setattr__(self, "scaling_factor", 1 / self.head_size)
```
dlwh, line 1467: i'd ratehr set this in `init` and not `__post_init__` (since I'd prefer to avoid setattr unless we have ot

src/levanter/optim/mup.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2025 The Levanter Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utilities for MuP-style learning rate scaling."""
+
+from __future__ import annotations
```
dlwh, line 6: rm?

dlwh (approved):

PR #1234, opened 2025-10-05
Add attention sink support to JAX Flash Attention

Calvin-Xu:
The JAX attention backend now natively supports attention sinks, and tests are added to actually test the JAX backend.

Completes one path of https://github.com/marin-community/levanter/issues/1215

### DIFF
--- a/src/levanter/layers/attention.py
+++ b/src/levanter/layers/attention.py
@@ -86,6 +86,7 @@ def dot_product_attention(
     scaling_factor: float | None = None,
     inference: bool = True,
     prng: PRNGKeyArray | None = None,
+    attn_sink: Optional[NamedArray] = None,
 ):
     """
     This method is similar to [haliax.nn.attention.dot_product_attention][] but it can use different backends for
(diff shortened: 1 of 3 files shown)


src/levanter/layers/attention.py
```diff
@@ -151,65 +152,127 @@ def dot_product_attention(
...
-                inference,
-                prng=prng,
-                attention_dtype=attention_dtype,
-                precision=precision,
-                flash_block_size=flash_block_size,
-                force_te=not was_default,
-                scaling_factor=scaling_factor,
-                logits_soft_cap=logits_soft_cap,
-            )
+            if attn_sink is None:
```
dlwh: imho slightly better to push this into _try_te_attention and let it emit a warning as applicable. In particular, the contract should be that if a backend was forced in config (e.g. the user specified NVTE), it ought to raise if it's not compatible

src/levanter/layers/attention.py
```diff
@@ -302,25 +357,54 @@ def dot_product_attention_with_sink(
...
+    flash_block_size: Optional[int] = None,
+    dropout: float = 0.0,
+    *,
+    logits_soft_cap: float | None = None,
+    scaling_factor: float | None = None,
+    inference: bool = True,
+    prng: PRNGKeyArray | None = None,
+):
+    """
+    Compatibility wrapper: forwards directly to `dot_product_attention` with `attn_sink=...`.
```
dlwh: can we just remove

src/levanter/models/flash_attention.py
```diff
@@ -68,22 +68,50 @@ def flash_attention(
...
+                Key,
+                q,
+                k,
+                v,
+                attn_sink,
+                mask=mask,
+                bias=bias,
+                attention_dtype=dtype,
+                precision=precision,
+                use_flash=False,  # force VANILLA
```
dlwh: can we not call dot_product_attention_with_sink in the attn_sink is None route too?

src/levanter/models/flash_attention.py
```diff
@@ -382,7 +444,23 @@ def do_inner_block(state):
...
     # dQ, (dK, dV) = hax.scan(do_kv_block, Tc)(dQ, jnp.arange(Tc.size))
     j, dQ, dK, dV = jax.lax.while_loop(lambda state: state[0] < Tc, do_kv_block, (0, dQ, dK, dV))
-    return dQ.rearrange(q.axes), dK.rearrange(k.axes), dV.rearrange(v.axes)
+
+    if attn_sink is not None:
+        q_batch_axes: Tuple[hax.Axis, ...] = hax.eliminate_axes(q.axes, (QPos, Key))
+        sink_prefix = attn_sink
+        for ax in q_batch_axes:
+            if ax not in sink_prefix.axes:
+                sink_prefix = sink_prefix.broadcast_axis(ax)
```
dlwh, line 430: pretty sure this could be one line

dlwh (approved):
few small things

PR #1236, opened 2025-10-07
Allow users to return additional metrics from loss functions.

rjpower:
User loss functions are automatically wrapped to have the form (loss,
metric_dict), where metric_dict is a mapping from str -> Metric.

Metrics are automatically constructed based on the key name, and support folding for proper accumulation.

### DIFF
--- a/src/levanter/callbacks/__init__.py
+++ b/src/levanter/callbacks/__init__.py
@@ -29,7 +29,9 @@ from levanter.utils.jax_utils import barrier_sync
 from levanter.utils.logging import save_xla_dumps_to_wandb
 
 
-def eval_loss_loop(loss_fn: LossFunctionWithMetrics, model, dataset, max_batches: Optional[int] = None, name: Optional[str] = None):
+def eval_loss_loop(
+    loss_fn: LossFunctionWithMetrics, model, dataset, max_batches: Optional[int] = None, name: Optional[str] = None
+) -> tuple[float, dict[str, float]]:
 
     total_loss = 0.0
     total_load_time = 0.0
--- a/src/levanter/callbacks/__init__.py
+++ b/src/levanter/callbacks/__init__.py
@@ -1,12 +1,12 @@
 # Copyright 2025 The Levanter Authors
 # SPDX-License-Identifier: Apache-2.0
 
+import os
 import sys
 import threading
 import time
-from typing import Callable, Optional
 from contextlib import contextmanager
-import os
+from typing import Callable, Optional
 
 import jax
 from tqdm_loggable.auto import tqdm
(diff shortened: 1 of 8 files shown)


rjpower:
I removed the defer_tracker_for_jit part b/c/ you said it wasn't working & was a bit confusing in the context, but happy to add it back; it seems like we can put it into the wrapped function here as well.

src/levanter/callbacks/__init__.py
dlwh: doc comment for loss_fn maybe? (I know this is my fault)

src/levanter/trainer.py
```diff
@@ -563,43 +625,64 @@ def _jit_train_step_fn_no_hook(self):
...
-            def obj_fun(trainable_model):
-                model = eqx.combine(trainable_model, state.model)
-                with hax.axis_mapping(self.compute_axis_mapping):
-                    model = self.mp.cast_to_compute(model)
-                    return self._raw_loss_function(model, *batch, **batch_kwargs, key=key).scalar()
-
-            new_state, updates = state.take_step(grads, obj_fun=obj_fun, loss=loss, key=new_key)
-            new_state = hax.shard(new_state, self.parameter_axis_mapping)
+        # Always returns (loss, grads, metrics) where metrics may be empty dict
+        loss, grads, loss_metrics = self._compute_gradients_microbatched(
```
dlwh: So, i'm not really sure what to do about this and it's a flaw with old approach, but _compute_gradients_microbatched  kind of breaks metrics. I think this implementaiton will average them across batches. I'm worried we'll have to define like Metric and a semiring thing (appropriately disguised but still 😨)

src/levanter/trainer.py
```diff
@@ -563,43 +625,64 @@ def _jit_train_step_fn_no_hook(self):
...
-                    hook_infos = self.hooks.run_jit_hooks(state, jit_info, force=False)
+        # Prepare metrics with train/ prefix for logging outside JIT
+        train_metrics = {f"train/{k}": v for k, v in loss_metrics.items()}
 
-        if _no_hooks:
-            return hax.shard_with_axis_mapping((loss, new_state, metrics, None), self.parameter_axis_mapping)
-        else:
-            # return loss, new_state, metrics, hook_infos
-            return hax.shard_with_axis_mapping((loss, new_state, metrics, hook_infos), self.parameter_axis_mapping)
+        # Sophia needs to be able to access the loss function in the optimizer
```
dlwh, line 650: i should probably just delete sophia

src/levanter/callbacks/__init__.py
rjpower: Sure, added a protocol with some documentation... I'll figure out a good place to put it...

src/levanter/trainer.py
rjpower: Ah good point, I'll fix it and try to infer sensible defaults if the user  returns a float.

PR #1238, opened 2025-10-07
Infer KV page count from HBM utilization

### LINKED ISSUE #1200: automatically calculate cache size based on model config and what the core inference loop needs

dlwh:
## Summary
- add an `hbm_utilization` knob to `InferenceEngineConfig` and expose it from the factory constructor
- infer a default KV page count via `eval_shape`/`tree_size` using the device HBM budget when `max_pages` is omitted
- fall back to the previous `max_seqs * max_pages_per_seq` sizing when memory statistics are unavailable

## Testing
- pre-commit run (via git commit hook)

fixes #1200 

------
https://chatgpt.com/codex/tasks/task_e_68e4b1ad30108331bee80d29e5b0195e

### DIFF
--- a/src/levanter/inference/engine.py
+++ b/src/levanter/inference/engine.py
@@ -6,6 +6,7 @@ import functools
 import logging
 import os
 import time
+import warnings
 from collections import deque
 from dataclasses import dataclass, field
 from typing import Optional, Sequence
@@ -1421,10 +1422,13 @@ class InferenceEngine:
             low = high
             while True:
                 high *= 2
-                if high > (1 << 31):
-                    raise ValueError(
-                        "Unable to infer KV cache size: search exceeded 2^31 pages. " "Provide `max_pages` explicitly."
+                if high > (1 << 20):
+                    warnings.warn(
+                        "KV cache size exceeded 1M pages during budget inference; "
+                        "aborting search and using current estimate."
                     )
+                    high = 1 << 20
+                    break
                 high_bytes = cache_bytes(high)
                 if high_bytes > budget:
                     break


src/levanter/utils/jax_utils.py
```diff
@@ -350,26 +351,126 @@ def create_fsdp_mesh(
...
+        total = 0.0
+        for buffer in device.live_buffers():
+            total += buffer.nbytes
+
+        usage[device] = total / (1024.0**3)
+
+    return usage
+
+
+def memory_info_string(color: bool) -> str:
```
rjpower: feels a little overkill but okay,  i might cleanup a bit with:

mem_types = { "free": free_mem or "unknown", "used": ..., "limit" }

src/levanter/inference/engine.py
```diff
@@ -1226,6 +1261,66 @@ def _all_done() -> bool:
                 self.results.pop(rid, None)
         return GenerationResult(tokens=outputs_list, logprobs=logprobs_list, total_generated=total_generated)
 
+    def write_kernel_jaxprs(self, path, log_artifacts: bool = True):
```
rjpower, line 1258: we don't have some auto-writer thing i take it?

src/levanter/inference/engine.py
```diff
@@ -1284,3 +1379,57 @@ def _ingest_outputs(self, outputs: _DecodeOutputs | None) -> int:
...
+            def initial_cache(num_pages: int) -> int:
+                table = PageTable.init(num_pages, config.max_seqs, config.page_size, max_pages_per_seq)
+                cache_shape = model.initial_cache(table, dtype=config.compute_dtype)
+                return cache_shape
+
+            cache_shape = eqx.filter_eval_shape(initial_cache, num_pages)
+
+            return _tree_byte_size(cache_shape)
+
+        bytes_one = cache_bytes(1)
```
rjpower, line 1407: rather than trying to compute a slope, we could binary-search over the space?

src/levanter/inference/engine.py
dlwh: there's a hacky thing in trainer for train_step but not in general no. haven't figured out what i'd want it to look like exactly...

src/levanter/inference/engine.py
dlwh: done!

src/levanter/inference/engine.py
```diff
@@ -1284,3 +1372,90 @@ def _ingest_outputs(self, outputs: _DecodeOutputs | None) -> int:
...
+
+        low = 1
+        high = guess
+        high_bytes = cache_bytes(high)
+
+        if high_bytes <= budget:
+            low = high
+            while True:
+                high *= 2
+                if high > (1 << 31):
```
rjpower: Nit, could just log warning and stop at some number X e.g. 1 << 20.

PR #1240, opened 2025-10-07
Once more unto the breach (of jit mesh nonsense)

dlwh:


src/levanter/data/_prp.py
```diff
@@ -75,6 +77,13 @@ def make(kind: PermType, length: int, prng_key: PRNGKeyArray, **kwargs) -> "Perm
             raise ValueError(f"Unknown permutation kind: {kind}")
 
 
+def _np_rng_from_jax_key(prng_key: PRNGKeyArray) -> np.random.Generator:
+    # Force CPU usage to avoid jit complaining. These are not critical path in Levanter.
+    with local_cpu_mesh():
+        key = jax.device_put(jax.device_get(prng_key))
```
rjpower, line 83: I won't claim I understand this but sure, the more device_* the better.

rjpower (approved):

PR #1248, opened 2025-10-10
Dataloading should fail if requested Mixture is not feasible

ahmeda14960:
Right now if we request mixtures through LMMixtureDataset that have neither the raw urls or a finished cached, levanter will just skip and not include the dataset in the mixture. This can lead to bad silent errors so instead we will complain loudly and fail

### DIFF
--- a/config/sampler/sample_llama8b.yaml
+++ b/config/sampler/sample_llama8b.yaml
@@ -13,11 +13,8 @@ trainer:
   model_axis_size: 4
 
 engine:
-  max_pages: 16384
+  max_pages: null
   max_seqs: 256
-  page_size: 8
-  max_pages_per_seq: 512
-  max_queued_tokens: 256
-  max_seqs_in_prefill: 16
-  max_prefill_size: 4096
+  page_size: 128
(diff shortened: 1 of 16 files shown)


src/levanter/data/text.py
```diff
@@ -1093,8 +1093,14 @@ def build_or_load_cache(
                 pass
 
         if source is None:
-            logger.warning(f"Skipping {split} because no source was provided")
-            return None
+            if split == "train":
```
rjpower, line 1096: Why is "train" special, should we always fail if we don't have the source?

dlwh: ok yeah so this isn't right. we have lots of sources where there is no training set (e.g. paloma)

ahmeda14960: Oh I was thinking we might not have a validation set while training but still continue training.... for example we might not have validation urls for each source in common pile, in which case this would fail.

For paloma, we always use it with a data config that has training urls correct? so this seems fine.

What are you thinking instead? require validation for each train source?

PR #1254, opened 2025-10-14
Minor changes to handle json and add docstrings by Cursors

chiheem:
1. lm-eval locked to a specific branch (which is actually main as of PR) to make uv happy. Maybe this will go away when we make levanter part of marin.
2. Added Kevin's class to convert non-serializable objects to string. These objects are sometimes emitted by lm-eval for unknown reasons.
3. Add docstrings using Cursor
4. Allow marin to override lm-eval config

### DIFF
--- a/src/levanter/eval_harness.py
+++ b/src/levanter/eval_harness.py
@@ -704,9 +704,13 @@ class LevanterHarnessLM(TemplateLM):
             # Copy and process generation kwargs
             processed_gen_kwargs = gen_kwargs.copy()
 
-            # Apply defaults from generation_kwargs (user config) first
+            # Override lm-eval defaults with our generation_kwargs (user config)
+            # This ensures our parameters take precedence over lm-eval's defaults
             for key, value in self.generation_kwargs.items():
-                processed_gen_kwargs.setdefault(key, value)
+                old_value = processed_gen_kwargs.get(key)
+                processed_gen_kwargs[key] = value
+                if old_value != value:
+                    logger.info(f"Overriding lm-eval config {key}={old_value} with {key}={value}")
 
             # Standardize kwargs using our _modify_gen_kwargs method
             processed_gen_kwargs = self._modify_gen_kwargs(processed_gen_kwargs)


src/levanter/eval_harness.py
```diff
@@ -711,16 +790,19 @@ def generate_until(self, requests) -> List[str]:
...
         engine_cfg = InferenceEngineConfig(
             max_stop_seqs=max_stop_seqs,
             max_stop_tokens=max_stop_tokens,
             max_seq_len=max_length,
             max_seqs=256,
             page_size=8,
             compute_dtype=jnp.bfloat16,
+            max_queued_tokens=256,
+            max_seqs_in_prefill=16,
+            max_prefill_size=max_length,
```
AlienKevin: With the new hbm_utilization parameter, we no longer need to hard code parameters like `max_queued_tokens`, `max_seqs_in_prefill`, and `max_prefill_size`.

src/levanter/eval_harness.py
```diff
@@ -86,6 +86,29 @@
...
+        Returns:
+            String representation of the object, or calls parent default method
+        """
+        if callable(obj):
+            return f"<function: {obj.__name__ if hasattr(obj, '__name__') else str(obj)}>"
+        elif hasattr(obj, "__dict__"):
+            return str(obj)
+        return super().default(obj)
+
+
```
AlienKevin: I also have a json encoder in this PR: https://github.com/marin-community/levanter/pull/1255. Marin also has a CustomJsonEncoder: https://github.com/marin-community/marin/blob/4dec0f6fdb33d72846a1a1a5279d0c6da6fc118d/src/marin/utilities/json_encoder.py#L26. Hopefully after the monorepo conversion is complete, we can just have a shared `CustomJsonEncoder`. For now, I think it makes sense to put this encoder in `levanter/src/levanter/utils/py_utils.py`.

After close inspection, none of the above three encoders are truely fail safe because the `str` and `repr` methods can be overridden by the subclass and throw exceptions. ChatGPT proposes the following implementation:
```
import json, datetime, decimal, uuid, pathlib, base64, enum
from dataclasses import is_dataclass, asdict

class FailSafeJSONEncoder(json.JSONEncoder):
    """
    A 'never-throw' JSON encoder:
    - Handles many common non-JSON types.
    - Degrades unknowns to a safe string payload.
    - Avoids blowing up on weird __repr__ or circulars.
    """
    def __init__(self, *args, bytes_strategy="base64", **kwargs):
        # bytes_strategy: "base64" | "repr" | "hex"
        super().__init__(*args, check_circular=True, **kwargs)
        self.bytes_strategy = bytes_strategy

    def default(self, obj):
        # Known clean conversions
        if isinstance(obj, (datetime.datetime, datetime.date, datetime.time)):
            # ISO 8601; preserves tzinfo if present
            try:
                return obj.isoformat()
            except Exception:
                return str(obj)

        if isinstance(obj, decimal.Decimal):
            # Prefer float; fallback to string if NaN/Inf
            try:
                f = float(obj)
                if f == float("inf") or f == float("-inf") or f != f:  # NaN check
                    return str(obj)
                return f
            except Exception:
                return str(obj)

        if isinstance(obj, uuid.UUID):
            return str(obj)

        if isinstance(obj, (set, frozenset)):
            return list(obj)

        if isinstance(obj, pathlib.Path):
            return str(obj)

        if isinstance(obj, complex):
            # JSON has no complex; encode as 2-tuple
            return {"__type__": "complex", "real": obj.real, "imag": obj.imag}

        if isinstance(obj, bytes):
            if self.bytes_strategy == "base64":
                return {"__type__": "bytes", "base64": base64.b64encode(obj).decode("ascii")}
            if self.bytes_strategy == "hex":
                return {"__type__": "bytes", "hex": obj.hex()}
            return repr(obj)

        if isinstance(obj, bytearray):
            return self.default(bytes(obj))

        if isinstance(obj, enum.Enum):
            # Serialize as its value when simple; else name
            try:
                val = obj.value
                # Make sure the value itself is JSON-serializable
                json.dumps(val)  # quick probe
                return val
            except Exception:
                return obj.name

        if is_dataclass(obj):
            # Convert dataclasses to dicts (lets the base encoder recurse)
            try:
                return asdict(obj)
            except Exception:
                pass  # fall through to string fallback

        # Functions / callables -> a safe label
        if callable(obj):
            name = getattr(obj, "__name__", None)
            return f"<function {name}>" if name else "<callable>"

        # Generic Python objects: DON'T explode—use a conservative string
        # Avoid user-defined __repr__ surprises where possible
        try:
            # Try a short, informative tag
            t = type(obj)
            addr = hex(id(obj))
            # Try __repr__, but guard it
            try:
                rep = repr(obj)
            except Exception:
                rep = t.__name__
            # Cap very long reprs to keep payloads sane
            if len(rep) > 500:
                rep = rep[:500] + "…"
            return {"__type__": f"{t.__module__}.{t.__name__}", "__repr__": rep, "__id__": addr}
        except Exception:
            # Absolute last resort
            return "<unserializable object>"
```
I think this is better in 2 ways:
1. Fail safe for any object type
2. Covers a broad range of types that we may encounter.

I'm open to discuss more and feel free to iterate on this starting point.

src/levanter/inference/engine.py
```diff
@@ -1177,40 +1177,40 @@ def _all_done() -> bool:
                 self.config.imputed_max_tokens_per_round,
                 self.config.max_rounds,
             )
-            submit_done = time.time()
+            # submit_done = time.time()
```
rjpower: Let's undo these? it doesn't hurt to calculate them?

src/levanter/inference/engine.py
```diff
@@ -1358,19 +1367,29 @@ def _extract_outputs(self, pending_outputs) -> int:
             dr = self.results.setdefault(rid, {}).setdefault(cid, DecodeResult(id=rid, choice=cid, token_list=[]))
             dr.done = True
 
-        num_finished = int(fins.sum()) if hasattr(fins, "sum") else 0
-        logger.info(f"extract: appended={appended} (drained={n}) unmapped={unmapped} finished_count={num_finished}")
+            # Print final complete text when sequence is finished
+            if print_every_n > 0 and dr.tokens_decoded > 0:
```
rjpower: This seems unnecessary, since you can just print the result at this point?

src/levanter/eval_harness.py
```diff
@@ -86,6 +86,29 @@
 logger = logging.getLogger(__name__)
 
 
+class SafeJSONEncoder(json.JSONEncoder):
```
rjpower: This looks duplicative with #1255 , maybe sync with @AlienKevin on which we should use.

src/levanter/eval_harness.py
```diff
@@ -236,12 +259,27 @@ def max_gen_toks(self) -> int:
         return self._generation_kwargs.get("max_gen_toks", 256)
 
     def make_harness_lm(self):
+        """
```
rjpower: I'm not sure this comment is adding much

src/levanter/inference/engine.py
chiheem: When you say "these", does it include printing "Decode iter ... (L1208)". I am having issues in the sense that they keep printing and overwhelm the output buffer when eval is running.

rjpower: I was just thinking the time calculation, we can keep calculating the time but change the log to logger.debug for example

chiheem: That is reasonable.

src/levanter/eval_harness.py
rjpower: Yeah, I'd avoid overthinking this for now, and simply use Kevin's version with a TODO that we should merge the JSON encoding work.

I don't think handling an exception in the repr call is worth doing -- doing _anything_ in Python can throw an exception -- the goal of the json encoding is to "try to produce a reasonable JSON version of well-known types with a string fallback", not "catch every possible exception".

Longer term I agree we should plan to merge implemtnations in the mono-repo. We may even want to adopt the Pydantic serializer: https://docs.pydantic.dev/latest/concepts/conversion_table/

chiheem: Added Kevin's code in `levanter/src/levanter/utils/py_utils.py` in the latest commit. Changed existing references to this new class.

src/levanter/eval_harness.py
chiheem: Removed these. Thanks!

src/levanter/eval_harness.py
chiheem: Resolved. See https://github.com/marin-community/levanter/pull/1254#discussion_r2430756864

src/levanter/inference/engine.py
chiheem: Agreed. Remove this and changed it to print to debug for consistency

src/levanter/inference/engine.py
chiheem: I have reverted the changes but made them print to logger.debug

src/levanter/eval_harness.py
```diff
@@ -391,17 +408,31 @@ def eot_token_id(self) -> int:
         return self.tokenizer.eos_token_id
 
     def set_current_task(self, task_name: str):
-        """Set the current task name for organizing sample outputs."""
+        """
```
rjpower: Nit, these comments tend to be more noise than signal, I'd tend to remove them, but it's up to you.

src/levanter/utils/py_utils.py
```diff
@@ -156,3 +164,105 @@ def set_global_rng_seeds(seed):
...
+    """
+
+    def __init__(self, *args, bytes_strategy="base64", **kwargs):
+        # bytes_strategy: "base64" | "repr" | "hex"
+        super().__init__(*args, check_circular=True, **kwargs)
+        self.bytes_strategy = bytes_strategy
+
+    def default(self, obj):
+        # Known clean conversions
+        if isinstance(obj, (datetime.datetime, datetime.date, datetime.time)):
```
rjpower, line 189: As noted:

remove the try/except around the 

* isoformat (what does it mean to have a date which can't be stringified)?
* for enum, the recursive json.dumps should also include CustomJsonEncoder, and remove the try/except
* dataclass todict should never fail
* callable and everything below it should just be repr(v)

src/levanter/utils/py_utils.py
```diff
@@ -156,3 +164,105 @@ def set_global_rng_seeds(seed):
...
+
+        if isinstance(obj, pathlib.Path):
+            return str(obj)
+
+        if isinstance(obj, complex):
+            # JSON has no complex; encode as 2-tuple
+            return {"__type__": "complex", "real": obj.real, "imag": obj.imag}
+
+        if isinstance(obj, bytes):
+            if self.bytes_strategy == "base64":
```
rjpower, line 214: this is overkill, just just base64 and don't make it a configuration.

rjpower (approved):
LGTM -- 

please get rid of some of the over-zealous try/except in the JSON encoding & the noisy docstring before submitting.

src/levanter/eval_harness.py
```diff
@@ -681,16 +713,31 @@ def generate_until(self, requests) -> List[str]:
         # Truncate from left if needed to fit model max length, accounting for generation tokens
         max_length = self.EvalPos.size
         for i, (toks, gen_kwargs) in enumerate(zip(prompt_token_lists, processed_kwargs_list)):
-            # Reserve space for generation tokens
-            max_gen_toks = gen_kwargs["max_gen_toks"]
-            max_ctx_len = max_length - max_gen_toks
+            # Dynamically compute max_gen_toks based on actual prompt length
```
AlienKevin: `max_gen_toks` is inherited from lm-eval-harness and it's configured in multiple benchmarks. Many backends (vLLM, SGL, APIs) also use this simple rule of left truncate if prompt_len + max_gen_toks > max_length: https://deepwiki.com/search/how-does-llm-backend-handle-ca_0bf4bbab-d556-4540-a1c1-0d89d9b1da14. To ensure consistency with lm-eval-harness, I think we should keep this field.

AlienKevin (commented):
LGTM, thanks! Just one nit about `max_gen_toks`, I don't think we should change the truncation logic.

src/levanter/eval_harness.py
rjpower: Replace all of the instances of JSON default -- we don't want to have both of these lying around.

src/levanter/eval_harness.py
chiheem: I understand that it came from lm-eval-harness.

However, I don't understand why do we want to prioritize generation over feeding in the entire prompt; if the input context is missing, do even want to evaluate the generation?

src/levanter/eval_harness.py
chiheem: Sorry. I removed the usages but not the class itself.

src/levanter/eval_harness.py
AlienKevin: yeah I think that's a very valid point. I think the logic may be that we still want to scramble whatever output the model can give. A model with short context length will just be penalized by the incomplete prompt but should still be able to generate output. I think lm-eval-harness's design choice here is definitely suboptimal. My take is we can bring this issue up to their repo and until they make any change in their design, we should stick with the official evaluation procedure, however poorly designed, to ensure a fair comparison against with other models and inference backends.

chiheem: I reviewed why I made such changes. 
1) The default AIME max_gen_toks is 32768. (https://github.com/EleutherAI/lm-evaluation-harness/blob/0c9072191e790204c4dfeb3285925b586a548c2e/lm_eval/tasks/aime/aime24.yaml#L25C1-L25C22)
2) Our max_length is only 8192
3) As such, we always truncate the prompt to an empty string if we follow the established rule.

chiheem: Thinking more about this... If we follow the established rules, this is what we will get:
1) prompt + max_gen_toks <= max_length --> no problem
2) max_gen_toks > max_length --> empty prompt
3) prompt > max_length --> partial to empty prompt

I wonder what is a reasonable thing to do.

rjpower: So if our intent is to compare against other implementations, then I agree with Kevin, we should follow the standard lm_eval behavior if possible, and have them change it upstream if it's incorrect. If the comparisons are for our own purposes, than we should do whatever we think is more proper.

Is it possible for us to bump max_length up to 32768? I assume it will tank performance since we might only be able to fit a single request per batch at that point -- we'd be at 4GB (bf16) for the KV-cache alone (though that should shard across the worker).

AlienKevin: > Is it possible for us to bump max_length up to 32768?

We can try this. If it's too slow perhaps we can create a customized [aime.yaml](https://github.com/EleutherAI/lm-evaluation-harness/blob/0c9072191e790204c4dfeb3285925b586a548c2e/lm_eval/tasks/aime/aime.yaml) that sets a shorter `max_gen_toks`? Since this prompt preparation procedure is standard and applies to all evals, I think it might be better to keep it stable and modify individual evals on a need-basis.

chiheem: I am testing if setting max_length = 32768 + 4096 works. The addition of 4096 is to account for the prompt.
Edit: This failed.
```
jaxlib._jax.XlaRuntimeError: RESOURCE_EXHAUSTED: Allocation (size=4718592) would exceed memory (size=1048576) :: #allocation9 [shape = 'u8[4718592]{0}', space=smem, size = 0x480000, tag = 'prefetched SMEM operand 1'] :: ragged_paged_attention.64
```

I am also thinking, what if I refactor the block into 2 functions: [default]maxGenToksFirst() and promptToksFirst(). We can toggle the behavior based on the task name or by passing in a flag.

AlienKevin: Or how about we add in a flag to eval_harness where we can override the gen_kwargs of lm-eval-harness? That way we can pass in a shorter `max_gen_toks` for some tasks.

chiheem: I thought of this. The thing is, it then requires the user to understand the number of prompt tokens (not easily available and differs instance to instance) to ensure that max_prompt_len + max_gen_toks <= max_length. I can implement this if we think that this is better.

AlienKevin: Yeah I agree, tho I can't think of a better implementation at the moment.

PR #1256, opened 2025-10-15
factor out a SequenceTable from PageTable/DecodeState, 

dlwh:
There were a bunch of things in DecodeState that duplicated what was in PageTable, and it was a big mess. Now it is a somewhat smaller mess.

### DIFF
--- a/src/levanter/inference/engine.py
+++ b/src/levanter/inference/engine.py
@@ -31,7 +31,7 @@ from levanter.inference.jit_scheduler import (
 )
 from levanter.inference.page_table import PageTable
 from levanter.inference.utils import INVALID, is_valid
-from levanter.layers.attention import KvPageCache
+from levanter.layers.kv_cache import PageCache
 from levanter.layers.sampler import Sampler
 from levanter.models.lm_model import LmHeadModel
 from levanter.utils.jax_utils import estimated_free_device_memory, sharded_tree_size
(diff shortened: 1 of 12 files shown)


src/levanter/inference/engine.py
```diff
@@ -448,19 +443,10 @@ def do_clone(gs_clone: GenState) -> GenState:
...
-                    assigned != slot_val,
-                    "Requested local slot mismatch during prefill.",
-                )
-                decode_state = dataclasses.replace(decode_state, page_table=page_table)
-                prompt_len = work.prompt_lengths.array[i].astype(jnp.int32)
+                decode_state, assigned = decode_state.reserve_slot(slot_val)
                 decode_state = decode_state.assign_seq(
                     local_slot_id=slot_val,
                     tokens=work.prompt_tokens["seq", i],
-                    seq_len=prompt_len,
```
dlwh, line 462: we actually want it to start out as 0 and get set during prefill

src/levanter/inference/engine.py
```diff
@@ -367,22 +364,20 @@ def _prefill_kernel(
...
 
     # Update decode_state (also enqueues into the main decode queue)
-    decode_state = gen_state.decode_state.update_tokens(new_tokens, new_slot_ids, log_probs, num_new_tokens)
+    decode_state = decode_state.update_tokens(new_tokens, new_slot_ids, log_probs, num_new_tokens)
 
     # Initialize outputs buffer and append prefill-sampled tokens
     outputs = _DecodeOutputs.init(
-        max_tokens=gen_state.decode_state.max_seqs * 2,
-        max_seqs=gen_state.decode_state.max_seqs,
+        max_tokens=decode_state.max_seqs * 2,
```
rjpower, line 376: RP: review this in more detail.

PR #1258, opened 2025-10-15
Explicitly copy to the CPU when preparing batches for TPU execution.

rjpower:


rjpower:
I think the right fix is to actually avoid entering the TPU mesh until we're only doing TPU operations, but maybe this will help for now.

PR #1263, opened 2025-10-16
Fix logprobs.tokens incorrect truncation by prompt_len

AlienKevin:
logprobs.tokens are redundantly truncated by prompt_len when it only contains generated tokens.

### DIFF
--- a/src/levanter/inference/openai.py
+++ b/src/levanter/inference/openai.py
@@ -520,8 +520,7 @@ async def _create_completion(ctx: InferenceContext, request: CompletionRequest) 
                 logprobs = None
                 if request.logprobs:
                     # Convert logprobs to API format
-                    prompt_len = generation.prompt_tokens
-                    generated_tokens = generation.tokens[prompt_len:]
+                    generated_tokens = generation.tokens
 
                     # Create token logprobs in OpenAI format
                     tokens = []


rjpower (approved):

rjpower:
This type of thing should be testable with a trivial (e.g. 1000 parameter) model -- we can just validate the outputs are the right size etc. If we initialize the weights deterministically we can even validate the predicted tokens & logprobs are all identical across changes.

I do this for the RL work and it's been a huge time-saver for me -- see test_cats_integration and config_helpers.py in marin for reference.

AlienKevin:
> This type of thing should be testable with a trivial (e.g. 1000 parameter) model -- we can just validate the outputs are the right size etc. If we initialize the weights deterministically we can even validate the predicted tokens & logprobs are all identical across changes.
> 
> I do this for the RL work and it's been a huge time-saver for me -- see test_cats_integration and config_helpers.py in marin for reference.

Yes! I'm preparing a PR to test greedy sampling results against HF reference implementation to prevent correctness regressions in the future.

dlwh:
CI depends on preemptible instances, and there haven't been preemptible instances. we should use a reserved node but don't currently (or actually move to v5p)

rjpower:
Yes I have a grand plan for having a server that sits and listens for webhooks, but it's stuck behind the grand plans for inference, ray and RL at the moment :)

rjpower:
I'll merge for you, let me see if I can adjust the rules to allow skipping CI once you have approval.

PR #1264, opened 2025-10-16
Add fail safe to stop generation if we hit max tokens.

rjpower:


rjpower:
I haven't been able to reproduce the issue that this is intended for unfortunately. It's some combination of big model + certain inputs.

@BabyChouSr saw infinite loops in generation where somehow the processing logic got stuck. This is just intended as a band-aid to ensure we always exit generation. I think moving the page allocation out and simplifying the TPU loop will help with this as well, and then we should be able to remove this.

This is just a hacky way of expressing:

for _ in range(max_batches):
   if _all_done(...)

rjpower:
I just changed it to be an actual for loop, I think that's more readable. Then it's obvious we're just looping for a maximum number of batches, and the TPU decoding logic simply lets us exit early if needed.

PR #1265, opened 2025-10-16
Basic correctness check against HF reference implementation by greedily sampling tokens

AlienKevin:
To prevent future correctness regressions in inference, this PR adds a basic correctness check in the spirit of vLLM's [test_basic_correctness.py](https://github.com/vllm-project/vllm/blob/938c43ea7fa72cf882403b97b3ea3884308a7e0f/tests/basic_correctness/test_basic_correctness.py) to check that Levanter outputs identical tokens to HuggingFace reference implementation under greedy sampling.

### DIFF
--- a/.agents/projects/gcs_serialization.md
+++ b/.agents/projects/gcs_serialization.md
@@ -1,0 +1,157 @@
+# GCS TensorStore Loader Revamp
+
+## Background
+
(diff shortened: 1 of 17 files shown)


tests/inference/test_inference_server.py
```diff
@@ -88,6 +89,88 @@ def test_client(baby_llama_config, loaded_model):
...
+
+        assert response.status_code == 200
+        payload = response.json()
+        choice = payload["choices"][0]
+        logprobs = choice.get("logprobs") or {}
+
+        tokens = logprobs.get("tokens") or []
+        if tokens:
+            token_ids = hf_tokenizer.convert_tokens_to_ids(tokens)
+        else:
```
rjpower: We shouldn't need this path?

AlienKevin: Good point, simplified the branching logic

PR #1270, opened 2025-10-17
Minimal inference engine for Levanter.

rjpower:


notes.md
```diff
@@ -0,0 +1,88 @@
...
+            new_k.array,
+            new_v.array,
+            K,
+        )
+```
+
+
+## Questions
+
+* Can we lift our KV cache management up to the CPU/top-level?
```
dlwh: i feel like only page allocation can be at this level? or is that what you mean

notes.md
```diff
@@ -0,0 +1,88 @@
...
+* What would we need to change to have per-sequence attention masks and e.g. left-pad pre-fill so we can allocate a page at a time?
```
dlwh: i don't think this is that much work probably

notes.md
```diff
@@ -0,0 +1,88 @@
...
+* Can we use the JAX KV format directly?
```
dlwh: we'd need to fuse the kv matmul which would slightly complicate HF export

dlwh (commented):
just read notes. will look at rest soon!

PR #1272, opened 2025-10-20
Support syncing weights from Levanter to vLLM

AlienKevin:
# Support syncing weights to vLLM
This PR leverages tpu-inference's `transfer_state_with_mappings` to transfer model weights from Levanter to vLLM on-the-fly without deserializing to disk.

Besides the weight transfer logic, the jax version is bumped to 0.7.2 to match vLLM's and mesh processing is updated in Levanter to account for changes to default value of `get_concrete_mesh`.

# Tests
The weight transfer logic is currently embedded in a unit test called `test_levanter_weight_transfer_to_vllm` which runs the following steps:
1. Loads the vLLM model with the Qwen3-0.6B base model weights
2. Verifies that generation matches expected outputs of base model
3. Loads the Qwen3-0.6B chat weights into Levanter
4. Transfers weights directly into the live vLLM model state from Levanter
5. Verifies that generation matches the expected outputs of chat model

The above test passed on v4-8 using the following command:
```bash
uv run --extra tpu pytest tests/test_vllm.py --log-cli-level=INFO -s
```

# References
@BabyChouSr initially found usage of [`transfer_state_with_mappings`](https://github.com/google/tunix/blob/c55e233303643b73576d84934d3ac4c69296c636/tunix/generate/vllm_sampler.py#L112-L120) in Tunix.

Turns out Tunix might have defined a custom version of `transfer_state_with_mappings` which might originate from tpu-inference:
1. [_sync_weights](https://github.com/vllm-project/tpu-inference/blob/42f7aac17564759e6114a138b667f3973377ae32/tpu_inference/runner/tpu_jax_runner.py#L752) in tpu-inference
2. [transfer_state_with_mappings](https://github.com/vllm-project/tpu-inference/blob/42f7aac17564759e6114a138b667f3973377ae32/tpu_inference/models/jax/utils/weight_utils.py#L465) from tpu-inference's weight_utils.py.
3. tpu-inference has a [unit test](https://github.com/vllm-project/tpu-inference/blob/42f7aac17564759e6114a138b667f3973377ae32/tests/models/jax/test_weight_loading.py#L60) that illustrates usage of transfer_state_with_mappings.

# TODOs
* Weight sync implementation should be extracted out of the unit test into a standalone module
* Test with more models besides Qwen 3 0.6B, e.g. Llama 3.2 1B

### DIFF
--- a/tests/test_vllm.py
+++ b/tests/test_vllm.py
@@ -191,34 +191,10 @@ def _run_generation_test_case(case: GenerationTestCase) -> None:
         del llm
 
 
-# Parametized test always fails on vLLM HBM usage exceeding limit, no matter how much HBM we allocated to it,
-# and no matter how we clear the Jax cache (delete all the live arrays, gc collect, clear cache, clear test cache).
-# vLLM will allocate all the assigned HBM to weights + KV cache. The conclusion is parametized test doesn't reset
-# Jax properly, therefore the 2nd test adds on top of the previous HBM usage. This is the workaround for that.
-# cf: https://github.com/google/tunix/blob/2b4ec654fa75c1f261f016e451066323de4dda7a/tests/generate/vllm_sampler_test.py#L87
-
 @pytest.mark.forked
-def test_vllm_model_generation_llama_base() -> None:
-    """Test that vLLM can generate from Llama base model correctly."""
-    case = GENERATION_TEST_CASES[0]  # llama_base
-    _run_generation_test_case(case)
-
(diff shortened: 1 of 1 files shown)


rjpower:
FYI We ran into issues with JAX 0.72 on Marin runs - there's an interaction with Ray where JAX crashes the process on TPUs. Let's make sure to test some jobs specifically on the Marin cluster with the new JAX version before we try to update here.

PR #1283, opened 2025-10-30
run 0.8.0 in CI

dlwh:


.github/workflows/tpu_unit_tests.yaml
```diff
@@ -35,10 +37,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jax_version: ["0.6.2", "0.7.2"]
+        jax_version: ["0.6.2", "0.7.2", "0.8.0"]
+        tpu_config:
```
ryan-williams, line 41: I think this `matrix` nullifies the `workspace_dispatch.inputs` above? i.e. no longer possible to just test one `{tpu_type,zone}`.

Maybe we should have one workflow that matrix's over a `workflow_call` to a single {TPU,zone} workflow test run?

ryan-williams: Actually for `workflow_dispatch`es, this might spawn 4 copies of the requested (tpu,zone), each with the 3 jax versions?

PR #1287, opened 2025-10-31
Minimal changes to run Qwen2.5 using eval_harness

chiheem:
Required changes to run Qwen2.5-Instruct.

Changes:
1. Rounding of vocab axis
2. Change handling max_seqs_in_prefill. Previous setting limits the model generation to 16 instances of of AIME24/25.
3. Add decode() function for qwen

Tests:
1. Successfully ran AIME24 benchmark from levanter-lm-eval-harness from marin

### DIFF
--- a/src/levanter/models/qwen.py
+++ b/src/levanter/models/qwen.py
@@ -168,7 +168,7 @@ class QwenDecoderLayer(LlamaDecoderLayer):
         ln_1 = config.mk_LayerNorm(config.Embed)
         ln_2 = config.mk_LayerNorm(config.Embed)
 
-        return QwenDecoderLayer(config, attn, mlp, ln_1, ln_2, None, None)
+        return QwenDecoderLayer(config, attn, mlp, ln_1, ln_2, post_attn_layernorm=None, post_mlp_layernorm=None)
 
     @named_call
     def __call__(
@@ -182,8 +182,6 @@ class QwenDecoderLayer(LlamaDecoderLayer):
 
         # Reuse the parent implementation for the rest
         return super().__call__(x, mask, key=key, pos_ids=pos_ids)
-
-    # decode and initial_cache are inherited from LlamaDecoderLayer
 
 
 # Modified transformer for Qwen


src/levanter/models/qwen.py
```diff
@@ -219,6 +254,49 @@ def __call__(
         x = self.norm(x)
         return x
 
+    @named_call
+    def decode(
```
AlienKevin, line 220: I noticed that the decode and initial_cache are directly copied from the LlamaTransformer superclass. I think those two methods should be inherited?

src/levanter/models/qwen.py
```diff
@@ -287,6 +365,61 @@ def init(cls, Vocab: Axis, config: QwenConfig, *, key) -> "QwenLMHeadModel":
 
         return QwenLMHeadModel(transformer, embeddings, lm_head)
 
+    def initial_cache(self, spec: PageTableSpec, *, dtype) -> ListCache[KvPageCache]:
```
AlienKevin, line 330: I noticed that initial_cache and decode are directly copied from https://github.com/marin-community/levanter/blob/77aa5913d28590c7b55e6ae5e004b4da5d4ba22d/src/levanter/models/llama.py#L510. Can we maybe inherit from `LlamaLMHeadModel` to reduce code duplication?

src/levanter/models/qwen.py
```diff
@@ -186,6 +190,37 @@ def __call__(
         output = residual + mlp_output
         return output
 
+    @named_call
```
AlienKevin: Could we inherit from [LlamaDecoderLayer](https://github.com/marin-community/levanter/blob/77aa5913d28590c7b55e6ae5e004b4da5d4ba22d/src/levanter/models/llama.py#L291C7-L291C24)? 

The `initial_cache` is identical and the `decode` function is also almost identical except for the two if conditions on `post_attn_layernorm` and `post_mlp_layernorm`. Since they are set to None by default, the behavior of llama's `decode` should be identical to this PR's implementation.

AlienKevin (commented):
@chiheem Thanks for testing and adding support for Qwen inference! One thing I noticed is a lot of code might be duplicated from Llama. Maybe it would be easier and more maintainable if Qwen inherit from Llama?

